PR #10, opened 2015-09-04
Add a minimal work-in-progress GUI with JavaFX

ThomasJClark:
Over time this will be filled in with actually widgets and actual
functionality, but for now this adds JavaFX to the build system and sets
up a basic interface.


javafx.plugin
```diff
@@ -0,0 +1,36 @@
+/*
```
JLLeitschuh: Shouldn't this file go in the gradle directory or the project?

src/main/java/edu/wpi/grip/Main.java
```diff
@@ -1,5 +1,33 @@
...
+import javafx.application.Application;
+import javafx.beans.value.ObservableValue;
+import javafx.fxml.FXMLLoader;
+import javafx.scene.Cursor;
+import javafx.scene.Parent;
+import javafx.scene.Scene;
+import javafx.stage.Screen;
+import javafx.stage.Stage;
+
+public class Main extends Application {
```
JLLeitschuh: How are we supposed to launch this application in headless mode now?
Shouldn't the application entry point be in `src/main/java/edu/wpi/grip/ui`?

javafx.plugin
JLLeitschuh: https://docs.gradle.org/current/userguide/organizing_build_logic.html#sec:build_sources

src/main/java/edu/wpi/grip/Main.java
ThomasJClark: Presumably headless mode would have a different main class.  We'll add then when we have a headless mode.

JLLeitschuh: Shouldn't this Main be the one that parses the input args and picks which one to launch?

javafx.plugin
ThomasJClark: I can't find an official source on this, but I would think all of the gradle stuff would be in the root directory.  I really just did it like that because of [the instructions for using the JavaFX plugin](http://speling.shemnon.com/blog/2013/10/29/javafx-gradle-plugin-0-dot-4-0-released/).

src/main/java/edu/wpi/grip/Main.java
ThomasJClark: There's only one to launch right now.  We can add more complexity when it becomes necessary.

PR #14, opened 2015-09-08
Default socket hint value gets cloned

JLLeitschuh:
This prevents the original from becoming the reference that all defaults point to.


src/main/java/edu/wpi/grip/core/SocketHint.java
```diff
@@ -69,7 +70,9 @@ public View getView() {
     }
 
     public T getDefaultValue() {
-        return defaultValue;
+        Cloner cloner = new Cloner();
```
ThomasJClark: Are you sure this works with OpenCV `Mat`s?  Maybe we just want to call the object's `.copy()` method.

PR #17, opened 2015-09-11
Add a Sink interface and DummySink implementation

ThomasJClark:
Sinks are outputs that typically subscribe to `SocketPublishedEvent` and write the values over some kind of network connection.

`DummySink` just prints out values to the console.  We should write a `NetworkTablesSink` eventually.


src/main/java/edu/wpi/grip/core/sinks/DummySink.java
```diff
@@ -0,0 +1,25 @@
...
+import com.google.common.eventbus.EventBus;
+import com.google.common.eventbus.Subscribe;
+import edu.wpi.grip.core.Sink;
+import edu.wpi.grip.core.Socket;
+import edu.wpi.grip.core.events.SocketPublishedEvent;
+
+/**
+ * A sink that just prints out any values it receives.
+ */
+public class DummySink implements Sink {
```
JLLeitschuh: I don't like this name but I'll let it slide.

PR #19, opened 2015-09-18
Add getName() and getDescription() to the Operation interface

ThomasJClark:
We need some way for operations to specify user-presentable strings of
information for the GUI.

### DIFF
--- a/src/main/java/edu/wpi/grip/core/operations/PythonScriptOperation.java
+++ b/src/main/java/edu/wpi/grip/core/operations/PythonScriptOperation.java
@@ -32,7 +32,7 @@ import java.util.Properties;
  *    ]
  * }</pre>
  *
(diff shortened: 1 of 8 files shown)


src/main/java/edu/wpi/grip/core/operations/PythonScriptOperation.java
```diff
@@ -61,12 +67,24 @@
...
 
     public PythonScriptOperation(URL url) throws PyException, IOException {
         this.sourceURL = Optional.of(url);
         this.sourceCode = Optional.absent();
         this.interpreter.execfile(url.openStream());
         this.getPythonVariables();
+
+        if (this.name == null) {
+            // If a name of the operation wasn't specified in the script, use the basename of the URL
+            final String path = url.getPath();
```
JLLeitschuh: URL is optional.

PR #20, opened 2015-09-18
Add basic JavaFX controls for rendering steps in the pipeline

ThomasJClark:


src/main/java/edu/wpi/grip/ui/SocketControlView.java
```diff
@@ -0,0 +1,172 @@
...
+        switch (socketHint.getView()) {
+            case NONE:
+                // The hint says not to show a control, so there's nothing to do here.
+                break;
+
+            case SPINNER: {
+                // To use a spinner, the socket should be able to hold a number
+                checkArgument(Number.class.isAssignableFrom(socketHint.getType()));
+
+                SpinnerValueFactory spinnerValueFactory;
```
JLLeitschuh: Why is this defined up here.

src/main/java/edu/wpi/grip/ui/SocketControlView.java
```diff
@@ -0,0 +1,172 @@
...
+
+                Number min = Double.MIN_VALUE, max = Double.MAX_VALUE;
+                if (domain != null && domain.length == 2) {
+                    // A spinner can have a min and max, if the domain holds two values.
+                    min = (Number) domain[0];
+                    max = (Number) domain[1];
+                }
+
+                spinnerValueFactory = new SpinnerValueFactory.DoubleSpinnerValueFactory(
+                        min.doubleValue(), max.doubleValue(), initialValue.doubleValue());
```
JLLeitschuh: But assigned down here

src/main/java/edu/wpi/grip/ui/SocketControlView.java
```diff
@@ -0,0 +1,172 @@
...
+            // will change appearance when the socket's value changes, such as a slider.
+            this.valueProperty.setValue(this.socket.getValue());
+        }
+    }
+
+    @Subscribe
+    public void onConnectionAdded(ConnectionAddedEvent event) {
+        // Disable the control for the socket if its value is set by a connection anyway
+        if (event.getConnection().getInputSocket() == this.socket)
+            this.controlPane.disableProperty().set(this.socket.isConnected());
```
JLLeitschuh: This is really bad practice. Either keep it on one line or wrap it in braces.
https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/

src/main/java/edu/wpi/grip/ui/SocketControlView.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    @Subscribe
+    public void onConnectionRemoved(ConnectionRemovedEvent event) {
+        // Enable the control for the socket if the connection influencing its value was removed
+        if (event.getConnection().getInputSocket() == this.socket)
+            this.controlPane.disableProperty().set(this.socket.isConnected());
```
JLLeitschuh: Same as above

src/main/java/edu/wpi/grip/ui/models/StepModel.java
```diff
@@ -0,0 +1,21 @@
+package edu.wpi.grip.ui.models;
+
+import edu.wpi.grip.core.Step;
+import javafx.beans.property.SimpleStringProperty;
+import javafx.beans.property.StringProperty;
+
+/**
+ * Created by tom on 9/14/15.
```
JLLeitschuh: Remove attribution

src/main/java/edu/wpi/grip/ui/SocketControlView.java
```diff
@@ -0,0 +1,172 @@
...
+                RangeSlider rangeSlider = new RangeSlider(min, max, initialValue, initialValue + 10);
+                this.controlPane.getChildren().add(rangeSlider);
+                this.valueProperty.bindBidirectional(rangeSlider.lowValueProperty());
+
+                break;
+            }
+            case SELECT:
+                // TODO: implement a select control once we have some operations that have enum inputs
+                break;
+        }
```
JLLeitschuh: No default case defined.

src/main/java/edu/wpi/grip/ui/SocketControlView.java
```diff
@@ -0,0 +1,172 @@
...
+        // Always set the socket to the value of valueProperty, which may be bound to a GUI property.
+        this.valueProperty.addListener((observableValue, o, t1) -> this.socket.setValue(observableValue.getValue()));
+
+        this.eventBus.register(this);
+    }
+
+    @Override
+    public void initialize(URL url, ResourceBundle resourceBundle) {
+        SocketHint<?> socketHint = this.socket.getSocketHint();
+        Object[] domain = socketHint.getDomain();
```
JLLeitschuh: Since the domain can be null shouldn't it be returning an Optionally wrapped domain?

src/main/java/edu/wpi/grip/ui/SocketControlView.java
```diff
@@ -0,0 +1,172 @@
...
+
+                SpinnerValueFactory spinnerValueFactory;
+                Number initialValue = (Number) socket.getValue();
+
+                Number min = Double.MIN_VALUE, max = Double.MAX_VALUE;
+                if (domain != null && domain.length == 2) {
+                    // A spinner can have a min and max, if the domain holds two values.
+                    min = (Number) domain[0];
+                    max = (Number) domain[1];
+                }
```
JLLeitschuh: What should happen if the domain is not two? Should we log a problem to the console?

src/main/java/edu/wpi/grip/ui/SocketControlView.java
```diff
@@ -0,0 +1,172 @@
...
+        this.socket = socket;
+        this.valueProperty = new SimpleObjectProperty();
+
+        try {
+            FXMLLoader fxmlLoader = new FXMLLoader(getClass().getResource("SocketControl.fxml"));
+            fxmlLoader.setRoot(this);
+            fxmlLoader.setController(this);
+            fxmlLoader.load();
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
JLLeitschuh: We probably want to implement our own irrecoverable exception implementation that extends RuntimeException.
This is much better practice than throwing a raw runtime exception. This also provides a common exception that we can catch and handle in a standardized way.

src/main/java/edu/wpi/grip/ui/SocketControlView.java
```diff
@@ -0,0 +1,172 @@
...
+            }
+            case SLIDER: {
+                // To use a slider, the socket should be able to hold a number
+                checkArgument(Number.class.isAssignableFrom(socketHint.getType()));
+
+                // A slider needs a min and max, so the domain should hold two values.
+                checkArgument(domain != null && domain.length == 2);
+                double min = ((Number) domain[0]).doubleValue(),
+                        max = ((Number) domain[1]).doubleValue(),
+                        initialValue = ((Number) socket.getValue()).doubleValue();
```
JLLeitschuh: Duplicate logic for slider. Same method calls just reads a little differently.
Consider reducing to one implementation.

src/main/java/edu/wpi/grip/ui/SocketControlView.java
```diff
@@ -0,0 +1,172 @@
...
+            case RANGE: {
+                // A range produces an array of two numbers, so the socket should expect an array that can hold numbers
+                checkArgument(socketHint.getType().isArray());
+                checkArgument(Number.class.isAssignableFrom(socketHint.getType().getComponentType()));
+
+                // Same
+                checkArgument(domain != null && domain.length == 2);
+                double min = ((Number) domain[0]).doubleValue(),
+                        max = ((Number) domain[1]).doubleValue(),
+                        initialValue = ((Number) socket.getValue()).doubleValue();
```
JLLeitschuh: Duplicate logic again.

src/main/java/edu/wpi/grip/ui/SocketControlView.java
JLLeitschuh: Also, why do you check if the domain is of length two above but not here.

JLLeitschuh:
This pull request should only have commits that haven't been merged. It would be better if you rebased your code on master.

src/main/java/edu/wpi/grip/core/Socket.java
```diff
@@ -123,12 +124,27 @@ public boolean isPublished() {
...
 
+    /**
+     * @param connected If true, this socket is automatically set by a connection
+     */
+    public void setConnected(boolean connected) {
+        this.connected = connected;
+    }
+
+    /**
+     * @return Weather or not this socket is automatically set
```
JLLeitschuh: Wrong `weather` should be `whether`

JLLeitschuh: This could be phrased better as `True when this socket is automatically connected`

src/main/java/edu/wpi/grip/ui/SocketControlView.java
```diff
@@ -0,0 +1,172 @@
...
+
+                RangeSlider rangeSlider = new RangeSlider(min, max, initialValue, initialValue + 10);
```
JLLeitschuh: Why is the initial value defined with 10 added to it?
Add a comment explaining this.

src/main/java/edu/wpi/grip/ui/SocketControlView.java
ThomasJClark: It's not, it's assigned on the same line it's defined.

src/main/java/edu/wpi/grip/ui/SocketControlView.java
ThomasJClark: This is an enum, so there are only these possible values.

src/main/java/edu/wpi/grip/ui/SocketControlView.java
ThomasJClark: Yeah, this is one place where it would be possible to use `Optional`.  It doesn't right know, so the client code that uses it doesn't either.

src/main/java/edu/wpi/grip/ui/SocketControlView.java
ThomasJClark: You've answered your own question...  It's not duplicate code because of that.

src/main/java/edu/wpi/grip/ui/SocketControlView.java
JLLeitschuh: Look at line 88. The type declaration is on a different line. I'm wondering why you chose to do that.

src/main/java/edu/wpi/grip/ui/SocketControlView.java
JLLeitschuh: Yes but the enum could be added to later. You should always define the default case in a case switch statement.

src/main/java/edu/wpi/grip/ui/SocketControlView.java
ThomasJClark: The idea with spinners is that the domain is optional, since there isn't really a need for a spinner to have an upper and lower bound (other than those imposed by the min/max values a double can hold)

I guess we could be really pedantic and throw an error if the domain exists but isn't something we can use, but I kind of want to avoid throwing errors for things relating to the hint.  The name itself implies that the world won't end if the current implementation can do exactly what it specified.

Once we have logging and error reporting in place (which we should eventually have) it might be a good idea to put a log statement anywhere that a hint is not 100% respected by the GUI.

src/main/java/edu/wpi/grip/ui/SocketControlView.java
ThomasJClark: `RuntimeException`s aren't really supposed to be caught.  I'm throwing a `RuntimeException` here because that should never fail, and if it does, I want to immediately close the program because there's no way it can work.  In practice, this should absolutely only happen if we do something weird during development, so we want to see the program crash and a stacktrace show up.

As for the "much better practice" argument, I'm not sure if I agree.  Since this exception should *theoretically* never be thrown, we don't have a way of naming it (like `FooFailedException` or `BarReturnedErrorException`), since we don't know what the exception means.  Plus I don't like bloating our class hierarchy with a bunch of one-off exceptions.

src/main/java/edu/wpi/grip/core/Socket.java
ThomasJClark: > Wrong `weather` should be `whether`

I take full responsibility for this failture.

src/main/java/edu/wpi/grip/ui/SocketControlView.java
JLLeitschuh: Actually, the real reason that I want to have a base uncachable exception is so that we can put methods inside it like `toMarkdown` or `toIssue` so that we can show a UI-screen to allow users to copy the error into an issue really easily.

src/main/java/edu/wpi/grip/ui/SocketControlView.java
JLLeitschuh: How about using asserts? Those can be turned off when we build a release.

ThomasJClark:
Travis is fixed, this can be merged if there are no other concerns.

PR #22, opened 2015-09-25
Step gui

ThomasJClark:


src/main/java/edu/wpi/grip/ui/controllers/MainWindowController.java
```diff
@@ -1,9 +1,134 @@
...
+                eventBus.post(new ConnectionAddedEvent(connection2));
+
+                Thread.sleep(1000);
+
+                eventBus.post(new ConnectionRemovedEvent(connection1));
+
+                Thread.sleep(1000);
+
+                eventBus.post(new StepRemovedEvent(step3));
+            } catch (Exception e) {}
```
JLLeitschuh: Probably shouldn't silently catch the exception. It should make some noise.

PR #24, opened 2015-09-27
Add and remove connections

ThomasJClark:
The basic work for adding and removing connections between steps is done.  You can add a new connection by clicking and dragging one socket to another.  You can remove connections by clicking either of the associated sockets.

Still todo:
* Ensure that only one connection is associated with a single input socket at any time
* Ensure that the user doesn't make "backwards" connections (these can cause infinite feedback loops and freeze the program if the values never converge)
* ~~Re-position `ConnectionView`s when the socket handles move around~~
* Maybe show a fancy animated `ConnectionView` while the user is dragging

### DIFF
--- a/src/main/java/edu/wpi/grip/core/Socket.java
+++ b/src/main/java/edu/wpi/grip/core/Socket.java
@@ -2,9 +2,13 @@ package edu.wpi.grip.core;
 
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Optional;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.eventbus.EventBus;
-import edu.wpi.grip.core.events.SocketChangedEvent;
-import edu.wpi.grip.core.events.SocketPublishedEvent;
+import com.google.common.eventbus.Subscribe;
+import edu.wpi.grip.core.events.*;
+
+import java.util.HashSet;
+import java.util.Set;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
@@ -16,12 +20,12 @@ import static com.google.common.base.Preconditions.checkNotNull;
  * results in are referred to as "output sockets".
  */
 public class Socket<T> {
-    private EventBus eventBus;
+    private final EventBus eventBus;
     private Step step;
-    private SocketHint<T> socketHint;
+    private final Set<Connection> connections = new HashSet<>();
+    private final SocketHint<T> socketHint;
     private Optional<T> value;
     private boolean published = false;
-    private boolean connected = false;
 
     /**
      * @param eventBus   The Guava {@link EventBus} used by the application.
@@ -36,6 +40,8 @@ public class Socket<T> {
         checkNotNull(eventBus);
         checkNotNull(socketHint);
         checkNotNull(value);
+
+        this.eventBus.register(this);
     }
 
     /**
@@ -49,6 +55,8 @@ public class Socket<T> {
 
         checkNotNull(eventBus);
         checkNotNull(socketHint);
+
+        this.eventBus.register(this);
     }
 
     /**
@@ -103,6 +111,13 @@ public class Socket<T> {
     }
 
     /**
+     * @return The set of connections that have this socket as an input or output
+     */
+    public Set<Connection> getConnections() {
+        return ImmutableSet.copyOf(this.connections);
+    }
+
+    /**
      * @param published If <code>true</code>, this socket will be published by any sink that is currently active.  For
      *                  example, it may be set as a NetworkTables value.
      */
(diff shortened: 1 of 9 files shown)


src/main/java/edu/wpi/grip/core/Socket.java
```diff
@@ -18,10 +21,10 @@
 public class Socket<T> {
     private EventBus eventBus;
     private Step step;
+    private Set<Connection> connections = new HashSet<>();
```
JLLeitschuh: Final?

src/main/java/edu/wpi/grip/core/Socket.java
```diff
@@ -103,6 +110,13 @@ public Step getStep() {
     }
 
     /**
+     * @return The set of connections that have this socket as an input or output
+     */
+    public Set<Connection> getConnections() {
+        return this.connections;
```
JLLeitschuh: Guava has an immutable set. Manipulations of this set should only happen inside of this class.

src/main/java/edu/wpi/grip/ui/SocketControlView.java
```diff
@@ -38,17 +45,17 @@
...
     private StackPane controlPane;
 
-    // The "handle" is a simple radio button next ot the socket identifier that shows weather or not there is a
-    // connection going to or from the socket.  If there is such a connection, the ConnectionView is rendered as a
-    // curve going from one handle to another.
-    @FXML
-    private RadioButton handle;
+    // The "handle" is a simple shape next ot the socket identifier that shows weather or not there is a connection
+    // going to or from the socket.  If there is such a connection, the ConnectionView is rendered as a curve going
+    // from one handle to another.
```
JLLeitschuh: Turn this into a javadoc comment. You are describing the object and its purposes.

src/main/java/edu/wpi/grip/ui/SocketControlView.java
```diff
@@ -38,17 +45,17 @@
...
+    private SocketHandleView handle;
```
JLLeitschuh: Final?

src/main/java/edu/wpi/grip/core/events/SocketConnectedChangedEvent.java
```diff
@@ -0,0 +1,36 @@
+package edu.wpi.grip.core.events;
+
+import com.google.common.base.MoreObjects;
+import edu.wpi.grip.core.Socket;
+
+/**
+ * This event is sent when the state of a socket (connected vs. disconnected) is changed.  Specifically, this will only
+ * be called if the number of connections either goes to zero when it was non-zero, or when it was zero and one is
+ * added.
+ * <p>
```
JLLeitschuh: `<br>`
For newline in javadocs.

ThomasJClark: I don't think so.  IntelliJ automatically adds a `<p>`, and it looks like that has the same effect.

src/main/java/edu/wpi/grip/ui/SocketHandleView.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * The bubble next to each socket that connections go in and out of
+ * <p>
```
JLLeitschuh: `<br>`

ThomasJClark: nope

src/main/java/edu/wpi/grip/ui/SocketHandleView.java
```diff
@@ -0,0 +1,135 @@
...
+
+        this.connectingProperty.addListener((observableValue, oldValue, isConnecting) ->
+                this.pseudoClassStateChanged(CONNECTING_PSEUDO_CLASS, isConnecting));
+
+        this.connectedProperty.addListener((observableValue, oldValue, isConnected) ->
+                this.pseudoClassStateChanged(CONNECTED_PSEUDO_CLASS, isConnected));
+
+        // When the user clicks on a socket, remove any connections associated with that socket.
+        this.setOnMouseClicked(mouseEvent -> {
+            Set<Connection> connections = this.controlView.getSocket().getConnections();
```
JLLeitschuh: You could make this even more generic and turn it into a `Collection`.

src/main/java/edu/wpi/grip/core/events/SocketConnectedChangedEvent.java
JLLeitschuh: Upon further inspection it seems you are right. Whatever floats your boat.

PR #27, opened 2015-09-29
Add SocketHint default values with Suppliers

ThomasJClark:
Instead of specifying a default value for sockets where no other value
is set, you can specify a function that returns an initial value for any
socket that isn't initialized with a value.  This ensures that all
sockets have a value, and also that those values can each be a new
instance (for mutable types like Mat).

For example, to make every socket have 5 as an initial value:

    sh = new SocketHint<>("my-socket", Integer.class, 5);

But to make every socket construct a new Mat as an initial value (so
they don't all share the same instance):

    sh = new SocketHint<>("my-socket", Mat.class, Mat::new);

When constructing the sockets themselves, it's not necessary to specify a
default value, but that constructor is still there in case you want to
immediately set a value other than the initial one from the socket hint.

### DIFF
--- a/src/main/java/edu/wpi/grip/core/Socket.java
+++ b/src/main/java/edu/wpi/grip/core/Socket.java
@@ -24,7 +24,7 @@ public class Socket<T> {
     private Step step;
     private final Set<Connection> connections = new HashSet<>();
     private final SocketHint<T> socketHint;
-    private Optional<T> value;
+    private T value;
     private boolean published = false;
 
     /**
@@ -35,7 +35,7 @@ public class Socket<T> {
     public Socket(EventBus eventBus, SocketHint<T> socketHint, T value) {
         this.eventBus = eventBus;
         this.socketHint = socketHint;
-        this.value = Optional.of(value);
+        this.value = value;
 
         checkNotNull(eventBus);
         checkNotNull(socketHint);
@@ -51,7 +51,7 @@ public class Socket<T> {
     public Socket(EventBus eventBus, SocketHint<T> socketHint) {
         this.eventBus = eventBus;
         this.socketHint = socketHint;
-        this.value = Optional.absent();
+        this.value = socketHint.createInitialValue();
 
         checkNotNull(eventBus);
         checkNotNull(socketHint);
@@ -73,8 +73,8 @@ public class Socket<T> {
      * @param value The value to store in this socket.
      */
     public void setValue(T value) {
-        if (!this.value.isPresent() || !this.value.get().equals(value)) {
-            this.value = Optional.of(this.getSocketHint().getType().cast(value));
+        if (!this.value.equals(value)) {
+            this.value = this.getSocketHint().getType().cast(value);
             eventBus.post(new SocketChangedEvent(this));
 
             // If the socket's value is set to be published, also send a SocketPublishedEvent to notify any sinks that
@@ -89,11 +89,7 @@ public class Socket<T> {
      * @return The value currently stored in this socket.
      */
     public T getValue() {
-        if (value.isPresent()) {
-            return value.get();
-        } else {
-            return socketHint.getDefaultValue();
-        }
+        return this.value;
     }
 
     /**
--- a/src/main/java/edu/wpi/grip/core/SocketHint.java
+++ b/src/main/java/edu/wpi/grip/core/SocketHint.java
@@ -3,6 +3,7 @@ package edu.wpi.grip.core;
 import com.google.common.base.MoreObjects;
 
 import java.util.Arrays;
+import java.util.function.Supplier;
 
 /**
  * A <code>SocketHint</code> is a descriptor that gives some information about one of the inputs or outputs of an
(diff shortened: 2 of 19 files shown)


src/main/java/edu/wpi/grip/core/SocketHint.java
```diff
@@ -12,44 +13,47 @@
 public class SocketHint<T> {
     public enum View {NONE, SPINNER, SLIDER, RANGE, SELECT}
 
-    private String identifier;
-    private Class<T> type;
-    private View view;
-    private T[] domain;
+    private final String identifier;
+    private final Class<T> type;
+    private final Supplier<T> initialValueSupplier;
```
JLLeitschuh: Should be optional

src/main/java/edu/wpi/grip/core/SocketHint.java
```diff
@@ -12,44 +13,47 @@
...
-     * @param domain       A hint at the range of values that this socket can hold.  For numeric types, this can consist
-     *                     of two elements that correspond to a minimum and maximum value.  The property does not make
-     *                     sense for all types and is left unspecified for some.
-     * @param defaultValue The value to use in place of the socket's value when none is specified.
-     */
-    public SocketHint(String identifier, Class<T> type, View view, T[] domain, T defaultValue) {
+    public SocketHint(String identifier, Class<T> type, Supplier<T> initialValueSupplier, View view, T[] domain) {
         this.identifier = identifier;
         this.type = type;
+        this.initialValueSupplier = initialValueSupplier;
```
JLLeitschuh: Should be `Optional.ofNullable(...)`

src/main/java/edu/wpi/grip/core/SocketHint.java
ThomasJClark: Why?

src/main/java/edu/wpi/grip/core/SocketHint.java
```diff
@@ -12,44 +13,47 @@
...
         this.view = view;
         this.domain = domain == null ? null : domain.clone();
-        this.defaultValue = defaultValue;
     }
 
-    public SocketHint(String identifier, Class<T> type, View view, T[] domain) {
-        this(identifier, type, view, domain, (domain != null && domain.length > 0) ? domain[0] : null);
+    public SocketHint(String identifier, Class<T> type, T initialValue, View view, T[] domain) {
+        this(identifier, type, () -> initialValue, view, domain);
```
JLLeitschuh: What if null?

src/main/java/edu/wpi/grip/core/SocketHint.java
```diff
@@ -12,44 +13,47 @@
...
+    }
+
+    public SocketHint(String identifier, Class<T> type, Supplier<T> initialValueSupplier, View view) {
+        this(identifier, type, initialValueSupplier, view, null);
+    }
+
+    public SocketHint(String identifier, Class<T> type, T initialValue, View view) {
+        this(identifier, type, () -> initialValue, view);
```
JLLeitschuh: Nullable

src/main/java/edu/wpi/grip/core/SocketHint.java
```diff
@@ -79,7 +83,6 @@ public String toString() {
                 .add("type", getType())
                 .add("view", getView())
                 .add("domain", Arrays.toString(getDomain()))
-                .add("defaultValue", getDefaultValue())
```
JLLeitschuh: You didn't add it to the tostring.

src/main/java/edu/wpi/grip/core/operations/opencv/AddOperation.java
```diff
@@ -13,9 +13,9 @@
  */
 public class AddOperation implements Operation {
     private SocketHint<Mat>
-            aHint = new SocketHint<>("a", Mat.class),
-            bHint = new SocketHint<>("b", Mat.class),
-            sumHint = new SocketHint<>("sum", Mat.class);
+            aHint = new SocketHint<Mat>("a", Mat.class, Mat::new),
```
JLLeitschuh: `<Mat>` isn't necessary can just be `<>`

src/main/java/edu/wpi/grip/core/Socket.java
```diff
@@ -35,7 +35,7 @@
     public Socket(EventBus eventBus, SocketHint<T> socketHint, T value) {
         this.eventBus = eventBus;
         this.socketHint = socketHint;
-        this.value = Optional.of(value);
+        this.value = value;
 
         checkNotNull(eventBus);
         checkNotNull(socketHint);
```
JLLeitschuh: Below this line you have `checkNotNull(value);` this can't be there. The value can be null by default.

src/test/java/edu/wpi/grip/core/PythonTest.java
```diff
@@ -71,7 +70,7 @@ public void testPythonWrongOutputCount() throws Exception {
         aSocket.setValue(a);
         bSocket.setValue(b);
 
-        assertNull(sumSocket.getValue());
+        assertEquals(0, sumSocket.getValue());
```
JLLeitschuh: While changing this you should add a message as a param.

src/test/java/edu/wpi/grip/core/PythonTest.java
```diff
@@ -85,7 +84,7 @@ public void testPythonWrongOutputType() throws Exception {
...
```
JLLeitschuh: While you are changing this anyways you should add a message.

src/main/java/edu/wpi/grip/core/SocketHint.java
JLLeitschuh: Lets discuss this in person.
Opencv has methods that take null as an allowable param. In some places it would actually be beneficial.

PR #28, opened 2015-09-29
Add a getDirection() method to sockets to distinguish inputs/outputs

ThomasJClark:
This can be used by the GUI to validate some user input, and also to
make it easier and cleaner to add specific things to the input and
output JavaFX controls (like options to preview outputs)

Basically, `Step` (as well as any other class that creates sockets) is
responsible for setting a direction on the socket after creating it.  By not
putting it in SocketHint, we avoid a ton of redundant boilerplate code in
every single operation and python script, and we also don't break
our existing code.

### DIFF
--- a/src/main/java/edu/wpi/grip/core/Connection.java
+++ b/src/main/java/edu/wpi/grip/core/Connection.java
@@ -6,6 +6,7 @@ import edu.wpi.grip.core.events.ConnectionRemovedEvent;
 import edu.wpi.grip.core.events.SocketChangedEvent;
 import edu.wpi.grip.core.events.StepRemovedEvent;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 /**
@@ -29,6 +30,10 @@ public class Connection<T> {
         checkNotNull(inputSocket);
         checkNotNull(outputSocket);
         checkNotNull(eventBus);
+        checkArgument(!Socket.Direction.INPUT.equals(outputSocket.getDirection()),
+                "outputSocket cannot be an input socket");
+        checkArgument(!Socket.Direction.OUTPUT.equals(inputSocket.getDirection()),
+                "inputSocket cannot be an output socket");
 
         if (inputSocket == outputSocket) {
             throw new IllegalArgumentException("inputSocket cannot be the same as outputSocket");
--- a/src/main/java/edu/wpi/grip/core/Socket.java
+++ b/src/main/java/edu/wpi/grip/core/Socket.java
@@ -15,13 +15,16 @@ import static com.google.common.base.Preconditions.checkNotNull;
 /**
  * A socket is a wrapper for a value that can be updated and passed around operations.  Sockets contain a set of hints
  * about the data they contain, as well as an actual value.
- * <p/>
+ * <p>
  * Sockets that are given to operations are referred to as "input sockets", and sockets that operations store their
  * results in are referred to as "output sockets".
  */
 public class Socket<T> {
+    public enum Direction {INPUT, OUTPUT}
+
     private final EventBus eventBus;
     private Step step;
+    private Direction direction;
     private final Set<Connection> connections = new HashSet<>();
     private final SocketHint<T> socketHint;
     private T value;
@@ -36,6 +39,7 @@ public class Socket<T> {
         this.eventBus = eventBus;
         this.socketHint = socketHint;
         this.value = value;
+        this.direction = Direction.INPUT;
 
         checkNotNull(eventBus);
         checkNotNull(socketHint);
@@ -52,6 +56,7 @@ public class Socket<T> {
         this.eventBus = eventBus;
         this.socketHint = socketHint;
         this.value = socketHint.createInitialValue();
+        this.direction = Direction.INPUT;
 
         checkNotNull(eventBus);
         checkNotNull(socketHint);
@@ -104,6 +109,24 @@ public class Socket<T> {
      */
     public Step getStep() {
         return step;
+    }
+
+
+    /**
+     * @param direction <code>INPUT</code> if this is the input to a step or sink, <code>OUTPUT</code> if this is the
+     *                  output of a step or source
+     */
+    public void setDirection(Direction direction) {
+        checkNotNull(direction);
+        this.direction = direction;
+    }
+
+    /**
+     * @return <code>INPUT</code> if this is the input to a step or sink, <code>OUTPUT</code> if this is the output of
+     * a step or source
+     */
+    public Direction getDirection() {
+        return this.direction;
     }
 
     /**
--- a/src/test/java/edu/wpi/grip/core/ConnectionTest.java
+++ b/src/test/java/edu/wpi/grip/core/ConnectionTest.java
@@ -14,7 +14,11 @@ public class ConnectionTest {
     @Test
     public void testInputSocketChanges() {
         Socket<Double> foo = new Socket<>(eventBus, fooHint);
+        foo.setDirection(Socket.Direction.OUTPUT);
+
         Socket<Double> bar = new Socket<>(eventBus, barHint);
+        bar.setDirection(Socket.Direction.INPUT);
+
         Connection<Double> connection = new Connection<>(eventBus, foo, bar);
 
         foo.setValue(testValue);
@@ -46,5 +50,16 @@ public class ConnectionTest {
     public void testInputSocketNotNull() {
         Socket<Double> foo = new Socket<>(eventBus, fooHint);
         Connection<Double> connection = new Connection<>(eventBus, foo, null);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testBackwardsConnection() {
+        Socket<Double> output = new Socket<>(eventBus, fooHint);
+        output.setDirection(Socket.Direction.OUTPUT);
+
+        Socket<Double> input= new Socket<>(eventBus, fooHint);
+        input.setDirection(Socket.Direction.INPUT);
+
+        new Connection<Double>(eventBus, input, output);
     }
 }
--- a/src/test/java/edu/wpi/grip/core/StepTest.java
+++ b/src/test/java/edu/wpi/grip/core/StepTest.java
@@ -34,6 +34,18 @@ public class StepTest {
     }
 
     @Test
+    public void testSocketDirection() {
+        Step step = new Step(eventBus, addition);
+        Socket<Double> a = (Socket<Double>) step.getInputSockets()[0];
+        Socket<Double> b = (Socket<Double>) step.getInputSockets()[1];
+        Socket<Double> c = (Socket<Double>) step.getOutputSockets()[0];
+
+        assertEquals(Socket.Direction.INPUT, a.getDirection());
+        assertEquals(Socket.Direction.INPUT, b.getDirection());
+        assertEquals(Socket.Direction.OUTPUT, c.getDirection());
+    }
+
+    @Test
     public void testGetOperation() {
         Step step = new Step(eventBus, addition);
 
--- a/src/main/java/edu/wpi/grip/core/Step.java
+++ b/src/main/java/edu/wpi/grip/core/Step.java
@@ -29,11 +29,13 @@ public class Step {
         inputSockets = operation.createInputSockets(eventBus);
         for (Socket<?> socket : inputSockets) {
             socket.setStep(this);
+            socket.setDirection(Socket.Direction.INPUT);
         }
 
         outputSockets = operation.createOutputSockets(eventBus);
         for (Socket<?> socket : outputSockets) {
             socket.setStep(this);
+            socket.setDirection(Socket.Direction.OUTPUT);
         }
 
         operation.perform(inputSockets, outputSockets);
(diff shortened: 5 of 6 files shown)


src/main/java/edu/wpi/grip/core/Connection.java
```diff
@@ -29,6 +30,10 @@ public Connection(EventBus eventBus, Socket<? extends T> outputSocket, Socket<T>
         checkNotNull(inputSocket);
         checkNotNull(outputSocket);
         checkNotNull(eventBus);
+        checkArgument(!Socket.Direction.INPUT.equals(outputSocket.getDirection()),
+                "outputSocket cannot be an input socket");
+        checkArgument(!Socket.Direction.OUTPUT.equals(inputSocket.getDirection()),
+                "inputSocket cannot be an output socket");
```
JLLeitschuh: These should be checks inside of socket.
Better: `inputSocket.isInput()'

ThomasJClark: How can a socket check this?  It doesn't know if it's being added as an input or output socket.

src/main/java/edu/wpi/grip/core/Socket.java
```diff
@@ -36,6 +39,7 @@ public Socket(EventBus eventBus, SocketHint<T> socketHint, T value) {
         this.eventBus = eventBus;
         this.socketHint = socketHint;
         this.value = value;
+        this.direction = Direction.INPUT;
```
JLLeitschuh: I don't like that this assumes a socket is automatically an input. I think forcing you to explicitly define the direction is better practice.

JLLeitschuh:
Why is the direction on Socket instead of Socket Hint?
More curious why you chose one over the other. It seems like this would be more valid in "hinting" at what the purpose of the socket is over defining the functionality of the socket.

ThomasJClark:
It's in the PR comment

JLLeitschuh:
Right. Missed that. Thanks

src/main/java/edu/wpi/grip/core/Connection.java
JLLeitschuh: The socket has a `direction` inside of it that it can check. You are reaching inside of a class to do the comparison where it would be better to politely ask the class for what it is.

JLLeitschuh:
Basically you have defined a getter and a setter on a class. This information can very easily be easily provided at instantiation.

src/main/java/edu/wpi/grip/core/Connection.java
ThomasJClark: I don't see what's wrong with that.  I'm calling a getter to get a field of the class, and doing a comparison with it.  There's no need to put the validation code inside of the socket class.

JLLeitschuh: > A fundamental precept of OO systems is that an object should not expose any of its implementation details. This way, you can change the implementation without changing the code that uses the object. It follows then that in OO systems you should avoid getter and setter functions since they mostly provide access to implementation details.
> -- Allen Holub [Why getter and setter methods are evil](http://www.javaworld.com/article/2073723/core-java/why-getter-and-setter-methods-are-evil.html)

src/test/java/edu/wpi/grip/core/ConnectionTest.java
```diff
@@ -14,7 +14,11 @@
     @Test
     public void testInputSocketChanges() {
         Socket<Double> foo = new Socket<>(eventBus, fooHint);
+        foo.setDirection(Socket.Direction.OUTPUT);
+
         Socket<Double> bar = new Socket<>(eventBus, barHint);
+        bar.setDirection(Socket.Direction.INPUT);
```
JLLeitschuh: Why are you doing this?

src/test/java/edu/wpi/grip/core/ConnectionTest.java
```diff
@@ -47,4 +51,26 @@ public void testInputSocketNotNull() {
...
         Connection<Double> connection = new Connection<>(eventBus, foo, null);
     }
+
+    @Test
+    public void testForwardsConnection() {
+        Socket<Double> output = new Socket<>(eventBus, fooHint);
+        output.setDirection(Socket.Direction.OUTPUT);
+
+        Socket<Double> input= new Socket<>(eventBus, fooHint);
+        input.setDirection(Socket.Direction.INPUT);
```
JLLeitschuh: Same here, why is this line here?

src/test/java/edu/wpi/grip/core/ConnectionTest.java
```diff
@@ -47,4 +51,26 @@ public void testInputSocketNotNull() {
...
+        new Connection<Double>(eventBus, output, input);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testBackwardsConnection() {
+        Socket<Double> output = new Socket<>(eventBus, fooHint);
+        output.setDirection(Socket.Direction.OUTPUT);
+
+        Socket<Double> input= new Socket<>(eventBus, fooHint);
+        input.setDirection(Socket.Direction.INPUT);
```
JLLeitschuh: Why is this here?

src/test/java/edu/wpi/grip/core/StepTest.java
```diff
@@ -34,6 +34,18 @@ public void testStep() {
...
     @Test
+    public void testSocketDirection() {
+        Step step = new Step(eventBus, addition);
+        Socket<Double> a = (Socket<Double>) step.getInputSockets()[0];
+        Socket<Double> b = (Socket<Double>) step.getInputSockets()[1];
+        Socket<Double> c = (Socket<Double>) step.getOutputSockets()[0];
+
+        assertEquals(Socket.Direction.INPUT, a.getDirection());
+        assertEquals(Socket.Direction.INPUT, b.getDirection());
+        assertEquals(Socket.Direction.OUTPUT, c.getDirection());
```
JLLeitschuh: You haven't proved anything about the underlying implementation with this test.
This basically proves that when you set a variable you can get it back out again the same way.

JLLeitschuh: Also, you forgot to add assertion messages.

src/main/java/edu/wpi/grip/core/Connection.java
ThomasJClark: It's not bad practice at all to have a getter return an enumeration, and you're not going to convince me that getters and setters are evil.

ThomasJClark: I realize that some blog you found disagrees with this idea, but this is how everybody else writes Java.

src/main/java/edu/wpi/grip/core/Socket.java
ThomasJClark: How can I force that? Throw an exception if it's not specified? Keep in mind that this field is set by steps, not operaions.  Operations specify which sockets are inputs buy return them from a certain method, and Steps immediately tag them with an enum (since that enum really describes the socket's relationship with the step,  and also for practical reasons already outlined)

src/test/java/edu/wpi/grip/core/ConnectionTest.java
```diff
@@ -47,4 +51,26 @@ public void testInputSocketNotNull() {
...
+
+    @Test
+    public void testForwardsConnection() {
+        Socket<Double> output = new Socket<>(eventBus, fooHint);
+        output.setDirection(Socket.Direction.OUTPUT);
+
+        Socket<Double> input= new Socket<>(eventBus, fooHint);
+        input.setDirection(Socket.Direction.INPUT);
+
+        new Connection<Double>(eventBus, output, input);
```
JLLeitschuh: What is this test proving? Some sort of assertion should be here even if you are just testing that an exception isn't thrown.

src/main/java/edu/wpi/grip/core/Connection.java
JLLeitschuh: Actually, this was used as teaching material in the OO class. Professor Fisler gave it to us as a reading.

src/main/java/edu/wpi/grip/core/Socket.java
JLLeitschuh: Force it by requiring it as a constructor argument.

src/main/java/edu/wpi/grip/core/Socket.java
```diff
@@ -106,6 +111,24 @@ public Step getStep() {
         return step;
     }
 
+
+    /**
+     * @param direction <code>INPUT</code> if this is the input to a step or sink, <code>OUTPUT</code> if this is the
+     *                  output of a step or source
+     */
+    public void setDirection(Direction direction) {
```
JLLeitschuh: If you are going to leave this setter here please at least remove the `public` specifier and leave it package access only.

src/main/java/edu/wpi/grip/core/Connection.java
ThomasJClark: * Sockets know nothing about connections or the direction they go.  Putting that logic in the socket class would be leaking `Connection`'s implementation details into somewhere they don't belong
* The `Direction` enum and the fact that sockets have a direction is already public.  Adding a getter doesn't leak anything.
* Again, you're not going to convince me that getters are evil, and there's no way we're going through the entire codebase and removing all of the getters.  All of the other fields in `Socket` and `SocketHint` have getters, so this one does too.

src/main/java/edu/wpi/grip/core/Socket.java
ThomasJClark: We've already discussed why it can't be a constructor argument

src/test/java/edu/wpi/grip/core/ConnectionTest.java
ThomasJClark: Yeah, this test is kind of useless

src/test/java/edu/wpi/grip/core/StepTest.java
ThomasJClark: I didn't forget, I just only add assertion messages when it's particularly unclear what the assertion is.

src/test/java/edu/wpi/grip/core/ConnectionTest.java
ThomasJClark: To explicitly set the directions of the two sockets so we can make a connection between them.  In the application, this is done in `Step` immediately after they're created, but for the test I'm manually setting them.

src/test/java/edu/wpi/grip/core/ConnectionTest.java
ThomasJClark: This test is kind of useless

src/main/java/edu/wpi/grip/core/Socket.java
ThomasJClark: This is actually a good idea, but Sources also need to be able to set socket directions, and those are in a different package.

JLLeitschuh: Why do they also need to set the direction?

ThomasJClark: So you can a connection from a source to a step

src/main/java/edu/wpi/grip/core/Connection.java
JLLeitschuh: Alternatively you could explicitly define a class called `InputSocket` and `OutputSocket`. Then you could make a connection only support one or the other. This would provide compile time safety.

src/main/java/edu/wpi/grip/core/Socket.java
JLLeitschuh: Operations construct sockets. You know when you are constructing a socket whether or not it is going to be an input or an output.

src/main/java/edu/wpi/grip/core/Connection.java
JLLeitschuh: That would clean up the constructor for Connection
```java
public Connection(EventBus eventBus, OutputSocket<? extends T> outputSocket, InputSocket<T> inputSocket)
```
Then you don't have to do this runtime checking anyways.

JLLeitschuh: That would allow you to change this:
```java
    @Subscribe
    public void onOutputChanged(SocketChangedEvent e) {
        if (e.getSocket() == outputSocket) {
            inputSocket.setValue(outputSocket.getValue());
        }
    }
```

To be something more like this:
```java
    @Subscribe
    public void onOutputChanged(SocketChangedEvent e) {
        if (e.getSocket() == outputSocket) {
            inputSocket.connect(outputSocket);
        }
    }
```
Allowing you to make the `setValue` function package protected on SocketInput.

src/test/java/edu/wpi/grip/core/ConnectionTest.java
ThomasJClark: So explicitly say that it's an input socket.  For the purposes of the unit test, I want to manually create a connection, which must be between an input socket and an output socket.

src/test/java/edu/wpi/grip/core/ConnectionTest.java
ThomasJClark: same

ThomasJClark:
Unless you have any comments beyond the ones I've addressed, I'm going to merge this today. I don't like merging my own PRs, but I don't mind doing it if it's getting in the way of any actual development happening and there doesn't seem to be any serious valid concerns about these changes.

JLLeitschuh:
You can't merge this one. You would then have overlapping implementations of input and output.

ThomasJClark:
Can you clarify what you mean by that?

PR #29, opened 2015-09-30
Adds InputSocket & OutputSocket

JLLeitschuh:
Since input sockets and output sockets have different functionality and
need to be distinguished they have been seperated into different
classes.

### DIFF
--- a/src/main/java/edu/wpi/grip/core/OutputSocket.java
+++ b/src/main/java/edu/wpi/grip/core/OutputSocket.java
@@ -1,0 +1,118 @@
+package edu.wpi.grip.core;
+
+
+import com.google.common.base.MoreObjects;
+import com.google.common.eventbus.EventBus;
+import edu.wpi.grip.core.events.SocketPreviewChangedEvent;
+import edu.wpi.grip.core.events.SocketPublishedEvent;
+
+/**
+ * Represents the output of an {@link Operation}. The OutputSocket also provides the ability to
+ * make the value in a socket published.
+ * @param <T> The type of the value that this socket stores.
+ */
+public class OutputSocket<T> extends Socket<T> {
+
+    /**
+     * Indicates if the socket is being previewed
+     */
+    private boolean previewed = false;
+    /**
+     * Indicates if this socket is published or not.
+     */
+    private boolean published = false;
+
+    /**
+     * @param eventBus   The Guava {@link EventBus} used by the application.
+     * @param socketHint See {@link #getSocketHint()}
+     * @param value      See {@link #getValue()}
+     */
+    public OutputSocket(EventBus eventBus, SocketHint<T> socketHint, T value) {
+        super(eventBus, socketHint, value, Direction.OUTPUT);
+    }
+
+    /**
+     * @param eventBus   The Guava {@link EventBus} used by the application.
+     * @param socketHint {@link #getSocketHint}
+     */
+    public OutputSocket(EventBus eventBus, SocketHint<T> socketHint){
+        super(eventBus, socketHint, Direction.OUTPUT);
+    }
+
+    @Override
+    public void setValue(T value){
+        // Check equality first before the value is ste
+        final boolean isNewValue = !this.getValue().equals(value);
+        // Set the value now that we know if there is a new value being stored
+        super.setValue(value);
+        // Now that the value is stored then publish the event
+        if (isNewValue) {
+            // If the socket's value is set to be published, also send a SocketPublishedEvent to notify any sinks that
+            // it has changed.
+            if (this.isPublished()) {
+                eventBus.post(new SocketPublishedEvent(this));
+            }
+        }
+    }
+
+    /**
+     * @param published If <code>true</code>, this socket will be published by any sink that is currently active.  For
+     *                  example, it may be set as a NetworkTables value.
+     */
+    public void setPublished(boolean published) {
+        /* Check if this value is changing */
+        final boolean change = isPublished() != published;
+        /* Save whether it was published before. */
+        final boolean wasPublished = isPublished();
+        /* Set this now so that it is published when the event handlers run */
+        this.published = published;
+        /*
+         * If the socket wasn't previously published and is now, send a SocketPublishedEvent to publish an initial
+         * value.
+         */
+        if (!wasPublished && change) {
+            eventBus.post(new SocketPublishedEvent(this));
+        }
+    }
+
+    /**
+     * @return Whether or not this socket should be published.
+     * @see #setPublished(boolean)
+     */
+    public boolean isPublished() {
+        return this.published;
+    }
+
+    /**
+     * @param previewed If <code>true</code>, this socket will be shown in a preview in the GUI.
+     */
+    public void setPreviewed(boolean previewed) {
+        boolean changed = previewed != this.previewed;
+        this.previewed = previewed;
+
+        // Only send an event if the field was actually changed
+        if (changed) {
+            eventBus.post(new SocketPreviewChangedEvent(this));
+        }
+    }
+
+    /**
+     * @return Whether or not this socket is shown in a preview in the GUI
+     * @see #setPreviewed(boolean) d(boolean)
+     */
+    public boolean isPreviewed() {
+        return this.previewed;
+    }
+
+
+    @Override
+    public String toString(){
+        return MoreObjects.toStringHelper(this)
+                .add("socketHint", getSocketHint())
+                .add("value", getValue())
+                .add("previewed", isPreviewed())
+                .add("published", isPublished())
+                .add("direction", getDirection())
+                .toString();
+    }
+}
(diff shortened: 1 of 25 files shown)


ThomasJClark:
Also, input and output sockets should be different controls

JLLeitschuh:
Please move your comments into the [file changes](https://github.com/WPIRoboticsProjects/GRIP/pull/29/files) so I can comment on them and they will show up in this PR.

ThomasJClark:
Which comments?  I all of my comments on individual lines are showing up here.

JLLeitschuh:
I don't see them.

JLLeitschuh:
I don't understand what you mean by different controls.

src/main/java/edu/wpi/grip/ui/SocketHandleView.java
```diff
@@ -88,29 +85,20 @@ public SocketHandleView(EventBus eventBus, SocketControlView controlView) {
...
-                // Check which socket was the input and which was the output.  The user can create a connection by
-                // dragging either into the other.
-                if (other.isInputSocket()) {
-                    inputSocket = other.getSocket();
-                    outputSocket = this.controlView.getSocket();
-                } else {
-                    inputSocket = this.controlView.getSocket();
-                    outputSocket = other.getSocket();
+                try {
+                    Connection connection = this.controlView.getSocket().connect(other.getSocket());
```
JLLeitschuh: This hides the implementation details from the UI. The UI doesn't need to know how a socket connects just that it does connect.

src/main/java/edu/wpi/grip/ui/SocketHandleView.java
```diff
@@ -88,29 +85,20 @@ public SocketHandleView(EventBus eventBus, SocketControlView controlView) {
...
-                @SuppressWarnings("unchecked")
-                Connection connection = new Connection(this.eventBus, outputSocket, inputSocket);
-                eventBus.post(new ConnectionAddedEvent(connection));
             });
         });
 
         // Accept input sockets being dragged onto output sockets and vice versa.
         this.setOnDragOver(dragEvent -> {
             draggingSocket.ifPresent(other -> {
-                if (other.isInputSocket() != this.controlView.isInputSocket()) {
```
JLLeitschuh: This exposes implementation details.

src/main/java/edu/wpi/grip/ui/SocketHandleView.java
```diff
@@ -88,29 +85,20 @@ public SocketHandleView(EventBus eventBus, SocketControlView controlView) {
...
+                if (other.canAcceptTransfer(this.controlView)) {
```
JLLeitschuh: This hides it and allows this rule to change.

ThomasJClark: No, sockets know nothing about connections (or "transfers" as you've apparently started calling them).  We've already defined the role of the socket class, let's not tack on more functionality that really shouldn't be in there.

src/main/java/edu/wpi/grip/ui/SocketHandleView.java
ThomasJClark: No, this is a UI thing.  Since you get an onDragOver as soon as you start dragging a socket, we need to make sure that the user has dragged it away to another socket before we actually handle those events.

src/main/java/edu/wpi/grip/ui/SocketHandleView.java
ThomasJClark: Sockets don't need to know how they connect, `Connection` is supposed to do that.

ThomasJClark:
> Also, I don't understand what you mean by different controls.
I mean that there should be one JavaFX control that represents input sockets and one for output sockets, instead of one `SocketControlView` that takes either and changes between two different controls.

src/main/java/edu/wpi/grip/core/OutputSocket.java
```diff
@@ -0,0 +1,86 @@
...
+     * @param eventBus   The Guava {@link EventBus} used by the application.
+     * @param socketHint {@link #getSocketHint}
+     */
+    public OutputSocket(EventBus eventBus, SocketHint<T> socketHint){
+        super(eventBus, socketHint);
+    }
+
+    @Override
+    public boolean setValue(T value){
+        boolean valueSet = super.setValue(value);
```
JLLeitschuh: Move this. Remove boolean.

src/main/java/edu/wpi/grip/core/Step.java
```diff
@@ -53,21 +46,21 @@ public Operation getOperation() {
...
-    public Socket<?>[] getOutputSockets() {
+    public OutputSocket<?>[] getOutputSockets() {
         return outputSockets;
     }
 
     @Subscribe
     public void onInputSocketChanged(SocketChangedEvent e) {
         // If this socket that changed is one of the inputs to this step, run the operation with the new value.
-        if (e.getSocket().getStep() == this) {
+        if (Arrays.asList(inputSockets).contains(e.getSocket())) {
```
ThomasJClark: Revert this change.  This has absolutely nothing to do with the rest of the commit.

src/main/java/edu/wpi/grip/ui/InputSocketView.java
```diff
@@ -25,7 +26,7 @@
 import static com.google.common.base.Preconditions.checkNotNull;
 
 /**
- * A JavaFX control that renders a {@link Socket} that is an input to a step.  This includes an identifier, a handle
+ * A JavaFX control that renders a {@link InputSocket} that is an input to a step.  This includes an identifier, a handle
```
ThomasJClark: a -> an

src/main/java/edu/wpi/grip/core/Step.java
JLLeitschuh: You want it in a separate PR or separate commit in this PR?

ThomasJClark: Nope

src/main/java/edu/wpi/grip/core/Socket.java
```diff
@@ -13,38 +12,39 @@
...
+ * Sockets that are given to operations are {@link InputSocket InputSockets}, and sockets that operations store their
+ * results are {@link OutputSocket OutputSockets}.
+ *
+ * @param <T> The type of the value that this socket stores
  */
-public class Socket<T> {
+public abstract class Socket<T> {
     public enum Direction {INPUT, OUTPUT}
 
     private final EventBus eventBus;
```
ThomasJClark: This should be made protected, right?  For when subclasses need to send events. (This isn't leaking anything since almost every class in the application has access to the same eventbus)

JLLeitschuh: Eh, I thought about that but this is better practice. If I make it protected then other objects in the package can access the variable.

src/main/java/edu/wpi/grip/core/Step.java
JLLeitschuh: Nope to which?

ThomasJClark: Both.  I mean, I guess you can do a PR if you want to, but that doesn't mean it'll get merged.

src/main/java/edu/wpi/grip/core/Socket.java
```diff
@@ -207,9 +152,7 @@ public String toString() {
         return MoreObjects.toStringHelper(this)
                 .add("socketHint", getSocketHint())
                 .add("value", getValue())
-                .add("published", isPublished())
-                .add("previewed", isPreviewed())
-                .add("connections", getConnections())
+                .add("direction", getDirection())
                 .toString();
     }
```
JLLeitschuh: InputSocket doesn't overload this method.

ThomasJClark: ok

src/main/java/edu/wpi/grip/core/OutputSocket.java
```diff
@@ -0,0 +1,121 @@
...
+        super(eventBus, socketHint, Direction.OUTPUT);
+        this.eventBus = eventBus;
+    }
+
+    @Override
+    public void setValue(T value){
+        // Check equality first before the value is ste
+        final boolean isNewValue = !this.getValue().equals(value);
+        // Set the value now that we know if there is a new value being stored
+        super.setValue(value);
```
JLLeitschuh: It is impossible to set the value of the parent class without calling super.

ThomasJClark: `value` and `eventBus` should be proected because the subclasses need to access them.  Your workaround is just a hack to intercept the private variables, it's not better design because you avoid the `protected` keyword.

I would definitely just make the fields  `protected` and avoid all of this.

JLLeitschuh: Protected means package access. Not what you want.

PR #31, opened 2015-09-30
Add buttons to set output sockets as previewed and published

ThomasJClark:
The user can use these to select which sockets to send over the network connection as the output of the algorithm, as well as which sockets to show in the preview pane (once it's implemented) to see in real-time what the algorithm is doing.

Because this only applies to output sockets, and there's already stuff that only applies to input sockets, this change splits up `SocketControlView` into two new classes (`InputSocketView` and `OutputSocketView`) rather than adding more and more conditionals into the  `SocketControlView` class to change the behavior for each case.

Future work may include separating `Socket` itself into two classes, but it would be much easier to do that after this change than before it, since `InputSocketView` and `OutputSocketView` can be updated separately to use the two new socket classes.  Until/unless this change is made, a `getDirection()` method is needed in `Socket`.


src/main/java/edu/wpi/grip/ui/InputSocketView.java
```diff
@@ -116,7 +110,7 @@ public void initialize(URL url, ResourceBundle resourceBundle) {
 
                 Number initialValue = (Number) socket.getValue();
 
-                StringBuilder stringBuilder = new StringBuilder();
+                StringBuilder stringsBuilder = new StringBuilder();
```
JLLeitschuh: ??

src/main/java/edu/wpi/grip/ui/OutputSocketView.java
```diff
@@ -0,0 +1,119 @@
...
+
+        this.handlePane.getChildren().add(this.handle);
+
+        // Show a button to publish the output to the current sink
+        this.publish.setSelected(this.socket.isPublished());
+        this.publish.selectedProperty().addListener(value -> this.socket.setPublished(this.publish.isSelected()));
+
+        // Show a button to choose if we want to preview the socket or not
+        this.preview.setSelected(this.socket.isPreviewed());
+        this.preview.selectedProperty().addListener(value -> this.socket.setPreviewed(this.preview.isSelected()));
```
JLLeitschuh: `isPreviewed()`

ThomasJClark: Nevermind, this is actually correct

PR #32, opened 2015-10-05
Add a preview pane

ThomasJClark:
For most types, the preview pane shows the `.toString()` of all output sockets marked as previewed.  For any output sockets that hold OpenCV Mats, it shows an actual image.

We can easily add more preview renderers by extending `SocketPreviewView` and adding some logic to `SocketPreviewViewFactory`.  For example, we might show a list view or something for sockets containing arrays.

I also added a Python script that returns a couple different pictures of Gompei, which should be useful for testing image stuff until sources are working with the GUI.  This is part of this PR because it's necessary to see that the feature works, but it should also be helpful for testing the OpenCV operations.


src/main/java/edu/wpi/grip/ui/preview/ImageSocketPreviewView.java
```diff
@@ -0,0 +1,85 @@
...
+     * This is also possible to do by using JavaCV, but the JavaCV method involves several intermediate conversions
+     * (Mat -> Frame -> BufferedImage -> JavaFX Image) and is way too slow to use for a real-time video.
+     */
+    private void convertImage() {
+        final Mat mat = this.valueProperty().get();
+        final int width = mat.cols();
+        final int height = mat.rows();
+        final int channels = mat.channels();
+
+        assert channels == 3;
```
JLLeitschuh: Add an assertion message for in case that this fails.

src/main/java/edu/wpi/grip/ui/preview/ImageSocketPreviewView.java
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * Convert a BGR-formatted OpenCV {@link Mat} into a JavaFX {@link Image}.
+     * <p>
+     * JavaFX understands ARGB pixel data, so one way to turn a Mat into a JavaFX image is to shift around the bytes
+     * from the Mat into an int array of pixels.
+     * <p>
+     * This is also possible to do by using JavaCV, but the JavaCV method involves several intermediate conversions
+     * (Mat -> Frame -> BufferedImage -> JavaFX Image) and is way too slow to use for a real-time video.
+     */
+    private void convertImage() {
```
JLLeitschuh: Don't know if you care, but this function is not Reentrant.
Ie, this function is not thread safe.

src/main/java/edu/wpi/grip/ui/preview/SocketPreviewViewFactory.java
```diff
@@ -0,0 +1,26 @@
...
+public class SocketPreviewViewFactory {
+
+    /**
+     * Create an instance of {@link SocketPreviewView} appropriate for the given socket.  Sockets of different types
+     * (like numbers, arrays, images, etc...) are rendered in different ways, and the role of this class is to figure
+     * out what control to use to render a given socket.
+     */
+    @SuppressWarnings("unchecked")
+    public static <T> SocketPreviewView<T> createPreviewView(EventBus eventBus, Socket<T> socket) {
+        if (socket.getSocketHint().getType() == Mat.class) {
```
JLLeitschuh: What if this is something that extends `Mat`?

src/main/java/edu/wpi/grip/ui/preview/ImageSocketPreviewView.java
ThomasJClark: done

src/main/java/edu/wpi/grip/ui/preview/PreviewsView.java
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * @see Socket#isPreviewed()
+ */
+public class PreviewsView extends VBox {
+
+    @FXML
+    private HBox previewBox;
+
+    private final EventBus eventBus;
+    private final List<OutputSocket<?>> previewedSockets;
```
JLLeitschuh: Might be better to make this a map of `OutputSocket` to `SocketPreviewView`
That would allow you to remove the exact objects instead of assuming that the indexes will line up between the two lists.
See below...

src/main/java/edu/wpi/grip/ui/preview/ImageSocketPreviewView.java
ThomasJClark: It has to be called in the application thread anyways (that's way `Platform.runLater` does)

src/main/java/edu/wpi/grip/ui/preview/SocketPreviewViewFactory.java
ThomasJClark: Nothing extends `Mat`

src/main/java/edu/wpi/grip/ui/preview/PreviewsView.java
```diff
@@ -0,0 +1,69 @@
...
+                this.previewedSockets.add(socket);
+                this.previewBox.getChildren().add(SocketPreviewViewFactory.createPreviewView(this.eventBus, socket));
+            }
+        } else {
+            // If the socket was just set as not previewed, remove both it and the corresponding control
+            int index = this.previewedSockets.indexOf(socket);
+            if (index != -1) {
+                this.previewedSockets.remove(index);
+                this.eventBus.unregister(this.previewBox.getChildren().remove(index));
+            }
```
JLLeitschuh: If you use the above suggestion you could make this not have to use the indexes.
I think assuming that the indexes will always line up is a bad idea.

src/main/java/edu/wpi/grip/ui/preview/SocketPreviewView.java
```diff
@@ -0,0 +1,63 @@
...
+        this.socket = socket;
+        this.valueProperty = new SimpleObjectProperty<>(this, "value", socket.getValue());
+
+        this.eventBus.register(this);
+
+        this.setText(this.socket.getSocketHint().getIdentifier());
+        this.getStyleClass().add("socket-preview");
+        this.setCollapsible(false);
+
+        this.setTextAlignment(TextAlignment.CENTER);
```
JLLeitschuh: Can't you do this with css?

JLLeitschuh: http://stackoverflow.com/a/29988060/3708426
```
-fx-text-alignment: center;
```

src/main/java/edu/wpi/grip/ui/preview/SocketPreviewView.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    public Socket<T> getSocket() {
+        return this.socket;
+    }
+
+    /**
+     * The value of the socket being previewed.  Subclasses bind this to some
```
JLLeitschuh: Incomplete sentence.

src/main/java/edu/wpi/grip/ui/preview/TextAreaSocketPreviewView.java
```diff
@@ -0,0 +1,28 @@
...
+public class TextAreaSocketPreviewView<T> extends SocketPreviewView<T> {
+
+    /**
+     * @param eventBus The EventBus used by the application
+     * @param socket   An output socket to preview
+     */
+    public TextAreaSocketPreviewView(EventBus eventBus, Socket<T> socket) {
+        super(eventBus, socket);
+
+        this.setStyle("-fx-pref-width: 20em;");
```
JLLeitschuh: Use css instead?

ThomasJClark: That is css.

src/main/java/edu/wpi/grip/ui/preview/SocketPreviewView.java
ThomasJClark: > Incomplete sentence.

Incomplete sentence.

src/main/java/edu/wpi/grip/ui/preview/SocketPreviewView.java
ThomasJClark: Turns out I can't do that at all.  I'm just gonna remove that line, it's not important that the title is centered.

src/main/java/edu/wpi/grip/ui/preview/ImageSocketPreviewView.java
JLLeitschuh: Okay.

src/main/java/edu/wpi/grip/ui/preview/SocketPreviewView.java
JLLeitschuh: `Subclasses bind this to some`
some what? I don't understand the comment.

src/main/java/edu/wpi/grip/ui/preview/TextAreaSocketPreviewView.java
JLLeitschuh: Use a class in a css file instead of hardcoding it in java.

src/main/java/edu/wpi/grip/ui/controllers/MainWindowController.java
```diff
@@ -72,13 +71,22 @@
...
+    }
+
     @Override
     public void initialize(URL url, ResourceBundle resourceBundle) {
+        PreviewsView previewPaneView = new PreviewsView(eventBus);
+
         PaletteView paletteView = new PaletteView(eventBus);
-        paletteView.operationsProperty().addAll(this.add, this.multiply);
-        this.topPane.getItems().add(paletteView);
+        paletteView.operationsProperty().addAll(this.add, this.multiply, this.gompeiOperation);
```
JLLeitschuh: Bad practice to mutate a list outside of the class that owns the list.

src/main/java/edu/wpi/grip/ui/preview/SocketPreviewView.java
ThomasJClark: I think I accidentally a line there.

src/main/java/edu/wpi/grip/ui/preview/TextAreaSocketPreviewView.java
ThomasJClark: So, the .css file is really for defining the style and layout of the program.  The point of this line is to set the number of characters that the textarea shows, and it uses CSS because it's the only way to specify it in ems instead of pixels.  There's no advantage in clarity or maintainability by putting that information in a .css file along with all of the style information.

PR #34, opened 2015-10-05
Adds OpenCV Bindings Generator

JLLeitschuh:
Adds a generator for an explicit set of opencv function.


.gitignore
```diff
@@ -77,3 +77,6 @@ hs_err_pid*
 #ghpages files
 node_modules
 bower_components
+
+#Generated files while being tested.
```
JLLeitschuh: I need to fix this

PR #36, opened 2015-10-07
Fix previews to work with opencv functions

ThomasJClark:
This fixes several issues with socket previews that came up when
integrating opencv functions

- They should only take `OutputSocket`s
- They should not rely on JavaFX properties to know when sockets change
- Image previews should not attempt to render empty images (we might
  want to make a placeholder image eventually)

### DIFF
--- a/src/main/java/edu/wpi/grip/ui/preview/SocketPreviewView.java
+++ b/src/main/java/edu/wpi/grip/ui/preview/SocketPreviewView.java
@@ -1,14 +1,9 @@
 package edu.wpi.grip.ui.preview;
 
 import com.google.common.eventbus.EventBus;
-import com.google.common.eventbus.Subscribe;
-import edu.wpi.grip.core.Socket;
-import edu.wpi.grip.core.events.SocketChangedEvent;
-import javafx.beans.property.ObjectProperty;
-import javafx.beans.property.SimpleObjectProperty;
+import edu.wpi.grip.core.OutputSocket;
 import javafx.scene.control.TitledPane;
 
-import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 /**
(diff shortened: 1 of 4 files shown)


src/main/java/edu/wpi/grip/ui/preview/SocketPreviewView.java
```diff
@@ -40,22 +32,7 @@ public SocketPreviewView(EventBus eventBus, Socket<T> socket) {
         this.setCollapsible(false);
     }
 
-    public Socket<T> getSocket() {
+    public OutputSocket<T> getSocket() {
```
JLLeitschuh: Change this signature to protected and the name should probably be `getOutputSocket()`

src/main/java/edu/wpi/grip/ui/preview/SocketPreviewView.java
```diff
@@ -17,21 +12,18 @@
...
-    private final Socket<T> socket;
-    private final ObjectProperty<T> valueProperty;
+    private final OutputSocket<T> socket;
 
     /**
      * @param eventBus The EventBus used by the application
      * @param socket   An output socket to preview
      */
-    public SocketPreviewView(EventBus eventBus, Socket<T> socket) {
+    public SocketPreviewView(EventBus eventBus, OutputSocket<T> socket) {
```
JLLeitschuh: You really don't need to pass the output socket to this abstract class.
It really doesn't need it. This class doesn't actually do anything with it.

ThomasJClark: Yes it does - it needs the socket to set the title text.

src/main/java/edu/wpi/grip/ui/preview/SocketPreviewView.java
ThomasJClark: Everywhere else it's called `getSocket()` unless a class needs to distinguish between the two (like `Connection`).  Also, calling it that sort of implies that this outputs something, which it doesn't.

PR #37, opened 2015-10-07
Feat/webcam source

JLLeitschuh:
- Adds WebcamSource
 - Adds an Operation to demo WebcamSource


src/main/java/edu/wpi/grip/core/sources/WebcamSource.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.concurrent.TimeoutException;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class WebcamSource implements Source {
+    private final SocketHint<Mat> imageOutputHint = new SocketHint<Mat>("Image", Mat.class, Mat::new);
+    private final OutputSocket<Mat> outputSocket;
+    final private ScheduledThreadPoolExecutor executor;
+    private OpenCVFrameGrabber grabber;
+    private final int FRAME_RATE = 30;
```
ThomasJClark: I'm not sure we should hard-code 30 fps here.  Are you sure there isn't a way block until a frame is ready?  Then we could just send off frames at whatever rate they come it.

JLLeitschuh: Yea, there is a 'getFPS' method, unfortunately it returns a double and there isn't any documentation on what the units are. I'm going to assume seconds.

src/main/java/edu/wpi/grip/core/sources/WebCamSource.java
```diff
@@ -0,0 +1,188 @@
...
+            this.value = value;
+            this.group = group;
+        }
+    }
+
+    private final SocketHint<Mat> imageOutputHint = new SocketHint<Mat>("Image", Mat.class, Mat::new);
+    private final SocketHint<Number> framerateOutputHint = new SocketHint<Number>("Framerate", Number.class, 0);
+    private OutputSocket<Mat> frameOutputSocket;
+    private OutputSocket<Number> frameRateOutputSocket;
+    private Optional<Thread> executor;
```
ThomasJClark: Why `Optional`?  I think it makes sense to start recording as soon as it's constructed.

src/main/java/edu/wpi/grip/core/sources/WebCamSource.java
```diff
@@ -0,0 +1,188 @@
...
+
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class WebCamSource implements Source {
+
+    public enum SourceType {
```
ThomasJClark: This really isn't necessary, is it?  It looks like you're just copying a JavaCV enum.  Now we have to make sure that our copy always stays in sync with the JavaCV one.

src/main/java/edu/wpi/grip/core/sources/WebCamSource.java
JLLeitschuh: Provide the ability to stop the video feed and start it again later?
You may want to shut off your webcam while you do other work.

src/main/java/edu/wpi/grip/core/sources/WebCamSource.java
JLLeitschuh: It is necessary, somewhere in the application we have to statically list all of the of the devices in order to provide a selection drop down to the user.
I decided that this was better than trying to use reflection to get the list.

src/main/java/edu/wpi/grip/core/sources/WebCamSource.java
```diff
@@ -0,0 +1,131 @@
...
+        }
+        try {
+            grabber.start();
+        } catch (FrameGrabber.Exception e) {
+            throw new IllegalStateException("A problem occurred trying to start the frame grabber", e);
+        }
+        // Store the grabber only once it has been started in the case that there is an exception.
+        this.grabber = Optional.of(grabber);
+
+        final Thread frameExecutor = new Thread(() -> {
```
ThomasJClark: Kind of nitpicky, but do you wanna rename this to `frameThread` since it's not an `Executor` anymore?

src/main/java/edu/wpi/grip/core/sources/WebCamSource.java
```diff
@@ -0,0 +1,131 @@
...
+                } catch (FrameGrabber.Exception e) {
+                    throw new IllegalStateException("Failed to grab image", e);
+                }
+                frameOutputSocket.setValue(convertToMat.convert(videoFrame));
+                long thisMoment = System.currentTimeMillis();
+                frameRateOutputSocket.setValue(1000/(thisMoment - lastFrame));
+                lastFrame = thisMoment;
+            }
+        });
+        frameExecutor.setUncaughtExceptionHandler(
```
ThomasJClark: Since the entire thread is defined above, it would probably be better to catch all exceptions within the thread.

src/main/java/edu/wpi/grip/core/sources/WebCamSource.java
```diff
@@ -0,0 +1,131 @@
...
+                (thread, exception) -> {
+                    System.err.println("WebCam Frame Grabber Thread crashed with uncaught exception:");
+                    exception.printStackTrace();
+                }
+        );
+        frameExecutor.start();
+        executor = Optional.of (frameExecutor);
```
ThomasJClark: Ctrl+Alt+L

src/main/java/edu/wpi/grip/core/sources/WebCamSource.java
```diff
@@ -0,0 +1,131 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Provides a way to generate a constantly updated {@link Mat} from a WebCam attached
+ * to the users computer.
+ */
+public class WebCamSource implements Source {
```
ThomasJClark: *Webcam

src/main/java/edu/wpi/grip/core/sources/WebCamSource.java
JLLeitschuh: As in wrap the entire thread in a try catch block?
Doing that will hide compile time errors for exceptions that need to be caught, right?
If we do it this way then the compiler will complain if you don't handle a catchable exception.

src/main/java/edu/wpi/grip/core/sources/WebCamSource.java
JLLeitschuh: I tried that, Intellij kept complaining about spelling. So I converted it to the other one.
You still want me to change it?

src/main/java/edu/wpi/grip/core/sources/WebCamSource.java
JLLeitschuh: Okay.

src/main/java/edu/wpi/grip/core/sources/WebCamSource.java
ThomasJClark: yeah

src/main/java/edu/wpi/grip/core/sources/WebcamSource.java
```diff
@@ -0,0 +1,137 @@
...
+
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Provides a way to generate a constantly updated {@link Mat} from a Webcam attached
+ * to the users computer.
```
ThomasJClark: user's

PR #39, opened 2015-10-09
Source GUI

ThomasJClark:
Adds GUI code and some core code for letting the user interact with sources.

The GUI components for adding and removing sources are not 100% implemented yet, but the controls for rendering the sources themselves are, as well as the underlying wiring for adding and removing sources.


src/main/java/edu/wpi/grip/ui/pipeline/PipelineView.java
```diff
@@ -120,6 +139,28 @@ private OutputSocketView findOutputSocketView(Socket socket) {
             }
         }
 
+        for (SourceView sourceView : this.getSources()) {
+            for (OutputSocketView socketView : sourceView.getOutputSockets()) {
+                if (socketView.getSocket() == socket) {
+                    return socketView;
+                }
+            }
+        }
```
JLLeitschuh:
```
Stream.concat(this.getSteps().stream().map(stepView -> stepView.getOutputSockets()),
                this.getSources().stream().map(stepView -> stepView.getOutputSockets()))
                .flatMap(l -> l.stream())
                .filter(s -> s.getSocket() == socket)
                .findFirst();
```
Could be done this way. Also, this way this way you aren't returning null.

src/main/java/edu/wpi/grip/ui/pipeline/PipelineView.java
```diff
@@ -120,6 +139,28 @@ private OutputSocketView findOutputSocketView(Socket socket) {
...
+     * @return The {@link SourceView} that corresponds with the given source
+     */
+
+    private SourceView findSourceView(Source source) {
+        for (SourceView sourceView : this.getSources()) {
+            if (sourceView.getSource() == source) {
+                return sourceView;
+            }
+        }
+
```
JLLeitschuh: If you use a stream then you wouldn't return null.

src/main/java/edu/wpi/grip/ui/pipeline/PipelineView.java
```diff
@@ -196,6 +237,16 @@ private void addConnectionView(Connection connection) {
     }
 
     @Subscribe
+    public void onSourceAdded(SourceAddedEvent event) {
+        Platform.runLater(() -> this.sources.getChildren().add(new SourceView(this.eventBus, event.getSource())));
+    }
+
+    @Subscribe
+    public void onSourceRemoved(SourceRemovedEvent event) {
+        Platform.runLater(() -> this.sources.getChildren().remove(findSourceView(event.getSource())));
```
JLLeitschuh: You don't check null here,

PR #40, opened 2015-10-09
Update the source-related GUI components

ThomasJClark:


src/main/java/edu/wpi/grip/core/sources/WebcamSource.java
```diff
@@ -34,12 +35,22 @@
      * Creates a Webcam source that can be used as an input to a pipeline
      * @param eventBus The EventBus to attach to
      */
-    public WebcamSource(EventBus eventBus) {
+    public WebcamSource(EventBus eventBus, int deviceNumber) {
         checkNotNull(eventBus, "Event Bus was null.");
+
+        this.name = "webcam " + deviceNumber;
```
JLLeitschuh: Capitalize? `Webcam`

src/main/java/edu/wpi/grip/ui/pipeline/AddSourceView.java
```diff
@@ -22,19 +35,65 @@
...
+            dialog.getDialogPane().setStyle(root.getStyle());
+            dialog.getDialogPane().getStylesheets().addAll(root.getStylesheets());
+
+            // If the user clicks OK, add a new webcam source
+            dialog.showAndWait().filter(Predicate.isEqual(ButtonType.OK)).ifPresent(result -> {
+                this.loadSourceExecutor.execute(() -> {
+                    final WebcamSource source = new WebcamSource(eventBus, cameraIndex.getValue());
+                    eventBus.post(new SourceAddedEvent(source));
+                });
+            });
```
JLLeitschuh: Can you remove sources yet? If you can then we need to handle having it close and then subsequently stop the webcam.

ThomasJClark: There should probably be `create` and `destroy` methods or something like that in `Source`, since this problem will also affect network sources

JLLeitschuh:
Tested and it works well.

src/main/java/edu/wpi/grip/ui/pipeline/AddSourceView.java
JLLeitschuh: Do you want to fix this in this PR?

src/main/java/edu/wpi/grip/core/sources/ImageFileSource.java
```diff
@@ -16,46 +17,60 @@
...
     /**
-     *
      * @param eventBus The event bus for the pipeline.
+     * @param imageURL The location on the file system where the image exists.
      */
-    public ImageFileSource(EventBus eventBus){
+    public ImageFileSource(EventBus eventBus, URL imageURL) {
         checkNotNull(eventBus, "Event Bus was null.");
+
+        this.name = imageURL.toString().substring(imageURL.toString().lastIndexOf('/') + 1);
```
JLLeitschuh: What if the url doesn't have a slash in it.

src/main/java/edu/wpi/grip/ui/pipeline/AddSourceView.java
```diff
@@ -22,19 +35,65 @@
 
     private final EventBus eventBus;
 
+    /**
+     * Sources typically have to block while they load, so that should be done in a separate daemon thread to avoid
+     * freezing up the GUI.
+     */
+    private final Executor loadSourceExecutor = Executors.newSingleThreadExecutor(runnable -> {
+        final Thread thread = new Thread(runnable);
+        thread.setDaemon(true);
```
JLLeitschuh: Thank you for taking care of this.

PR #42, opened 2015-10-11
Only make sockets publishable if their socket hint says so

ThomasJClark:
This adds a "publishable" field to SocketHint that controls if a socket
can be set to be published to a sink.  It doesn't make sense for a lot
of sockets to be able to be published

Not preventing it would create confusion (mats would likely be published
somewhere as the result of Mat::toString(), which is almost certainly
not was the user things).  Plus, the publish buttons create visual
clutter.

Since the default setting is `false`, we should make sure that anything
that could be published is set as publishable.  This probably means
manually tagging certain opencv output parameters as publishable.

### DIFF
--- a/src/main/java/edu/wpi/grip/core/SocketHint.java
+++ b/src/main/java/edu/wpi/grip/core/SocketHint.java
@@ -18,6 +18,7 @@ public class SocketHint<T> {
     private final Supplier<T> initialValueSupplier;
     private final View view;
     private final T[] domain;
+    private boolean publishable;
 
     /**
      * @param identifier           A user-presentable name for the socket, such as "Blur Radius".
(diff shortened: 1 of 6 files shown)


src/main/java/edu/wpi/grip/core/SocketHint.java
```diff
@@ -27,13 +28,25 @@
      *                             can consist of two elements that correspond to a minimum and maximum value.  The
      *                             property does not make sense for all types and is left unspecified for some.
      * @param initialValueSupplier A function that returns an initial value for newly-created sockets.
+     * @param publishable          If this isn't true, the socket won't be able to be set as published, and the GUI
+     *                             should also not show a button to publish it.  This is false by default.
```
JLLeitschuh: Why is it false by default?

ThomasJClark: Most things probably won't be intended to be publishable, so it's more convenient to not have to use the really long constructor

src/main/java/edu/wpi/grip/core/SocketHint.java
```diff
@@ -56,6 +69,14 @@ public SocketHint(String identifier, Class<T> type, T initialValue) {
         this(identifier, type, () -> initialValue);
     }
 
+    protected void setPublishable(boolean publishable) {
```
JLLeitschuh: I don't see the value in this function. If you want to make a socket publishable then it should be set in the constructor.

ThomasJClark: That's why it's protected. Only a unit test uses it.

JLLeitschuh: Having a method exclusively for unit tests isn't really a good idea. If you REALLY, (this doesn't seem like a valid reason to use reflection) want to run a method in a unit test that should be private use reflection to run it.

Also, the unit test has access to the constructor too.

PR #43, opened 2015-10-11
Make the connection curves slightly nicer

ThomasJClark:
Connections are rendered as a cubic curve.  This change adjusts the
positions of the control points so they are influenced by the positions
of both sockets.  Blender's node editor seems to use the same algorithm.


JLLeitschuh:
Cool!

PR #48, opened 2015-10-12
Make canny work with the GUI

ThomasJClark:
Checkbox controls are added, and the preview pane can now show grayscale
images like the one that canny outputs

### DIFF
--- a/src/main/java/edu/wpi/grip/ui/pipeline/InputSocketView.java
+++ b/src/main/java/edu/wpi/grip/ui/pipeline/InputSocketView.java
@@ -163,10 +163,23 @@ public class InputSocketView extends GridPane implements Initializable {
 
                 break;
             }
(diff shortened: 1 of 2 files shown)


src/main/java/edu/wpi/grip/ui/pipeline/InputSocketView.java
```diff
@@ -163,10 +163,18 @@ public void initialize(URL url, ResourceBundle resourceBundle) {
 
                 break;
             }
+            case CHECKBOX: {
+                CheckBox checkBox = new CheckBox();
```
JLLeitschuh: Check can cast

JLLeitschuh: assign default value

PR #49, opened 2015-10-12
Refactors the enumeration generation code

JLLeitschuh:
Generator can now generate enumerations for any opencv source files

### DIFF
--- a/buildSrc/src/main/java/edu/wpi/gripgenerator/FileParser.java
+++ b/buildSrc/src/main/java/edu/wpi/gripgenerator/FileParser.java
@@ -80,7 +80,7 @@ public class FileParser {
         CompilationUnit compilationUnit = readFile(INPUT_URL);
         Map<String, CompilationUnit> returnMap = new HashMap<>();
         DefaultValueCollector collector = new DefaultValueCollector();
-        collector.add(new PrimitiveDefaultValue(new PrimitiveType(PrimitiveType.Primitive.Double)){
+        collector.add(new PrimitiveDefaultValue(new PrimitiveType(PrimitiveType.Primitive.Double)) {
             @Override
             protected Set<String> getDefaultValues() {
                 return Collections.singleton("CV_PI");
@@ -95,7 +95,7 @@ public class FileParser {
             }
         });
 
-        collector.add(new EnumDefaultValue("edu.wpi.grip.core.operations.opencv.enumeration", "FlipCode", "X_AXIS", "Y_AXIS", "BOTH_AXIS"));
+        collector.add(new EnumDefaultValue("edu.wpi.grip.core.operations.opencv.enumeration", "FlipCode", "X_AXIS", "Y_AXIS", "BOTH_AXES"));
 
         OperationList operationList = new OperationList(
                 new ImportDeclaration(new NameExpr("edu.wpi.grip.generated.opencv_core"), false, true),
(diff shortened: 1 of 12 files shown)


buildSrc/src/main/java/edu/wpi/gripgenerator/FileParser.java
```diff
@@ -111,7 +117,61 @@ private static CompilationUnit readFile(URL url) {
...
                 new DefinedMethod("cornerHarris", false, "Mat", "Mat"),
-                new DefinedMethod("cornerEigenValsAndVecs", false, "Mat", "Mat")
+                new DefinedMethod("cornerEigenValsAndVecs", false, "Mat", "Mat"),
+                new DefinedMethod("threshold", false,
+                        new DefinedParamType("Mat"),
+                        new DefinedParamType("Mat"),
+                        new DefinedParamType("double"),
+                        new DefinedParamType("double"),
+                        new DefinedParamType("int").setLiteralDefaultValue("THRESH_BINARY")
+                ).addDescription("Applies a fixed-level threshold to each array element. " +
```
ThomasJClark: This description is way too long.  Keep it a line or two, and users can refer to the opencv documentation for more information

ThomasJClark:
Not all enumerations work - `flipCode` still shows up as a number.

ThomasJClark:
`dtype` also shows up as a number, although we can safely get rid of that for now

ThomasJClark:
I keep getting unknown interpolation method errors when I use `INTER_MAX`, `WARP_FILL_OUT`, or `WARP_INVERSE_MAP`.  I think it's generating the wrong values for those

JLLeitschuh:
They aren't the wrong values. I can guarantee that. Do they start working when you try to resize bigger instead of smaller?

JLLeitschuh:
Flip code I have to write by hand and add.

JLLeitschuh:
I have to write a rule to ignore dtype.

ThomasJClark:
> They aren't the wrong values. I can guarantee that. Do they start working when you try to resize bigger instead of smaller?

What?  Where are you getting this information?  The opencv documentation I'm reading contradicts this.  `INTER_MAX` is the number of interpolation modes, and `WARP_FILL_OUTLIERS` and `WARP_INVERSE_MAP` are flags.  The generator is treating them as enum values.

ThomasJClark:
> Flip code I have to write by hand and add.

> I have to write a rule to ignore dtype.

Ok, that's fine.  Can you do these in this PR?

JLLeitschuh:
Hmmm.. Perhaps if an enum is used as a flag instead of as an enum then it should be a list of checkboxes:
For example: `warpAffine` uses `INTER_LINEAR` but it also uses the flags.

buildSrc/src/main/java/edu/wpi/gripgenerator/FileParser.java
JLLeitschuh: I think its better to make it long and then hide it with ellipses. (ie `...`)

JLLeitschuh:
Problem:
The input arrays and the output array can all have the same or different depths. For example, you
can subtract to 8-bit unsigned arrays and store the difference in a 16-bit signed array. Depth of
the output array is determined by dtype parameter. In the second and third cases above, as well as
in the first case, when src1.depth() == src2.depth(), dtype can be set to the default -1. In this
case the output array will have the same depth as the input array, be it src1, src2 or both.

What if the two array input are not the same type? Then -1 won't work.

PR #50, opened 2015-10-15
Adds Operations for Creating Points and Sizes and Accessing Fields on Mats

JLLeitschuh:


src/main/java/edu/wpi/grip/core/operations/opencv/MatFieldAccessor.java
```diff
@@ -0,0 +1,77 @@
...
+import edu.wpi.grip.core.OutputSocket;
+import edu.wpi.grip.core.SocketHint;
+import org.bytedeco.javacpp.opencv_core.Mat;
+import org.bytedeco.javacpp.opencv_core.Size;
+
+public class MatFieldAccessor implements CVOperation{
+    private static final Mat defaultsMat = new Mat();
+    private final SocketHint matHint = new SocketHint("input", Mat.class, Mat::new, SocketHint.View.NONE);
+    private final SocketHint sizeHint = new SocketHint("size", Size.class, Size::new);
+    private final SocketHint isNullHint = new SocketHint("is null", Boolean.class, defaultsMat.isNull());
```
ThomasJClark: Mats shouldn't ever be `null`.  If this returns true, I'm pretty sure something in the application would crash (or more likely OpenCV would segfault), so having this as an output is not useful.

src/main/java/edu/wpi/grip/core/operations/opencv/MatFieldAccessor.java
```diff
@@ -0,0 +1,77 @@
...
+    private final SocketHint emptyHint = new SocketHint("empty", Boolean.class, defaultsMat.empty());
+    private final SocketHint channelsHint = new SocketHint("channels", Number.class, defaultsMat.channels());
+    private final SocketHint colsHint = new SocketHint("cols", Number.class, defaultsMat.cols());
+    private final SocketHint rowsHint = new SocketHint("rows", Number.class, defaultsMat.rows());
+    private final SocketHint highValueHint = new SocketHint("high value", Number.class, defaultsMat.highValue());
+    private final SocketHint depthHint = new SocketHint("depth", Number.class, defaultsMat.depth());
+    private final SocketHint typeHint = new SocketHint("type", Number.class, defaultsMat.type());
```
ThomasJClark: This is an enum, not a Number

src/main/java/edu/wpi/grip/core/operations/opencv/MatFieldAccessor.java
JLLeitschuh: Makes sense.

src/main/java/edu/wpi/grip/core/operations/opencv/PointGeneratorOperation.java
```diff
@@ -0,0 +1,45 @@
...
+    private final SocketHint<Point> outputHint = new SocketHint<Point>("point", Point.class, Point::new, SocketHint.View.NONE);
+
+    @Override
+    public String getName() {
+        return "Point";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Creates a Point";
```
ThomasJClark: I would name this "New Point" and word the description as "Create a point".  Same for Size.

src/main/java/edu/wpi/grip/core/operations/opencv/PointGeneratorOperation.java
```diff
@@ -0,0 +1,45 @@
+package edu.wpi.grip.core.operations.opencv;
+
+
+import com.google.common.eventbus.EventBus;
+import edu.wpi.grip.core.InputSocket;
+import edu.wpi.grip.core.OutputSocket;
+import edu.wpi.grip.core.SocketHint;
+import org.bytedeco.javacpp.opencv_core.Point;
+
+public class PointGeneratorOperation implements CVOperation {
```
ThomasJClark: I think `NewPointOperation` would be a better name for this class (and similarly for the Size one)

src/main/java/edu/wpi/grip/core/operations/opencv/SizeGeneratorOperation.java
```diff
@@ -0,0 +1,46 @@
...
+
+import com.google.common.eventbus.EventBus;
+import edu.wpi.grip.core.InputSocket;
+import edu.wpi.grip.core.OutputSocket;
+import edu.wpi.grip.core.SocketHint;
+import org.bytedeco.javacpp.opencv_core.Size;
+
+public class SizeGeneratorOperation implements CVOperation {
+
+    private final SocketHint<Number> widthHint = new SocketHint<Number>("width", Number.class, -1, SocketHint.View.SPINNER, new Integer[]{Integer.MIN_VALUE, Integer.MAX_VALUE}, false);
```
ThomasJClark: Should the minimum value be 0?  I don't think negative sizes make sense, and OpenCV will probably spit back a bunch of errors if we try to to that.

JLLeitschuh: Perhaps the minimum should be -1. The default for a lot of opencv functions is `Point(-1,-1)`

src/main/java/edu/wpi/grip/core/operations/opencv/MatFieldAccessor.java
```diff
@@ -0,0 +1,65 @@
...
+    private final SocketHint emptyHint = new SocketHint("empty", Boolean.class, defaultsMat.empty());
+    private final SocketHint channelsHint = new SocketHint("channels", Number.class, defaultsMat.channels());
+    private final SocketHint colsHint = new SocketHint("cols", Number.class, defaultsMat.cols(), SocketHint.View.NONE, null, true);
+    private final SocketHint rowsHint = new SocketHint("rows", Number.class, defaultsMat.rows(), SocketHint.View.NONE, null, true);
+    private final SocketHint highValueHint = new SocketHint("high value", Number.class, defaultsMat.highValue());
+
+
+    @Override
+    public String getName() {
+        return "Unwrap Mat";
```
JLLeitschuh: Get Mat Info

PR #51, opened 2015-10-16
Adds a Color Picker for OpenCV Scalar Values

JLLeitschuh:


JLLeitschuh:
First, please comment on the diff, not the commit. 
Secondly, there's no way to modify what set value does from the case switch statement, is there?

ThomasJClark:
> Secondly, there's no way to modify what set value does from the case switch statement, is there?

There definitely is - just add a listener to both properties and do the conversion in the listeners.  Most of the cases use `bindBidirectional` because they happen to use the same type for both the JavaFX property and the socket, but if there's some conversion that needs to happen, you just need to add two listeners.

```
final ColorPicker colorPicker = new ColorPicker(scalarToColor((Scalar) this.socket.getValue()));
this.controlPane.getChildren().add(colorPicker);

this.valueProperty.addListener(observable ->
        colorPicker.setValue(scalarToColor((Scalar) this.valueProperty.getValue())));

colorPicker.valueProperty().addListener(observable ->
        this.valueProperty.setValue(colorToScalar(colorPicker.getValue())));
```

As another note, you probably also want to make it work on `Mat`s as well as `Scalar`s.  Mats are sometimes useful for holding colors.

JLLeitschuh:
Okay, I understand the first part. Thanks!

What problems with the built in JavaFX Colorpicker do you forsee? It gives you complete control over all of the values.

The problem with supporting color pickers for Mats is that we also have to define a size the Mat needs to be as well as the color it represents. So thats going to need to be a multi input type.

ThomasJClark:
Mats that hold colors would probably be 3x1.

The JavaFX color picker doesn't let you slide around a color value in real-time, so it's not very useful for selecting a threshold value or something.  You'd have to keep bringing up the dialog, guessing, and checking how close it was.  The dialog also doesn't show up for me at all.

JLLeitschuh:
The color picker does real time updating for me. I don't know why it doesn't work on linux.

ThomasJClark:
The dialog is tiny and shows up behind the main window

JLLeitschuh:
Screenshot? This doesn't happen for me.

ThomasJClark:
> Screenshot?

Do you not believe me, or do you need clarification on I meant?  Either way, here's the screenshot.
screenshot

JLLeitschuh:
I needed clarification.

JLLeitschuh:
This seems to be an open bug: https://bugs.openjdk.java.net/browse/JDK-8088478

src/main/java/edu/wpi/grip/ui/pipeline/input/ColorInputSocketView.java
```diff
@@ -0,0 +1,70 @@
...
+import javafx.scene.paint.Color;
+import org.bytedeco.javacpp.opencv_core.Scalar;
+
+/**
+ * An {@link InputSocketView} that lets the user set the value of a {@link Scalar} with a {@link ColorPicker}
+ */
+public class ColorInputSocketView extends InputSocketView<Scalar> {
+
+    private final ColorPicker colorPicker;
+    private final ChangeListener<Color> updateSocketFromColor;
```
JLLeitschuh: Codacy Issue found: [Perhaps 'updateSocketFromColor' could be replaced by a local variable.](https://www.codacy.com/app/WPIRoboticsProjects/GRIP/file/1275184114/issues/source?bid=2943107&fileBranchId=2943116#l19)

ThomasJClark:
- Choosing "FILLED" for lineType results in the step crashing
- AFAIK, there's nothing useful you can do in GRIP that could be connected to `CV Rectangle`.  What exactly is the usecase for this?
- Custom Color still doesn't work.  The dialog shows up behind the window, and is tiny on my screen.

JLLeitschuh:
The dialog looks fine in that screenshot.

ThomasJClark:
No it doesn't

JLLeitschuh:
Just fixed it. Try it now

JLLeitschuh:
> AFAIK, there's nothing useful you can do in GRIP that could be connected to CV Rectangle. What exactly is the usecase for this?

Here is the use case: #485

ThomasJClark:
Jonathan, can you explain what the use case is?  There's almost nothing you can do with this PR besides draw a static rectangle.

ghead:
As for a use case for a static rectangle (which can't be drawn unless this issue is fixed), in past years we have drawn on the smartdashboard to help with manual aiming when auto is not working for some reason.  A static rectangle drawn on the output from GRIP would serve as an aiming assist - align the goal with the static rectangle and shoot.

AustinShalit:

JLLeitschuh:
Yea, this one worked fine on all but fedora.
:( Its cool to use... But I don't know how practical it is.

AustinShalit:
What was the issue on fedora?

JLLeitschuh:
Something about the color picker UI showing up way too small and it wasn't on the top level.

For some reason it showed up behind the GRIP UI.

AustinShalit:
Just add it and open an issue?

SamCarlberg:
That would mean the app's broken on certain systems. I'd prefer to avoid that. We may have to make our own color picker.

PR #52, opened 2015-10-17
Fix paths for ImageFileSource to be OS-independant

ThomasJClark:
Instead of using a java URL object, ImageFileSource should just take a
File and convert it to the right path format using the Paths class.

I tested this on a Windows machine, and it works.


JLLeitschuh:
You need to fix the failure.

PR #54, opened 2015-10-23
Show operation names in previews

ThomasJClark:
If you have a bunch of previews called "dst", it's helpful to be able to
tell what they are the output of.

### DIFF
--- a/src/main/java/edu/wpi/grip/ui/preview/SocketPreviewView.java
+++ b/src/main/java/edu/wpi/grip/ui/preview/SocketPreviewView.java
@@ -27,9 +27,25 @@ public abstract class SocketPreviewView<T> extends TitledPane {
 
         this.eventBus.register(this);
 
-        this.setText(this.socket.getSocketHint().getIdentifier());
+        this.setText(this.getTitle());
         this.getStyleClass().add("socket-preview");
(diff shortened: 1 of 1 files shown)


src/main/java/edu/wpi/grip/ui/preview/SocketPreviewView.java
```diff
@@ -27,11 +27,27 @@ public SocketPreviewView(EventBus eventBus, OutputSocket<T> socket) {
...
         this.getStyleClass().add("socket-preview");
         this.setCollapsible(false);
     }
 
+    /**
+     * @return A string of text to display in the GUI to identify this preview, including the name of the output socket
+     * and what operation it is the output of.
+     */
+    private String getTitle() {
+        String title = "";
```
JLLeitschuh: StringBuilder is faster.

JLLeitschuh: Okay actually, the compiler will turn it into a string builder.

ThomasJClark: It doesn't matter. They both do the same thing and this way is more readable.

PR #57, opened 2015-10-28
Adds JavaFX Testing to Build

JLLeitschuh:
JavaFX tests should now run as part of the build.
Tests should work headless on Travis.

### DIFF
--- a/src/test/java/edu/wpi/grip/ui/controllers/MainWindowControllerTest.java
+++ b/src/test/java/edu/wpi/grip/ui/controllers/MainWindowControllerTest.java
@@ -1,0 +1,30 @@
+package edu.wpi.grip.ui.controllers;
+
+
+import edu.wpi.grip.Main;
+import javafx.stage.Stage;
+import org.junit.Test;
+import org.testfx.framework.junit.ApplicationTest;
+import org.testfx.matcher.base.NodeMatchers;
+
+import static org.testfx.api.FxAssert.verifyThat;
+
+
+public class MainWindowControllerTest extends ApplicationTest {
+
+    @Override
+    public void start(Stage stage) throws Exception {
+        new Main().start(stage);
+    }
+
+    @Test
+    public void testShouldCreateNewOperationInPipelineView() {
+        // Given:
+        clickOn("#python-multiply-operation");
+
+        // Then:
+        verifyThat(".pipeline", NodeMatchers.hasChildren(1, ".python-multiply-step"));
+    }
+
+
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,9 @@ addons:
       - g++-4.8
       - oracle-java8-installer
 
+before_install:
+  - export DISPLAY=:99.0
+  - sh -e /etc/init.d/xvfb start
 
 install: ./gradlew assemble --stacktrace
 
--- a/build.gradle
+++ b/build.gradle
@@ -15,6 +15,7 @@ sourceSets{
 
 repositories {
     mavenCentral()
+    jcenter()
 }
 
 dependencies {
@@ -30,7 +31,10 @@ dependencies {
     compile group: 'org.bytedeco.javacpp-presets', name: 'opencv', version: '3.0.0-1.0', classifier: "macosx-x86_64"
     compile group: 'org.bytedeco.javacpp-presets', name: 'opencv', version: '3.0.0-1.0', classifier: "windows-x86_64"
     compile group: 'org.bytedeco.javacpp-presets', name: 'opencv', version: '3.0.0-1.0', classifier: "windows-x86"
-    testCompile group: 'junit', name: 'junit', version: '4.11'
+    testCompile group: 'org.hamcrest', name: 'hamcrest-all', version: '1.3'
+    testCompile group: 'junit', name: 'junit', version: '4.12'
+    testCompile group: 'org.testfx', name: 'testfx-core', version: '4.0.+'
+    testCompile group: 'org.testfx', name: 'testfx-junit', version: '4.0.+'
 }
 
 compileJava.source sourceSets.generated.java, sourceSets.main.java
--- a/src/main/java/edu/wpi/grip/ui/OperationView.java
+++ b/src/main/java/edu/wpi/grip/ui/OperationView.java
@@ -4,6 +4,7 @@ import com.google.common.eventbus.EventBus;
 import edu.wpi.grip.core.Operation;
 import edu.wpi.grip.core.Step;
 import edu.wpi.grip.core.events.StepAddedEvent;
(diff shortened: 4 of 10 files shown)


src/main/java/edu/wpi/grip/core/Operation.java
```diff
@@ -22,6 +22,10 @@
      */
     String getDescription();
 
+    default String getCSSName(){
```
ThomasJClark: Operation should not have anything to do with CSS, since it's not in the gui package.

JLLeitschuh: Where should this method reside then?

JLLeitschuh:
22%! Jeez! That's much higher than I expected.

src/test/java/edu/wpi/grip/ui/controllers/MainWindowControllerTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+import static org.testfx.api.FxAssert.verifyThat;
+
+
+public class MainWindowControllerTest extends ApplicationTest {
+
+    @Override
+    public void start(Stage stage) throws Exception {
+        new Main().start(stage);
+        sleep(5, TimeUnit.SECONDS);
```
ThomasJClark: Why sleep for 5 seconds?

src/main/java/edu/wpi/grip/core/Operation.java
JLLeitschuh: Or I could rename it to `getHyphenatedName`

src/test/java/edu/wpi/grip/ui/controllers/MainWindowControllerTest.java
JLLeitschuh: - [ ] Ooops, left over from testing. Will remove.

src/main/java/edu/wpi/grip/core/Operation.java
ThomasJClark: I would not put this method in `Operation` at all.

JLLeitschuh: I need the method somewhere. Where would it be best located?

src/main/java/edu/wpi/grip/core/Operation.java
```diff
@@ -22,6 +22,10 @@
      */
     String getDescription();
 
+    default String getHyphenatedName(){
```
JLLeitschuh: If you want this somewhere else please offer a proposition as to where it should reside.

ThomasJClark: `StepView` would make sense since it's the only place that uses this to set a CSS class.

It looks like you also use the same value to set the ID in `OperationView`, so you'll also need some way to compute that.  I don't see any reason why this needs to be the same value as a CSS class used in `StepView`, so this can be a separate thing.

JLLeitschuh: I mean in all practicality the name and the description are purely UI elements as well. The are never actually used inside of the pipeline.

src/main/java/edu/wpi/grip/ui/util/CSSElementUtility.java
```diff
@@ -0,0 +1,59 @@
+package edu.wpi.grip.ui.util;
+
+import edu.wpi.grip.core.Connection;
+import edu.wpi.grip.core.Operation;
+import edu.wpi.grip.core.Step;
+
+public final class CSSElementUtility {
```
ThomasJClark: I like that this isn't in a random default method in the core package anymore, but there has to be a better name for this class.  Something that makes it clear that this is for looking up elements in unit tests.  Also documentation that makes that clear.

src/main/java/edu/wpi/grip/ui/util/CSSElementUtility.java
```diff
@@ -0,0 +1,59 @@
...
+
+    private CSSElementUtility(){}
+
+
+    /**
+     * Returns the CSS id name for an operation
+     * @param operation The operation to get the class name for
+     * @return The CSS id for the operation. To use as a css selector then prepend the string with a '#'
+     */
+    public static String idNameFor(Operation operation){
```
ThomasJClark: What does this have to do with CSS?

src/main/java/edu/wpi/grip/ui/util/CSSElementUtility.java
```diff
@@ -0,0 +1,59 @@
...
+    private static StringBuilder nameFor(Operation operation){
+        return shortNameFor(operation).append("-operation");
+    }
+
+    /**
+     * Creates the base short name for the operation.
+     * @param operation The operation.
+     * @return The short base of the operation's name.
+     */
+    private static StringBuilder shortNameFor(Operation operation){
```
ThomasJClark: The documentation says literally nothing about this method - it's just a rephrasing of the method name and signature.

I don't see any use for it, and since there's no justification in the documentation, I think it should probably not exist.

src/main/java/edu/wpi/grip/ui/util/CSSElementUtility.java
```diff
@@ -0,0 +1,59 @@
...
+                + "-to-" +
+                connection.getInputSocket().getSocketHint().getIdentifier();
+    }
+
+    /**
+     * Creates the base name for the operation.
+     * @param operation The operation to generate the name for.
+     * @return The base of the operation's name.
+     */
+    private static StringBuilder nameFor(Operation operation){
```
ThomasJClark: Why obfuscate this by splitting up into a bunch of different methods that pass around a `StringBuilder`?  I would just write the above methods using string concatenation.

src/main/java/edu/wpi/grip/ui/util/CSSElementUtility.java
JLLeitschuh: ID vs Class
The id is assigned as the elements ID param instead of assigning it as a class on the element.

src/main/java/edu/wpi/grip/ui/util/CSSElementUtility.java
JLLeitschuh: I had another method that used it before but removed it.

src/main/java/edu/wpi/grip/ui/util/CSSElementUtility.java
JLLeitschuh: Do you have a suggested name?

src/main/java/edu/wpi/grip/ui/util/CSSElementUtility.java
JLLeitschuh: Its private so it doesn't really matter anyways. I'll remove it.

PR #58, opened 2015-10-28
Refactor InputSocketView into an abstract class

ThomasJClark:
`InputSocketView`s are now constructed by `InputSocketViewFactory`, which reads the `SocketHint` to determine what subclass to use.  This will allow us to add more advanced controls by writing new classes instead of adding new switch cases.

This also fixes a few bugs where certain control views didn't work in all cases. For example, if a spinner is updated through a connection, the number in the spinner didn't actually change before.


src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketViewFactory.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.List;
+
+/**
+ * Factory for constructing editable controls for input sockets
+ */
+public class InputSocketViewFactory {
+
+    /**
+     * Create an instance of {@link InputSocketView} appropriate for the given socket.
+     */
```
JLLeitschuh: Javadoc comment invalid

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketView.java
```diff
@@ -0,0 +1,99 @@
...
+
+    public Node getContent() {
+        return this.contentProperty().get();
+    }
+
+    public ObjectProperty<Node> contentProperty() {
+        if (this.contentProperty == null) {
+            this.contentProperty = new SimpleObjectProperty<>(this, "content");
+            this.contentProperty.addListener(o -> this.contentPane.getChildren().setAll(this.getContent()));
+        }
```
JLLeitschuh: Why is this lazy loaded? Add a comment explaining why it's lazy loaded.

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketViewFactory.java
ThomasJClark: Doesn't really matter, does it? A lot of methods we have don't have a tag for every parameter because it's better to write comments that are helpful to people than ones that blindly follow rules.

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketViewFactory.java
```diff
@@ -0,0 +1,64 @@
...
+                if (socketHint.getType().equals(Boolean.class)) {
+                    return (InputSocketView<T>) new CheckboxInputSocketView(eventBus, (InputSocket<Boolean>) socket);
+                } else {
+                    throw new RuntimeException("Could not create view for socket.  CHECKBOX views must be Booleans. "
+                            + socket.toString());
+                }
+        }
+
+        throw new RuntimeException("Could not create view for socket. " + socket.toString());
+    }
```
JLLeitschuh: Illegal state exceptions are better.

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketViewFactory.java
```diff
@@ -0,0 +1,64 @@
...
+            case SELECT:
+                return new SelectInputSocketView<>(eventBus, socket);
+
+            case CHECKBOX:
+                if (socketHint.getType().equals(Boolean.class)) {
+                    return (InputSocketView<T>) new CheckboxInputSocketView(eventBus, (InputSocket<Boolean>) socket);
+                } else {
+                    throw new RuntimeException("Could not create view for socket.  CHECKBOX views must be Booleans. "
+                            + socket.toString());
+                }
```
JLLeitschuh: Default is more verbose.

src/main/java/edu/wpi/grip/ui/pipeline/input/ListSpinnerInputSocketView.java
```diff
@@ -0,0 +1,61 @@
...
+ * An {@link InputSocketView} that lets the user set the value with a {@link Spinner}
+ */
+public class ListSpinnerInputSocketView<T> extends InputSocketView<T> {
+
+    private final SpinnerValueFactory<T> valueFactory;
+    private final InvalidationListener updateSocketFromSpinner;
+
+    /**
+     * @param socket an input socket where the domain contains all of the possible values to choose from
+     */
```
JLLeitschuh: Invalid javadoc.

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketView.java
ThomasJClark: Lazy loaded? Assuming you mean "lazily evaluated" (if not correct me), all JavaFX properties are written like this. I can't think of a useful comment about it - I think it's clear what this means.

src/main/resources/edu/wpi/grip/ui/pipeline/input/InputSocket.fxml
```diff
@@ -2,17 +2,13 @@
...
         <Label fx:id="identifier" styleClass="socket-identifier" GridPane.rowIndex="0"
                GridPane.columnIndex="1"/>
-        <StackPane fx:id="controlPane" styleClass="socket-control" GridPane.rowIndex="1"
+        <StackPane fx:id="contentPane" styleClass="socket-control" GridPane.rowIndex="1"
                    GridPane.columnIndex="1"/>
     </children>
-    <stylesheets>
-        <URL value="@../roboto/Roboto.css"/>
-        <URL value="@../GRIP.css"/>
-    </stylesheets>
```
JLLeitschuh: Just curious why this is no longer necessary.

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketViewFactory.java
```diff
@@ -0,0 +1,64 @@
...
+            case SLIDER:
+                if (socketHint.getType().equals(Number.class)) {
+                    return (InputSocketView<T>) new SliderInputSocketView(eventBus, (InputSocket<Number>) socket);
+                } else {
+                    throw new RuntimeException("Could not create view for socket.  SLIDER views must be Numbers. "
+                            + socket.toString());
+                }
+
+            case RANGE:
+                if (socketHint.getType().equals(List.class)) {
```
JLLeitschuh: What if this is a java primitive array?

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketView.java
```diff
@@ -0,0 +1,99 @@
...
+
+        return this.contentProperty;
+    }
+
+    public void setContent(Node node) {
```
JLLeitschuh: Protected?

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketView.java
```diff
@@ -0,0 +1,99 @@
...
+        this.contentProperty().set(node);
+    }
+
+    public Label getIdentifier() {
```
JLLeitschuh: protected?

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketView.java
```diff
@@ -0,0 +1,99 @@
...
+
+    public Socket<T> getSocket() {
+        return this.socket;
+    }
+
+    public SocketHandleView getHandle() {
+        return this.handle;
+    }
+
+    public Node getContent() {
```
JLLeitschuh: Private?

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketViewFactory.java
ThomasJClark: There needs to be a separate throw statement inside each case so we can say why it failed (ie: "CHECKBOX views must be Booleans")

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketViewFactory.java
ThomasJClark: Why?

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketView.java
```diff
@@ -0,0 +1,99 @@
...
+
+        this.identifier.setText(socket.getSocketHint().getIdentifier());
+
+        this.handle = new SocketHandleView(this.eventBus, this.socket);
+        this.add(this.handle, 0, 0);
+
+        this.eventBus.register(this);
+    }
+
+    public Socket<T> getSocket() {
```
JLLeitschuh: Protected?

src/main/resources/edu/wpi/grip/ui/pipeline/input/InputSocket.fxml
ThomasJClark: It was never necessary, I just included it in most .fxml files because I copied and pasted MainWindow.fxml as a template.

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketViewFactory.java
ThomasJClark: Then it will throw a `RuntimeException` because it is not a list.

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketView.java
```diff
@@ -0,0 +1,99 @@
...
+    @FXML
+    private Label identifier;
+
+    @FXML
+    private StackPane contentPane;
+
+    private ObjectProperty<Node> contentProperty;
+
+    /**
+     * The "handle" is a simple shape next ot the socket identifier that shows weather or not there is a connection
```
JLLeitschuh: Spelling on weather.

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketView.java
```diff
@@ -0,0 +1,99 @@
...
+        this.add(this.handle, 0, 0);
+
+        this.eventBus.register(this);
+    }
+
+    public Socket<T> getSocket() {
+        return this.socket;
+    }
+
+    public SocketHandleView getHandle() {
```
JLLeitschuh: Protected?

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketView.java
```diff
@@ -0,0 +1,99 @@
...
+     * The "handle" is a simple shape next ot the socket identifier that shows weather or not there is a connection
+     * going to or from the socket.  If there is such a connection, the ConnectionView is rendered as a curve going
+     * from one handle to another.
+     */
+    private SocketHandleView handle;
+
+    private final EventBus eventBus;
+    private final InputSocket<T> socket;
+
+    protected InputSocketView(EventBus eventBus, InputSocket<T> socket) {
```
JLLeitschuh: Javadocs?

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketView.java
ThomasJClark: No, protected prevents classes outside of this package from accessing this method.  The correct modifier here is "public".

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketView.java
ThomasJClark: Technically this still compiles if I make that private, but the way it currently is follows the JavaFX property conventions.

src/main/java/edu/wpi/grip/ui/pipeline/input/NumberSpinnerInputSocketView.java
```diff
@@ -0,0 +1,66 @@
...
+        final double max = ((Number) domain[1]).doubleValue();
+        final double initialValue = socket.getValue().doubleValue();
+
+        this.valueFactory = new SpinnerValueFactory.DoubleSpinnerValueFactory(min, max, initialValue);
+        this.updateSocketFromSpinner = o -> this.getSocket().setValue(this.valueFactory.getValue());
+        this.valueFactory.valueProperty().addListener(this.updateSocketFromSpinner);
+
+        final Spinner<Double> spinner = new Spinner<>(this.valueFactory);
+        spinner.setEditable(true);
+        spinner.disableProperty().bind(this.getHandle().connectedProperty());
```
JLLeitschuh: You do this in every class. Why not wrap this in a method in the parent class and simply pass it the spinner.

src/main/java/edu/wpi/grip/ui/pipeline/input/NumberSpinnerInputSocketView.java
```diff
@@ -0,0 +1,66 @@
...
+import javafx.scene.control.SpinnerValueFactory;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+/**
+ * An {@link InputSocketView} that lets the user set the value of a number with a {@link Spinner}
+ */
+public class NumberSpinnerInputSocketView extends InputSocketView<Number> {
+
+    private final static Object[] DEFAULT_DOMAIN = new Object[]{Double.MIN_VALUE, Double.MAX_VALUE};
```
JLLeitschuh: Do you really mean `-Double.MAX_VALUE`?

ThomasJClark: Yes

src/main/java/edu/wpi/grip/ui/pipeline/input/NumberSpinnerInputSocketView.java
```diff
@@ -0,0 +1,66 @@
...
+        super(eventBus, socket);
+
+
+        Object[] domain = socket.getSocketHint().getDomain();
+        if (domain == null) {
+            domain = DEFAULT_DOMAIN;
+        }
+
+        checkArgument(domain.length == 2 && domain[0] instanceof Number && domain[1] instanceof Number,
+                "Spinners must have a domain with two numbers (min and max)");
```
JLLeitschuh: Perhaps this sort validation should be done inside of the constructor for the SocketHints. That way the we will get the error in the unit tests because all of the operations get instantiated for them but only some of the operations will get this check as it is only ensured in the UI code.

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketView.java
JLLeitschuh: Why isn't it instantiated in the constructor?

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketViewFactory.java
JLLeitschuh: More verbose. RuntimeExecptions are bad practice to throw. You should always try to throw a subclass of it.
Or maybe its an IllegalArgument.

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketViewFactory.java
JLLeitschuh: Yes? But there should be a default case in a case switch statement. That final default case should throw the exception not the way it is now. Sorry, I seem to have put this comment on the wrong line.

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketViewFactory.java
JLLeitschuh: Thats expected behavior then, okay, just checking.

src/main/java/edu/wpi/grip/ui/pipeline/input/NumberSpinnerInputSocketView.java
JLLeitschuh: Yes what?

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketViewFactory.java
```diff
@@ -0,0 +1,64 @@
...
+            case CHECKBOX:
+                if (socketHint.getType().equals(Boolean.class)) {
+                    return (InputSocketView<T>) new CheckboxInputSocketView(eventBus, (InputSocket<Boolean>) socket);
+                } else {
+                    throw new RuntimeException("Could not create view for socket.  CHECKBOX views must be Booleans. "
+                            + socket.toString());
+                }
+        }
+
+        throw new RuntimeException("Could not create view for socket. " + socket.toString());
```
JLLeitschuh: This should be in a default case instead of where it is right now. Sorry, this is where I put my comment on the wrong line.

src/main/resources/edu/wpi/grip/ui/pipeline/input/InputSocket.fxml
JLLeitschuh: Okay.

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketViewFactory.java
JLLeitschuh: Fine.

JLLeitschuh:
If this is done on top of the TestFX commit then you can also add tests to make sure that these work correctly too.

JLLeitschuh:
Also, you need to rebase or merge this with master

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketViewFactory.java
ThomasJClark: k

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketViewFactory.java
ThomasJClark: k

src/main/java/edu/wpi/grip/ui/pipeline/input/NumberSpinnerInputSocketView.java
ThomasJClark: Because it's applied to a different control in each subclass, and there's no way for the parent class to get that control (adding a way to do that would probably be considered breaking encapsulation).  

Using `Property` objects to bind the state to each subclass seemed like an appropriate way to nicely wrap this sort of thing.

src/main/java/edu/wpi/grip/ui/pipeline/input/NumberSpinnerInputSocketView.java
ThomasJClark: I don't think the core code should know the capabilities of each JavaFX control.

src/main/java/edu/wpi/grip/ui/pipeline/input/NumberSpinnerInputSocketView.java
JLLeitschuh: In parent class:
```
protected void attachDisable(Node node){
    node.disableProperty().bind(getHandle().connectedProperty());
}
```

src/main/java/edu/wpi/grip/ui/pipeline/input/NumberSpinnerInputSocketView.java
JLLeitschuh: The socket hint is only for the UI. That's the only reason it exists.

src/main/java/edu/wpi/grip/ui/pipeline/input/NumberSpinnerInputSocketView.java
```diff
@@ -0,0 +1,66 @@
...
+import javafx.scene.control.SpinnerValueFactory;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+/**
+ * An {@link InputSocketView} that lets the user set the value of a number with a {@link Spinner}
+ */
+public class NumberSpinnerInputSocketView extends InputSocketView<Number> {
+
+    private final static Object[] DEFAULT_DOMAIN = new Object[]{-Double.MAX_VALUE, Double.MAX_VALUE};
```
JLLeitschuh: So currently the default domain doesn't include 0.

src/main/java/edu/wpi/grip/ui/pipeline/input/NumberSpinnerInputSocketView.java
ThomasJClark: Properties are how you bind things in JavaFX.  One of the points of having a standard property API is that I don't have to write an `attachDisable()` method every time I want to be able to bind something.

src/main/java/edu/wpi/grip/ui/pipeline/input/NumberSpinnerInputSocketView.java
ThomasJClark: But it's a hint. It shouldn't contain GUI logic.

JLLeitschuh: Its not GUI logic, this is model logic. GUI logic would be modifying some GUI element.
This check is determining if the SocketHint is valid to create a view which is what we want.
It's determining the validity of a socket hint. That should be a sask left left to the hint. Better pattern is to have a class validate itself instead of having something external validate it.

This would make it far easier to spot problems before they occur.

src/main/java/edu/wpi/grip/ui/pipeline/input/NumberSpinnerInputSocketView.java
JLLeitschuh: Whatever, thats fine. My thoughts were that if you need to attach a disable to every single one of the InputSocketViews there should be a unified way of doing it.

src/main/java/edu/wpi/grip/ui/pipeline/input/NumberSpinnerInputSocketView.java
ThomasJClark: One way or another, this switch statement needs to exist in the gui package. There is absolutely no way it can be anywhere else because it needs to create JavaFX controls.   

And the default case and the else statements that are invalid need to do something, and throwing an exception seems reasonable. (now that I think about it, using an assert might be better, but I won't try to justify that right now) 

Since these conditionals that throw exceptions already do the validation in this class, there's no need to do the same validation in another class. It would just be copying and pasting this same switch statement into SocketHint. 

The reason why the validation is done here is because the factory is the class capable of making sockethints to views, so it should be the one to validate if a certain mapping is possible. A socket is not invalid because we can't make a JavaFX control for it, and like you said, a class should validate itself. 

Although this validation was originally supposed to be in Inputsocketview, this was really the intention with socket hints from the beginning, not the hacked-together solution you seem to think.

ThomasJClark: Ignore the ordered list, I'm not sure how I did that and I don't think I can edit comments on mobile.

JLLeitschuh: I want to preface this with a "I'm not adamantly against this pull request, I'm looking to have a discussion about how to do this right." As to your above points: 
1. I agree.
2. Whatever
3. I would prefer to see a `public void validate()` method on SocketHints. Not copy and paste.
4. A SocketHint is invalid if it prevents the creation of an step in the pipeline. This would prevent a step from being created.
5. I'm not arguing to keep validation in the InputSocketView.

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketView.java
JLLeitschuh: You still haven't answered this question. What I really mean is "Lazy instantiation".

src/main/java/edu/wpi/grip/ui/pipeline/input/NumberSpinnerInputSocketView.java
JLLeitschuh: What I'm trying to avoid is having operations with invalid socket hints in our release application.
Perhaps a better way of handling this problem is to have a UI test that adds every single one of our operations to the pipeline and ensures that no errors are thrown.

If you think this is a better plan I'll do that instead.

ThomasJClark: I agree. If we want to verify that the gui works with all of the operations we have, gui unit tests seen appropriate.

src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketView.java
ThomasJClark: I don't know, that's just how javafx properties are typically done.

PR #62, opened 2015-10-31
Adds ExceptionView for uncaught exeption handling

JLLeitschuh:


ThomasJClark:
The dialog is way too small.  See [AddSourceView.java](https://github.com/WPIRoboticsProjects/GRIP/blob/master/src/main/java/edu/wpi/grip/ui/pipeline/AddSourceView.java#L82) for a way to fix that.

untitled

ThomasJClark:
Ctrl+Alt+L

src/main/java/edu/wpi/grip/Main.java
```diff
@@ -14,6 +15,16 @@ public static void main(String[] args) {
 
     @Override
     public void start(Stage stage) throws Exception {
+
+
+        Thread.currentThread().setUncaughtExceptionHandler((thread, throwable) -> {
+            System.out.println("Handler caught exception: "+throwable.getMessage());
```
ThomasJClark: This println is probably unnecessary

src/main/java/edu/wpi/grip/ui/ExceptionView.java
```diff
@@ -0,0 +1,191 @@
...
+    private final ButtonType copyToClipboardBtnType = new ButtonType("Copy To Clipboard");
+    private final ButtonType openGitHubIssuesBtnType = new ButtonType("Open GitHub Issues");
+
+    /**
+     * @see <a href="http://code.makery.ch/blog/javafx-dialogs-official/">Inspiration</a>
+     * @param throwable The throwable exception to display this alert for.
+     * @param services The host services, allows access to {@link HostServices#showDocument(String)} in order to display
+     *                 the issue website.
+     */
+    public ExceptionView(final Throwable throwable, final HostServices services) {
```
ThomasJClark: Is there a reason you didn't use fxml for this?  I think that having hundreds of lines of java code to set styles will just make maintaining the gui a lot harder

src/main/java/edu/wpi/grip/ui/ExceptionView.java
```diff
@@ -0,0 +1,191 @@
...
+     * @return The markdown for the exception.
+     */
+    private String generateExceptionMessage(Throwable throwable){
+        final StringWriter sw = new StringWriter();
+        final PrintWriter pw = new PrintWriter(sw);
+        throwable.printStackTrace(pw);
+
+        return new StringBuilder(sw.toString()
+                /* Allow users to maintain anonymity */
+                .replace(System.getProperty("user.home"), "$HOME").replace(System.getProperty("user.name"),"$USER"))
```
ThomasJClark: I don't think this is really necessary.  If someone contacts us with an issue, they must have a github account or an email, so we would know who they are.  If they use only psuedonyms on github, they can probably remove their username manually.

src/main/java/edu/wpi/grip/ui/ExceptionView.java
```diff
@@ -0,0 +1,191 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Displays an alert with an exception and provides a place for a user to supply information about what caused the error.
+ * Includes a textbox with formatted markdown to allow the issue to be pasted into GitHub easily.
+ * Also, provides links with quick access to the githup issue page.
+ */
+public final class ExceptionView extends Alert {
+    private static final String PROJECT_ISSUE_LINK = "https://github.com/WPIRoboticsProjects/GRIP/issues";
+    private static final String ISSUE_PROMPT_QUESTION = "What were the actions that caused this error occurred?";
```
ThomasJClark: grammar

src/main/java/edu/wpi/grip/ui/ExceptionView.java
```diff
@@ -0,0 +1,191 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Displays an alert with an exception and provides a place for a user to supply information about what caused the error.
+ * Includes a textbox with formatted markdown to allow the issue to be pasted into GitHub easily.
+ * Also, provides links with quick access to the githup issue page.
+ */
+public final class ExceptionView extends Alert {
+    private static final String PROJECT_ISSUE_LINK = "https://github.com/WPIRoboticsProjects/GRIP/issues";
```
ThomasJClark: Having this link seems like a good idea.  I could probably be "https://github.com/WPIRoboticsProjects/GRIP/issues/new" to take the user right to the new issue page

src/main/java/edu/wpi/grip/ui/ExceptionView.java
```diff
@@ -0,0 +1,191 @@
...
+        expContent.add(issuePasteLabel, 0, 0);
+        expContent.add(issueText, 0, 1);
+
+        // Set expandable Exception into the dialog pane.
+        this.getDialogPane().setExpandableContent(expContent);
+        this.getDialogPane().setExpanded(true);
+
+
+        // Prevent these two buttons from causing the alert to close
+        final Button copyToClipboardBtn = (Button) this.getDialogPane().lookupButton(copyToClipboardBtnType);
```
ThomasJClark: TextArea lets you right click and copy, and our users will certainly be computer literate enough to know that

src/main/java/edu/wpi/grip/ui/ExceptionView.java
```diff
@@ -0,0 +1,191 @@
...
+/**
+ * Displays an alert with an exception and provides a place for a user to supply information about what caused the error.
+ * Includes a textbox with formatted markdown to allow the issue to be pasted into GitHub easily.
+ * Also, provides links with quick access to the githup issue page.
+ */
+public final class ExceptionView extends Alert {
+    private static final String PROJECT_ISSUE_LINK = "https://github.com/WPIRoboticsProjects/GRIP/issues";
+    private static final String ISSUE_PROMPT_QUESTION = "What were the actions that caused this error occurred?";
+
+    private static final String systemOptions[] = {
```
ThomasJClark: I'm not sure where this list came from, but I think the properties that might be useful are `os.name`, `os.version`, `os.arch`, `java.vm.name`, `java.version`, and `javafx.version`.

ThomasJClark:
The "OK" button should probably say something like "ignore".  "OK" seems to imply that it will do something, when really it just hides the dialog so the user can't get the stacktrace.

src/main/java/edu/wpi/grip/ui/ExceptionView.java
JLLeitschuh: That sends you to the login if you haven't created an account and aren't logged in. I'd rather have this open a browser to the issue list.

src/main/java/edu/wpi/grip/ui/ExceptionView.java
JLLeitschuh: I made up the list. There isn't any downside to getting this information from the user's system.
I'd rather have more info and not need it than not have the information and wish it was there.

src/main/java/edu/wpi/grip/ui/ExceptionView.java
JLLeitschuh: Ease of use. Sure they can click and drag all the way to the bottom of the potentially giant TextArea. This speeds it up. I want to increase the chances that the user will send us an issue by making it as easy as possible.

src/main/java/edu/wpi/grip/ui/ExceptionView.java
JLLeitschuh: Ease of use. 
I know I've been using libraries and had to do this myself and it has been a pain. Trying to reduce the number of steps a user has to take.
We are dealing with high school students, there are laws out there about allowing students to maintain anonymity.

JLLeitschuh:
I think changing it to `Close` would be better

JLLeitschuh:
I changed it to close so that it makes more sense.

ThomasJClark:
Can you fix the tiny dialog issue before this is merged?

src/main/java/edu/wpi/grip/ui/ExceptionView.java
```diff
@@ -0,0 +1,190 @@
...
+import java.io.StringWriter;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Displays an alert with an exception and provides a place for a user to supply information about what caused the error.
+ * Includes a textbox with formatted markdown to allow the issue to be pasted into GitHub easily.
+ * Also, provides links with quick access to the githup issue page.
+ */
+public final class ExceptionView extends Alert {
```
ThomasJClark: `ExceptionDialog` or `ExceptionAlert` would be more appropriate I think, since this isn't a JavaFX control

JLLeitschuh:
I don't know how. If you have any suggestions or code blurbs to toss my way I'd be happy to fix it. As of present that bug is impossible for me to reproduce.

src/main/java/edu/wpi/grip/ui/ExceptionView.java
JLLeitschuh: Yes
- [ ] Fix name

ThomasJClark:
final ExceptionAlert alert = new ExceptionAlert(throwable, getHostServices());
    
    final DialogPane pane = alert.getPane();
    pane.styleProperty().bind(root.styleProperty());
    pane.getStyleSheets().setAll(root.getStyleSheets());

    alert.showAndWait();

ThomasJClark:
Idea: make the "Open GitHub Issue" button go here:

    https://github.com/WPIRoboticsProjects/GRIP/issues/new?title=[EXCEPTION NAME]&body=[ANONYMIZED STACKTRACE AND SYSTEM INFO AND STUFF]

This will open the page for a new issue with everything already filled out.  I think this solves all of the issues each of us had with the other's idea for an exception dialog

* Users are immediately taken to a form to describe the issue (like you want), but it's on GitHub's website (like I want)
* We can include automatic anonymization and markdown-formatted system information (like you want), but the text in our dialog can just have only an unabridged stack trace (like I want)
* We don't require users to copy and paste anything, so there are less steps to do and we have a better chance of getting feedback
* A lot of FRC teams might not already have GitHub accounts, so this URL automatically takes you to a registration page

ThomasJClark:
This is good, just reformat it and I'll hit the merge button

src/main/java/edu/wpi/grip/ui/ExceptionAlert.java
```diff
@@ -0,0 +1,191 @@
...
+        return systemInfo.append("\n").toString();
+    }
+
+    /**
+     * Creates the text that gets pasted into the new github issue.
+     *
+     * @param message The message from the input box
+     * @return The fully constructed issue text.
+     */
+    private String issueText(String message) {
```
ThomasJClark: It doesn't look like the `message` parameter is ever used.  Remove it?

src/main/java/edu/wpi/grip/ui/ExceptionAlert.java
```diff
@@ -0,0 +1,191 @@
...
+        dialogContent.setMaxHeight(Double.MAX_VALUE);
+
+
+        final Label issuePromptText = new Label(ISSUE_PROMPT_TEXT);
+        issuePromptText.setWrapText(true);
+
+        final Label issuePasteLabel = new Label(COPY_PASTE_LABEL_TEXT);
+        issuePasteLabel.setWrapText(true);
+
+        final TextArea issueText = new TextArea(issueText(""));
```
ThomasJClark: The textarea shouldn't include system info and stuff, just the stack trace is fine.

ThomasJClark:
Also there are a couple unused variables in the `ExceptionView` class - they should be removed

src/main/java/edu/wpi/grip/ui/ExceptionAlert.java
JLLeitschuh: That's not what we decided in the meeting earlier this week. This is what we decided upon.
Won't fix.

JLLeitschuh:
I'll take a look.

src/main/java/edu/wpi/grip/ui/ExceptionAlert.java
ThomasJClark: See my comment above.  Having the github link allows us to have the anonymized stack trace in any issues, and still have the dialog show the actual stack trace without extra markdown stuff

JLLeitschuh:
Taken care of.

JLLeitschuh:
> The textarea shouldn't include system info and stuff, just the raw, unabridged stack trace is fine. All of the stuff should still go to the github issue though.
> See my comment above. Having the github link allows us to have the anonymized stack trace in any issues, and still have the dialog show the actual stack trace without extra markdown stuff.

This isn't what we decided in the meeting.

ThomasJClark:
Again, see my previous comment.  We didn't "decide" on the same thing, but adding that GitHub link makes it possible to accomplish what both of us wanted.  I'm trying to get to a compromise here, and I think the end result is must better than either of our original ideas.

Although I don't believe it's absolutely necessary, the user anonymization is a valid thing to do for reports we collect, and the system information is without a doubt useful.  They're inconvenient to include in the dialog, though, since we may want unabridged stack traces for our own purposes, and because the markdown table gets in the way of seeing the exception without scrolling.

We decided this inconvenience was an acceptable compromised for the improvements it would make to issues we'll get, but now that we've found a way to automatically fill in the issue form, we don't have to make that compromise.  We can still have all of the stuff you wanted to do, but we just don't have to show the user that stuff in the alert.

Since you haven't spoken on the actual issue and only on whose solution was "decided", I want to make sure we're clear on 

- We don't need to anonymize the user's home directory *in a dialog that we show the user*
- We don't need to show users their own operating system version as raw markdown 
- We *do* want to include these changes in the abridged error report that we receive in the GitHub issue

If you've moved on from this PR and want to work on one of the many others you have open, that's fine, just let me know so I can take over.

JLLeitschuh:
What if the way that we open the browser doesn't work on some users computer?

ThomasJClark:
I doubt that would ever happen, but in that case the user would have to find the GitHub and report the issue.  They would have to put in the system info manually, but I don't see that being a problem

PR #64, opened 2015-11-02
Reorder steps

ThomasJClark:


src/main/java/edu/wpi/grip/core/Pipeline.java
```diff
@@ -104,6 +104,19 @@ public void onStepRemoved(StepRemovedEvent event) {
     }
 
     @Subscribe
+    public void onStepMoved(StepMovedEvent event) {
+        final Step step = event.getStep();
+
+        synchronized (this) {
+            final int oldIndex = this.steps.indexOf(step);
+            this.steps.remove(oldIndex);
```
JLLeitschuh: Doesn't a list have a method `.remove(Object)`

src/main/resources/edu/wpi/grip/ui/pipeline/Step.fxml
```diff
@@ -1,53 +1,44 @@
...
+        <HBox>
+            <children>
+                <Button fx:id="deleteButton" onMouseClicked="#deleteStep" styleClass="delete" text="✕"
+                        HBox.hgrow="NEVER">
+                    <tooltip>
+                        <Tooltip text="Delete Step"/>
+                    </tooltip>
+                </Button>
+                <Pane HBox.hgrow="ALWAYS"/>
+                <Button fx:id="moveLeftButton" onMouseClicked="#moveStepLeft" styleClass="move-left"
```
JLLeitschuh: Is there a disabled property that this should have on it if it is at the left or right edge of the pipeline?

src/main/java/edu/wpi/grip/core/events/StepMovedEvent.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+public class StepMovedEvent {
+    private final Step step;
+    private final int distance;
+
+    /**
+     * @param step     The step being moved
+     * @param distance The number of indices (positive or negative) to move the step by
+     */
+    public StepMovedEvent(Step step, int distance) {
```
JLLeitschuh: Why did you decide to use distance instead of the index to move to?

JLLeitschuh:
Where does the index checking occur to make sure that the specified index is within the number of steps in the pipeline?

src/main/java/edu/wpi/grip/core/events/StepMovedEvent.java
ThomasJClark: Because steps don't know what index they're at, and a distance better maps to what the buttons actually do

ThomasJClark:
The index checking occurs in subscribers.  Anything that manages a list of steps (ie: [`Pipeline`](https://github.com/WPIRoboticsProjects/GRIP/pull/64/files#diff-09c5c581ac8c2d03086fd30d1bedd817R114)) is responsible for keeping track of where steps are.  Things that send `StepMovedEvent`s do not have the infromation about the rest of the steps, so that's not where the validation goes.

src/main/java/edu/wpi/grip/core/Pipeline.java
```diff
@@ -104,6 +104,19 @@ public void onStepRemoved(StepRemovedEvent event) {
...
+
+            final int newIndex = Math.min(Math.max(oldIndex + event.getDistance(), 0), this.steps.size());
+            this.steps.add(newIndex, step);
+        }
```
JLLeitschuh: So the index check here is using the java built in index out of bounds?

ThomasJClark: it's using `Math.min`/`Math.max` so if a step is moved past the end of the list it appears at the end.

JLLeitschuh: Ah. Okay. That makes sense. Can you put a comment explaining that above the line.
It's not really all that clear from the code.

src/main/java/edu/wpi/grip/core/events/StepMovedEvent.java
JLLeitschuh: Okay.

PR #65, opened 2015-11-02
Update image previews in the UI thread

ThomasJClark:
This is probably the cause of the mysterious scenegraph
NullPointerException that came up before.  The scenegraph should only be
manipulated from the UI thread, and the preview code is actually running
in the thread that eventBus.post was called in, which is the webcam
executor thread.  When multiple threads are touchin the scenegraph
probably hundreds of times per second, a race condition occasionally
happens.

We'll probably need some stress testing to make sure it's actually
fixed.

### DIFF
--- a/src/main/java/edu/wpi/grip/ui/preview/ImageSocketPreviewView.java
+++ b/src/main/java/edu/wpi/grip/ui/preview/ImageSocketPreviewView.java
@@ -5,6 +5,7 @@ import com.google.common.eventbus.Subscribe;
 import com.google.common.primitives.UnsignedBytes;
 import edu.wpi.grip.core.OutputSocket;
 import edu.wpi.grip.core.events.SocketChangedEvent;
+import javafx.application.Platform;
 import javafx.scene.image.Image;
 import javafx.scene.image.ImageView;
 import javafx.scene.image.PixelFormat;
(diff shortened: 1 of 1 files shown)


src/main/java/edu/wpi/grip/ui/preview/ImageSocketPreviewView.java
```diff
@@ -56,62 +57,69 @@ public void onSocketChanged(SocketChangedEvent event) {
...
-            return;
-        }
-
-        // If the size of the Mat changed for whatever reason, allocate a new image with the proper dimensions and a buffer
-        // big enough to hold all of the pixels in the image.
-        if (this.image == null || this.image.getWidth() != width || this.image.getHeight() != height) {
-            this.image = new WritableImage(width, height);
-            this.pixels = IntBuffer.allocate(width * height);
-        }
+        synchronized (this) {
```
JLLeitschuh: Why not make the method synchronized?

src/main/java/edu/wpi/grip/ui/preview/ImageSocketPreviewView.java
```diff
@@ -56,62 +57,69 @@ public void onSocketChanged(SocketChangedEvent event) {
...
-                }
+            final PixelFormat<IntBuffer> argb = PixelFormat.getIntArgbInstance();
 
-                break;
+            Platform.runLater(() -> {
+                synchronized (this) {
+                    this.image.getPixelWriter().setPixels(0, 0, width, height, argb, this.pixels, width);
+                    this.imageView.setImage(this.image);
+                }
+            });
```
JLLeitschuh: Not totally clear, is this the only thing that actually changed?

src/main/java/edu/wpi/grip/ui/preview/ImageSocketPreviewView.java
ThomasJClark: Synchronized just means the the method won't be called reentrantly. This isn't really useful here because it's only ever called from one thread. The synchronized blocks are used to make sure this method and the lambda passed to `runLater` don't run at the same time, since they both access `this.pixels`.

src/main/java/edu/wpi/grip/ui/preview/ImageSocketPreviewView.java
ThomasJClark: Yes, the rest of the diff is  just the  indentation changing

src/main/java/edu/wpi/grip/ui/preview/ImageSocketPreviewView.java
JLLeitschuh: Why is it not synchronized on `this.pixels` then? (This is more me trying to understand).

ThomasJClark: That was my first thought too, but [`this.pixels` might be reallocated](https://github.com/WPIRoboticsProjects/GRIP/pull/65/files#diff-94edecec25826645dd26d6b5355ed371R82), and we wouldn't want one thread accessing an old instance of `this.pixels` while a new thread is concurrently using the newer instance.

JLLeitschuh: Makes sense.

PR #70, opened 2015-11-03
Add Type to InputSocketView and OutputSocketView

JLLeitschuh:


src/main/java/edu/wpi/grip/ui/pipeline/OutputSocketView.java
```diff
@@ -102,6 +105,7 @@ public void initialize(URL url, ResourceBundle resourceBundle) {
 
         // Set the label on the control based on the identifier from the socket hint
         this.identifier.setText(socketHint.getIdentifier());
+        this.type.setText(this.socket.getSocketHint().getType().getSimpleName());
```
ThomasJClark: `this.type.setText(socketHint.getType`... would work and be more concise and consistent with the previous line

JLLeitschuh:
Resolved

PR #71, opened 2015-11-03
Fixes generated peform InputOutput param sockets

JLLeitschuh:
Mats set as InputOutput param types now use `assignTo` instead of `setTo`

### DIFF
--- a/buildSrc/src/main/java/edu/wpi/gripgenerator/templates/SocketHintDeclarationCollection.java
+++ b/buildSrc/src/main/java/edu/wpi/gripgenerator/templates/SocketHintDeclarationCollection.java
@@ -79,7 +79,7 @@ public class SocketHintDeclarationCollection {
     }
 
     private Expression generateCopyExpression(DefinedParamType type, String inputParmId, int inputIndex, String outputParamId, int outputIndex) {
-        // GOAL: ((OutputSocket<Mat>) outputs[0]).getValue().setTo(((InputSocket<Mat>) inputs[0]).getValue());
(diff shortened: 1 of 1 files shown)


ThomasJClark:
What does this fix?

JLLeitschuh:
Set to doesn't work. Set to is designed to be used with Scalars. As an example: `mat.setTo(new Mat(new Scalar(3)));`

ThomasJClark:
Ok, but what's actually fixed now?  I don't think I understand what was wrong

JLLeitschuh:
This is why rectangle wasn't working.

JLLeitschuh:
I just had to fix a comment that was wrong. This is read to merge.

buildSrc/src/main/java/edu/wpi/gripgenerator/settings/DefinedParamType.java
```diff
@@ -66,8 +67,8 @@ public DefinedParamType(String type, Parameter param) {
     }
 
     public DefinedParamType(String type, DefinedParamState state) {
-        this.type = type;
-        this.state = state;
+        this(type);
+        this.state = Optional.ofNullable(state);
```
ThomasJClark: Why is the state optional?  It seems like something like that would be required.  What happens to a socket if it doesn't have a state?  (Also, why is it called "state" and not "direction"?)

JLLeitschuh: I goofed up when I amended my commit. That's not part of the PR now.

JLLeitschuh:
Okay, now it's ready. :p

PR #72, opened 2015-11-04
Fixes the DefinedParamState passed to constructor not taking priority

JLLeitschuh:
Before the `setOutput` would force the state of the param. Now it only
sets it if it hasn't been set by the constructor.

Related #71

### DIFF
--- a/buildSrc/src/main/java/edu/wpi/gripgenerator/settings/DefinedParamType.java
+++ b/buildSrc/src/main/java/edu/wpi/gripgenerator/settings/DefinedParamType.java
@@ -25,7 +25,7 @@ import static com.github.javaparser.ASTHelper.createReferenceType;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 public class DefinedParamType {
-    public enum DefinedParamState {
+    public enum DefinedParamDirection {
         INPUT,
         OUTPUT,
         INPUT_AND_OUTPUT;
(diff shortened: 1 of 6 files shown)


buildSrc/src/main/java/edu/wpi/gripgenerator/settings/DefinedParamType.java
```diff
@@ -41,7 +41,7 @@ public boolean isOutput() {
     }
 
     private final String type;
-    private DefinedParamState state;
+    private Optional<DefinedParamState> state;
```
ThomasJClark: Why is this an Optional?  Shouldn't the state be required?  And why is it called "state" and not "direction"?

JLLeitschuh: I could change it to direction. It defaults to Input when requested.

ThomasJClark: Who though?  I would expect it to not be optional at all.

JLLeitschuh: It's less `Optional` and more undefined until requested. This is better than using null.

ThomasJClark: Why not just initialize it to `DefinedParamState.INPUT`, since that's what happens by default anyways?

JLLeitschuh:
This is ready to merge

PR #73, opened 2015-11-04
Makes Feedback Loops Impossible

JLLeitschuh:
Adds functionality to prevent connections from being made if they will create feedback loops in the pipeline.

### DIFF
--- a/src/main/java/edu/wpi/grip/core/Connection.java
+++ b/src/main/java/edu/wpi/grip/core/Connection.java
@@ -7,6 +7,8 @@ import edu.wpi.grip.core.events.SocketChangedEvent;
 import edu.wpi.grip.core.events.SourceRemovedEvent;
 import edu.wpi.grip.core.events.StepRemovedEvent;
 
+import java.util.Set;
+
 import static com.google.common.base.Preconditions.checkNotNull;
 
 /**
@@ -22,7 +24,7 @@ public class Connection<T> {
      * @param outputSocket The socket to listen for changes in.
      * @param inputSocket  A different socket to update when a change occurs in the first.
      */
-    public Connection(EventBus eventBus, OutputSocket<? extends T> outputSocket, InputSocket<T> inputSocket) {
+    public Connection(EventBus eventBus, OutputSocket<? extends T> outputSocket, InputSocket<T> inputSocket) throws InfiniteLoopException {
         this.eventBus = eventBus;
         this.outputSocket = outputSocket;
         this.inputSocket = inputSocket;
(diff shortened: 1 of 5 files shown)


src/main/java/edu/wpi/grip/core/Connection.java
```diff
@@ -31,11 +34,33 @@ public Connection(EventBus eventBus, OutputSocket<? extends T> outputSocket, Inp
...
+     * Determines if this connection will create an infinite loop
+     *
+     * @param outputSocket The input socket to search for sockets
+     * @param inputSocket  The output socket to search for sockets
+     * @return False if this will create an infinite loop
+     */
+    private boolean willCreateInfiniteLoop(OutputSocket<?> outputSocket, InputSocket<?> inputSocket) {
+        // There can't be any loops between two sockets that aren't owned by steps
+        if (!outputSocket.getStep().isPresent()) return false;
+        checkArgument(inputSocket.getStep().isPresent(), "An input socket must be owned by a step. Only output sockets can be owned by a Source");
```
ThomasJClark: Wouldn't you just return false in this case?  If we ever have a situation where an inputsocket isn't a step (which we don't currently), there's no way it could be in a loop.  I'm not sure if this bit of extra validation should be in this method.

src/main/java/edu/wpi/grip/core/Connection.java
```diff
@@ -79,4 +104,13 @@ public void removeConnection(SourceRemovedEvent e) {
             }
         }
     }
+
+    /**
+     * Thrown when the user tries to create a connection that will result in an infinite loop.
+     */
+    public static class InfiniteLoopException extends Error {
```
ThomasJClark: I'm not sure this should be an exception.  Exceptions are for signaling error conditions, not input validation.

A better solution might be to use the `willCreateInfiniteLoop` method to prevent users from creating the loops in the first place instead of printing an exception if they do.

src/main/java/edu/wpi/grip/core/Connection.java
JLLeitschuh: I'd rather use an assertion statement

src/main/java/edu/wpi/grip/core/Connection.java
JLLeitschuh: I beg to differ. Look at the java `URL` object. It throws an exception if the URL is malformed.

ThomasJClark: I don't see how that's relevant.

I'd much rather users couldn't use the UI to do an invalid action and cause an exception whenever possible.  I think it's confusing if you *can* connect steps in a loop, but nothing happens and an error is printed to the console or something.

We could also only allow connections form left to right in the pipeline, so the step with the output socket must be before the step with the input socket.  This would also prevent loops, and we wouldn't really lose any functionality by adding that rule (we should add that rule eventually anyways)

src/main/java/edu/wpi/grip/core/Connection.java
ThomasJClark: why?

JLLeitschuh: It seems like it is more of an assertion sort of thing. Its an invalid state for the program to be in.

src/main/java/edu/wpi/grip/core/Connection.java
JLLeitschuh: I'd rather use exceptions over expecting future developers to call another validation method prior to using the connection class. This really is a unique error that would cause a legitimate concern.

JLLeitschuh:
It now uses an exception instead of an error.

PR #74, opened 2015-11-04
Makes the Step in Socket an Optional value

JLLeitschuh:
Since Sockets wont have steps for Sources this makes sense to prevent it from
being null.

### DIFF
--- a/src/main/java/edu/wpi/grip/core/Socket.java
+++ b/src/main/java/edu/wpi/grip/core/Socket.java
@@ -4,9 +4,13 @@ import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.eventbus.EventBus;
 import com.google.common.eventbus.Subscribe;
-import edu.wpi.grip.core.events.*;
+import edu.wpi.grip.core.events.ConnectionAddedEvent;
+import edu.wpi.grip.core.events.ConnectionRemovedEvent;
+import edu.wpi.grip.core.events.SocketChangedEvent;
+import edu.wpi.grip.core.events.SocketConnectedChangedEvent;
 
 import java.util.HashSet;
+import java.util.Optional;
 import java.util.Set;
 
 import static com.google.common.base.Preconditions.checkNotNull;
@@ -24,7 +28,7 @@ public abstract class Socket<T> {
     public enum Direction {INPUT, OUTPUT}
 
     protected final EventBus eventBus;
-    private Step step;
+    private Optional<Step> step = Optional.empty();
     private final Direction direction;
     private final Set<Connection> connections = new HashSet<>();
     private final SocketHint<T> socketHint;
@@ -94,14 +98,14 @@ public abstract class Socket<T> {
     /**
      * @param step The step that this socket is part of.
      */
-    public void setStep(Step step) {
+    public void setStep(Optional<Step> step) {
         this.step = step;
     }
 
     /**
      * @return The step that this socket is part of, or <code>null</code> if it has not been set.
      */
-    public Step getStep() {
+    public Optional<Step> getStep() {
         return step;
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/edu/wpi/grip/core/Step.java
```diff
@@ -67,7 +67,7 @@ public void onInputSocketChanged(SocketChangedEvent e) {
         final Socket socket = e.getSocket();
 
         // If this socket that changed is one of the inputs to this step, run the operation with the new value.
-        if (socket.getStep() == this && socket.getDirection().equals(Socket.Direction.INPUT)) {
+        if (socket.getStep().isPresent() && socket.getStep().get() == this && socket.getDirection().equals(Socket.Direction.INPUT)) {
```
ThomasJClark: You can do `socket.getStep().ifPresent(step ->`...  to avoid the manual check.

Another way of doing it is `if (socket.getStep().equals(Optional.of(this))`

src/main/java/edu/wpi/grip/core/Socket.java
```diff
@@ -95,13 +99,13 @@ public T getValue() {
      * @param step The step that this socket is part of.
      */
     public void setStep(Step step) {
```
ThomasJClark: Should be `setStep(Optional<Step> step) {`

src/main/java/edu/wpi/grip/core/Step.java
JLLeitschuh: `ifPresent` returns a void function

src/main/java/edu/wpi/grip/core/Socket.java
JLLeitschuh: Why?

src/main/java/edu/wpi/grip/core/Step.java
JLLeitschuh: I thought you were adamantly against using the `.equals` method in case it was overloaded in our code eventually.

ThomasJClark: Could you clarify these comments?  In particular,
* `ifPresent` doesn't return anything, and I'm not sure why it would matter if it did.
* I don't think `Optional::equals` is ever going to change on us, and I don't think I've ever said I did.  We couldn't change it in our code if we wanted to, since `Optional` is `final`, but I'm not sure why that matters since we would never do that.

All I'm suggestion is that we make idiomatic use of the `java.util.Optional` class.

src/main/java/edu/wpi/grip/core/Socket.java
ThomasJClark: Because step is an `Optional`.

JLLeitschuh: Whatever, this doesn't make any sense but I'll do it if it makes you happy. I don't think its a good design decision. Socket should be empty if unset and if you set it by calling `setStep(Step step)` because that will assign the step into the optional value. I don't think it should be possible to redefine the step as empty after you have assigned it.

src/main/java/edu/wpi/grip/core/Step.java
```diff
@@ -64,10 +66,10 @@ public Operation getOperation() {
 
     @Subscribe
     public void onInputSocketChanged(SocketChangedEvent e) {
-        final Socket socket = e.getSocket();
+        final Socket<?> socket = e.getSocket();
 
         // If this socket that changed is one of the inputs to this step, run the operation with the new value.
-        if (socket.getStep() == this && socket.getDirection().equals(Socket.Direction.INPUT)) {
+        if( socket.getStep().equals(Optional.of(this)) && socket.getDirection().equals(Socket.Direction.INPUT)) {
```
ThomasJClark: Ctrl+Shift+L

JLLeitschuh:
Perhaps it would be better to have a function 
````java
void assignStep(Step step){
    this.step.ifPresent(()-> throw new CallOutOfOrder("Can not assign step after it has already been assigned"));
    this.socket = Optional.of(step)
}
```
If you disagree just go ahead and merge it.

PR #76, opened 2015-11-04
Adds EventHistory class to observe history for easy undo

JLLeitschuh:
Currently not working. Allows you to click undo and have the
last event that fired be rolled back.

### DIFF
--- a/src/main/java/edu/wpi/grip/core/EventHistory.java
+++ b/src/main/java/edu/wpi/grip/core/EventHistory.java
@@ -1,0 +1,50 @@
+package edu.wpi.grip.core;
+
+
+import com.google.common.eventbus.EventBus;
+import com.google.common.eventbus.Subscribe;
+import edu.wpi.grip.core.events.RedoPublishedEvent;
+import edu.wpi.grip.core.events.UndoPublishedEvent;
+import edu.wpi.grip.core.events.UndoableEvent;
+
+import java.util.Optional;
+import java.util.Stack;
+
+public class EventHistory {
+    private final Stack<UndoableEvent> undoableEvents = new Stack<>();
+    private final Stack<UndoableEvent> redoableEvents = new Stack();
+    private final EventBus eventBus;
+    private Optional<UndoableEvent> lastEventPosted = Optional.empty();
+
+    public EventHistory(EventBus eventBus){
+        this.eventBus = eventBus;
+        eventBus.register(this);
+    }
+
+    @Subscribe
+    public void onUndoPublishedEvent(UndoPublishedEvent event){
+        publishedEventHandler(undoableEvents, redoableEvents);
+    }
+
+    @Subscribe
+    public void onRedoPublishedEvent(RedoPublishedEvent event){
+        publishedEventHandler(redoableEvents, undoableEvents);
+    }
+
+    private void publishedEventHandler(Stack<UndoableEvent> eventStack, Stack<UndoableEvent> otherStack){
+        if (!eventStack.isEmpty()){
+            final UndoableEvent undoableEvent = eventStack.pop().createUndoEvent();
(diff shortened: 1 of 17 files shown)


src/main/java/edu/wpi/grip/core/EventHistory.java
```diff
@@ -0,0 +1,43 @@
...
+import com.google.common.eventbus.Subscribe;
+import edu.wpi.grip.core.events.UndoPublishedEvent;
+import edu.wpi.grip.core.events.UndoableEvent;
+
+import java.util.LinkedList;
+import java.util.Optional;
+import java.util.Queue;
+
+public class EventHistory {
+    private final Queue<UndoableEvent> undoableEvents = new LinkedList<>();
```
ThomasJClark: Why a queue? Wouldn't a stack be better, so we can pop the most recently done action?

JLLeitschuh: Thats what my problem was! Thank you!

JLLeitschuh: What do you know... Hahah. Using the correct data type actually solved my problem :laughing:

JLLeitschuh:
So this works in all cases now except for when doing a StepRemoved where there are connections on that step. What happens currently is that the last connection removed tries to be re-added but it cant because the socket was removed.

ThomasJClark:
That's because the `ConnectionRemovedEvent`s are sent by a subscriber after the relevant `StepRemovedEvent` is sent.  So, according to the event history, the connections were technically removed after the step.  I'm not sure how to solve that.

JLLeitschuh:
Could we change the order that the events fire?

ThomasJClark:
I don't think so.  There isn't much way around the order we send the events that I can think of without really hacking things up, and the order that `@Subscribe` methods are called is undefined by Guava, so we can't just use a hacky solution that relies on that.

One solution might be to have `EventHistory` insert things in the middle of the stack where appropriate.  So, it would receive the same events in the same order, but it would know that `ConnectionRemovedEvent`s for steps that don't exist anymore should be pushed to before the corresponding `StepRemovedEvent`.  This saves us from putting order-dependent hacks into the codebase, and the only consequence I can think of is that we would probably want to use a linked list instead of a stack

ThomasJClark:
Some other thoughts:

* If an event is undone, we could add the original event to a separate "redo" stack so you can also redo undone steps.
* Is there really anything gained by having separate `AddEvent` and `RemoveEvent` interfaces?  I see what it does, but I'm not sure it's necessary. They could just all implement `UndoableEvent`.
* `UndoableEvent` could have a `getName()` method so we could have a menu option that says something like "Undo Add Blur" or "Redo Remove Connection".

src/main/java/edu/wpi/grip/ui/controllers/MainWindowController.java
```diff
@@ -38,6 +42,14 @@
                 + context.getSubscriber() + " to " + context.getSubscriberMethod(), exception).printStackTrace(System.err);
     });
```
JLLeitschuh: Codacy Issue found: [Perhaps 'eventHistory' could be replaced by a local variable.](https://www.codacy.com/app/WPIRoboticsProjects/GRIP/file/1275438274/issues/source?bid=2943107&fileBranchId=2943115#l45)

JLLeitschuh: Codacy Issue found: [Avoid unused private fields such as 'eventHistory'.](https://www.codacy.com/app/WPIRoboticsProjects/GRIP/file/1275438274/issues/source?bid=2943107&fileBranchId=2943115#l45)

PR #78, opened 2015-11-05
Adds an Object to publish all openCV operations to the EventBus

JLLeitschuh:


src/main/java/edu/wpi/grip/core/operations/OpenCVOperations.java
```diff
@@ -0,0 +1,25 @@
...
+import edu.wpi.grip.generated.CVOperations;
+
+public class OpenCVOperations extends CVOperations {
+
+    /**
+     * Adds all of the generated OpenCV operations to the pipeline
+     * @param eventBus The Guava Event Bus
+     */
+    public static void publishAllOperations(final EventBus eventBus){
+        eventBus.post(new OperationAddedEvent(new NewPointOperation()));
```
ThomasJClark: IMO these operations should be added in the same place as the rest of the manually written ones.  This class only needs to add the generated operations

src/main/java/edu/wpi/grip/core/operations/OpenCVOperations.java
```diff
@@ -0,0 +1,25 @@
...
+
+
+import com.google.common.eventbus.EventBus;
+import edu.wpi.grip.core.events.OperationAddedEvent;
+import edu.wpi.grip.core.operations.opencv.MatFieldAccessor;
+import edu.wpi.grip.core.operations.opencv.NewPointOperation;
+import edu.wpi.grip.core.operations.opencv.NewSizeOperation;
+import edu.wpi.grip.generated.CVOperations;
+
+public class OpenCVOperations extends CVOperations {
```
ThomasJClark: This class hierarchy is a little confusing.  OpenCVOperations extends CVOperations?  It sounds like these should really be the same class, and you can just have the generated stick in the method below.

src/main/java/edu/wpi/grip/core/operations/OpenCVOperations.java
```diff
@@ -0,0 +1,25 @@
...
+
+    /**
+     * Adds all of the generated OpenCV operations to the pipeline
+     * @param eventBus The Guava Event Bus
+     */
+    public static void publishAllOperations(final EventBus eventBus){
```
ThomasJClark: `addOperations` would be a more accurate name.  Nothing is published here.  Also ctrl+shift+L

src/main/java/edu/wpi/grip/core/operations/OpenCVOperations.java
JLLeitschuh: No

src/main/java/edu/wpi/grip/core/operations/OpenCVOperations.java
JLLeitschuh: I disagree. They are all opencv methods just generated vs handwritten. If they aren't a composite operation then they should all be created in the same place.

src/main/java/edu/wpi/grip/core/operations/OpenCVOperations.java
JLLeitschuh: You are more than welcome to take a crack at doing this yourself if you'd like though. It would give you the opportunity to become familiar with the javaparser library.

src/main/java/edu/wpi/grip/core/operations/OpenCVOperations.java
JLLeitschuh: `postAllOperations` might be better.

src/main/java/edu/wpi/grip/core/operations/OpenCVOperations.java
ThomasJClark: ?

ThomasJClark: > You are more than welcome to take a crack at doing this yourself if you'd like though. It would give you the opportunity to become familiar with the javaparser library.

Ok, I'll do that tonight

Issue #61, opened 2015-10-30
Operations To Remove


JLLeitschuh:
Please create a list of operations and inputs that should be removed and leave them here.

- [x] batch distance
- [x] ddepth
- [x] houghlines
- [x] mask <== Better to keep and extend the functionality.
- [ ] cvtColor -> dstCn

PR #81, opened 2015-11-05
Increases functionality of opencv core generated code

JLLeitschuh:
Adds various functions and descriptions to the core set of
opencv functions. Now fully supports the use of the Mask param.

Related: #61

### DIFF
--- a/buildSrc/src/main/java/edu/wpi/gripgenerator/FileParser.java
+++ b/buildSrc/src/main/java/edu/wpi/gripgenerator/FileParser.java
@@ -201,12 +201,20 @@ public class FileParser {
                 new DefinedMethod("rectangle", false,
                         new DefinedParamType("Mat", DefinedParamType.DefinedParamDirection.INPUT_AND_OUTPUT),
                         new DefinedParamType("Point"))
-        ).setDirectionDefaults(DefinedParamType.DefinedParamDirection.OUTPUT, "dst").setIgnoreDefaults("dtype");
+        ).setDirectionDefaults(DefinedParamType.DefinedParamDirection.OUTPUT, "dst")
+                .setIgnoreDefaults("dtype", "ddepth");
         new OpenCVMethodVisitor(collection).visit(imgprocDeclaration, compilationUnits);
         collection.generateCompilationUnits(collector, compilationUnits, operations);
         return compilationUnits;
     }
 
+    /**
+     * Defines all of the methods that get parsed out of OpenCV Core.
+     * @param coreDeclaration The source from the core file
+     * @param collector The collection of predefined fields
+     * @param operations The list of operations that will be generated at the end.
+     * @return The map of the name of the file to the contents of the new source file
+     */
     public static Map<String, CompilationUnit> parseOpenCVCore(CompilationUnit coreDeclaration, DefaultValueCollector collector, OperationList operations) {
         Map<String, CompilationUnit> compilationUnits = new HashMap<>();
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/java/edu/wpi/gripgenerator/FileParser.java
```diff
@@ -216,43 +224,79 @@ public Expression getDefaultValue(String defaultValue) {
...
+                        .addDescription("Calculates the per-element bit-wise disjunction of two Mats"),
+                new DefinedMethod("bitwise_xor", false, "Mat", "Mat")
+                        .addDescription("Calculates the per-element bit-wise \\\"exclusive or\\\" operation on two Mats"),
+                new DefinedMethod("bitwise_not", false, "Mat", "Mat")
+                        .addDescription("Calculates per-element bit-wise inversion of the input Mat")
+        )
+                .setDirectionDefaults(DefinedParamType.DefinedParamDirection.INPUT_AND_OUTPUT, "dst")
+                .setIgnoreDefaults("dtype");
+
+        DefinedMethodCollection collection2 = new DefinedMethodCollection(baseClassName,
```
ThomasJClark: Why is there a `collection2`?  Why can't they all be in one collection?

buildSrc/src/main/java/edu/wpi/gripgenerator/FileParser.java
```diff
@@ -216,43 +224,79 @@ public Expression getDefaultValue(String defaultValue) {
...
+                ).addDescription("Performs the per-element comparison of two Mats"),
+                new DefinedMethod("max", false, "Mat", "Mat")
+                        .addDescription("Calculates per-element maximum of two Mats"),
                 new DefinedMethod("min", false, "Mat", "Mat")
+                        .addDescription("Calculates per-element minimum of two Mats"),
+                new DefinedMethod("mulTransposed", false, "Mat", "Mat")
+                        .addDescription("Calculates the product of a matrix and its transposition. " +
+                                "The optional delta matrix subtracted from src before the " +
+                                "multiplication."),
+                new DefinedMethod("transpose", false, "Mat", "Mat")
```
ThomasJClark: These non-image mat operations are not useful right now because we have nothing that deals with non-image mats

buildSrc/src/main/java/edu/wpi/gripgenerator/FileParser.java
JLLeitschuh: Because the first one handles `dst` as an input and an output. The second group does not.

buildSrc/src/main/java/edu/wpi/gripgenerator/FileParser.java
JLLeitschuh: Transpose works though. It flips the image over its diagonal axis.

buildSrc/src/main/java/edu/wpi/gripgenerator/FileParser.java
JLLeitschuh: This allows you to take full advantage of the mask parameter as it was intended to be used.

buildSrc/src/main/java/edu/wpi/gripgenerator/FileParser.java
ThomasJClark: We can do that with flip, though

JLLeitschuh:
Closes #106

PR #82, opened 2015-11-05
Allows Webcam Sources to be removed

### LINKED ISSUE #80: Can't Remove WebcamSource
It is impossible to remove a webcam source.

JLLeitschuh:
Webcameras can now be removed when the delete button is pressed.

Closes #80

### DIFF
--- a/src/main/java/edu/wpi/grip/core/sources/WebcamSource.java
+++ b/src/main/java/edu/wpi/grip/core/sources/WebcamSource.java
@@ -2,9 +2,11 @@ package edu.wpi.grip.core.sources;
 
 
 import com.google.common.eventbus.EventBus;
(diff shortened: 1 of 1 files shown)


ThomasJClark:
`Pipeline::onSourceRemoved` should unregister the source from the event bus.  Besides that, this is good

JLLeitschuh:
That doesn't work. If the source is unregistered before the event is passed off to it the it will never be removed.
I'll have to do it inside of the source.

JLLeitschuh:
Resolved

ThomasJClark:
Ok, makes sense.  Now that I think about it, it would probably make sense for `Step`, `Connection`, etc, to also similarly unregister themselves instead of having `Pipeline` do it.  Single responsibility and encapsulation and such.

PR #83, opened 2015-11-05
Adds MinMaxLoc OpenCV Operation

JLLeitschuh:


src/main/java/edu/wpi/grip/core/operations/opencv/MinMaxLoc.java
```diff
@@ -0,0 +1,75 @@
...
+            minValOutputHint = new SocketHint<Number>("minVal", Number.class, 0),
+            maxValOutputHint = new SocketHint<Number>("maxVal", Number.class, 0);
+
+    private final SocketHint<Point>
+            minLocOutputHint = new SocketHint<Point>("minLoc", Point.class, Point::new, SocketHint.View.NONE, null),
+            maxLocOutputHint = new SocketHint<Point>("maxLoc", Point.class, Point::new, SocketHint.View.NONE, null);
+
+    @Override
+    public String getName() {
+        return "Min Max Location";
```
ThomasJClark: "Find Min and Max" would be a good name for this operation

src/main/java/edu/wpi/grip/core/operations/opencv/MinMaxLoc.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @Override
+    public String getDescription() {
+        return "Finds the global minimum and maximum in an a single chanel Matrix.";
```
ThomasJClark: *Find

*single channel matrix (or just say "grayscale image")

src/main/java/edu/wpi/grip/core/operations/opencv/MinMaxLoc.java
```diff
@@ -0,0 +1,75 @@
...
+
+/** Operation to call {@link opencv_core#minMaxLoc} */
+public class MinMaxLoc implements CVOperation {
+
+    private final SocketHint<Mat>
+            srcInputHint = new SocketHint<Mat>("src", Mat.class, Mat::new, SocketHint.View.NONE, null),
+            maskInputHint = new SocketHint<Mat>("mask", Mat.class, Mat::new, SocketHint.View.NONE, null);
+
+    private final SocketHint<Number>
+            minValOutputHint = new SocketHint<Number>("minVal", Number.class, 0),
```
ThomasJClark: You can give these actual names like "Minimum Value", "Maximum Value", etc...

src/main/java/edu/wpi/grip/core/operations/opencv/MinMaxLoc.java
```diff
@@ -0,0 +1,75 @@
...
+                new OutputSocket(eventBus, maxLocOutputHint),
+        };
+    }
+
+    @Override
+    public void perform(InputSocket<?>[] inputs, OutputSocket<?>[] outputs) {
+        final Mat src = (Mat) inputs[0].getValue();
+        Mat mask = (Mat) inputs[1].getValue();
+        if (mask.empty()) mask = null;
+        final DoublePointer minVal = new DoublePointer(1);
```
ThomasJClark: Is there a reason why you allocate new doubles here?  Why not just use the ones in the output sockets?

JLLeitschuh: Because they aren't Double Pointers. They aren't Numbers. We need the raw numerical values. We have no way to handle Double Pointers.

PR #89, opened 2015-11-06
Sorts Operations by name and allows them to be filtered

JLLeitschuh:
You can now use the search filter to find operations in the pallet.
Additionally when the are added to the application they are
automaitically sorted by name


src/main/java/edu/wpi/grip/ui/OperationView.java
```diff
@@ -88,5 +88,31 @@ public void initialize(URL url, ResourceBundle resourceBundle) {
             Step step = new Step(this.eventBus, this.operation);
             this.eventBus.post(new StepAddedEvent(step));
         });
+
+        // Ensures that when this element is hidden that it also removes its size calculations
+        this.managedProperty().bind(this.visibleProperty());
+    }
+
+    // Cache the lower case name so that it isn't calculated each time the search bar is used
+    private String nameLoweCase = null;
```
ThomasJClark: I doubt this is actually a bottleneck.  Have you profiled it?

src/main/java/edu/wpi/grip/ui/OperationView.java
```diff
@@ -88,5 +88,31 @@ public void initialize(URL url, ResourceBundle resourceBundle) {
...
+    /**
+     * Uses the search string from the palette and allows this element to determine if it should be visible
+     * @param lowerCaseSearchString The search string that should be used to determine visibility
+     */
+    void determineVisibility(String lowerCaseSearchString){
```
ThomasJClark: > `){`

ಠ_ಠ

JLLeitschuh: Oops.

src/main/java/edu/wpi/grip/ui/OperationView.java
```diff
@@ -88,5 +88,31 @@ public void initialize(URL url, ResourceBundle resourceBundle) {
...
+        assert lowerCaseSearchString.equals(lowerCaseSearchString.toLowerCase()) : "The string passed to determine visibility was not lower case";
```
ThomasJClark: A more robust solution might be to accept any string and do a `toLowerCase()` inside this method.  I very much doubt that would have the slightest performance impact.

src/main/java/edu/wpi/grip/ui/OperationView.java
```diff
@@ -88,5 +88,31 @@ public void initialize(URL url, ResourceBundle resourceBundle) {
...
+            this.setVisible(true);
+        } else {
+            this.setVisible(false);
+        }
+    }
+
+    /*
+     * Intentionally left package protected
+     */
+    String getOperationName(){
```
ThomasJClark: Better to just add a `public Operation getOperation()`.  We're probably going to do a lot more with sorting and filtering and stuff than just by name, and it would be nice to not have to add a method to this class every time we want to get a different thing from `Operation`.

src/main/java/edu/wpi/grip/ui/OperationView.java
JLLeitschuh: Fine. I suppose you are right.

src/main/java/edu/wpi/grip/ui/OperationView.java
```diff
@@ -88,5 +88,31 @@ public void initialize(URL url, ResourceBundle resourceBundle) {
...
+            this.setVisible(true);
+        } else {
+            this.setVisible(false);
+        }
+    }
+
+    /*
+     * Intentionally left package protected
+     */
+    Operation getOperation() {
```
JLLeitschuh: Codacy Issue found: [Use explicit scoping instead of the default package private level](https://www.codacy.com/app/WPIRoboticsProjects/GRIP/file/1273862759/issues/source?bid=2943107&fileBranchId=2943112#l115)

src/main/java/edu/wpi/grip/ui/OperationView.java
```diff
@@ -88,5 +88,31 @@ public void initialize(URL url, ResourceBundle resourceBundle) {
...
+        this.managedProperty().bind(this.visibleProperty());
+    }
+
+
+    /**
+     * Uses the search string from the palette and allows this element to determine if it should be visible
+     *
+     * @param searchString The search string that should be used to determine visibility
+     */
+    void determineVisibility(String searchString) {
```
JLLeitschuh: Codacy Issue found: [Use explicit scoping instead of the default package private level](https://www.codacy.com/app/WPIRoboticsProjects/GRIP/file/1273862759/issues/source?bid=2943107&fileBranchId=2943112#l103)

JLLeitschuh: Codacy Issue found: [Avoid reassigning parameters such as 'searchString'](https://www.codacy.com/app/WPIRoboticsProjects/GRIP/file/1273862759/issues/source?bid=2943107&fileBranchId=2943112#l103)

src/main/java/edu/wpi/grip/ui/OperationView.java
```diff
@@ -88,5 +88,28 @@ public void initialize(URL url, ResourceBundle resourceBundle) {
...
+        this.managedProperty().bind(this.visibleProperty());
+    }
+
+
+    /**
+     * Uses the search string from the palette and allows this element to determine if it should be visible
+     *
+     * @param searchString The search string that should be used to determine visibility
+     */
+    public void determineVisibility(final String searchString) {
```
ThomasJClark: `PaletteView` should set the visibility of operations.  Putting searching stuff inside of the class being hidden instead of the class that handles searching does not seem like good practice.

We also might want to change the algorithm to use an actual fuzzy matching function

src/main/java/edu/wpi/grip/ui/PaletteView.java
```diff
@@ -37,6 +42,14 @@ public PaletteView(EventBus eventBus) {
             throw new RuntimeException(e);
         }
 
+        this.operationSearch.textProperty().addListener(
+                change -> {
+                    // Store the value so it isn't computed multiple times
```
ThomasJClark: What value?

src/main/java/edu/wpi/grip/ui/OperationView.java
JLLeitschuh: Its an, I act on myself, instead of an, I am acted upon design pattern. Better encapsulation.

src/main/java/edu/wpi/grip/ui/PaletteView.java
JLLeitschuh: Good call.

src/main/java/edu/wpi/grip/ui/OperationView.java
ThomasJClark: I know how design patterns work.  I disagree with the way you applied one.

This is bad encapsulation because the search algorithm is leaked into the operation view.  The operation view is a button that you click to add a step.  It shouldn't do any searching.  The palette, on the other hand, actually has a search bar and listeners for searching, so it should call the search algorithm.

src/main/java/edu/wpi/grip/ui/PaletteView.java
```diff
@@ -49,6 +62,12 @@ public void clearOperations() {
 
     @Subscribe
     public void onOperationAdded(OperationAddedEvent event) {
-        this.operations.getChildren().add(new OperationView(this.eventBus, event.getOperation()));
+        OperationView view = new OperationView(this.eventBus, event.getOperation());
+        view.determineVisibility(this.operationSearch.getText());
```
ThomasJClark: This is a little hacky.  Why not just have a `filterOperations` invalidation listener, and add it to both `this.getChildren()` and `this.operationSearch.textProperty()`?

JLLeitschuh: I don't understand

ThomasJClark: Have a listener specifically for filtering operations:

    final InvalidationListener filterOperations = observable ->
        this.operations.getChildren().forEach(node -> {
            final Operation operation = ((OperationView) node).getOperation();
            node.setVisible(myTextSearchAlgorithmMatches(this.textField.getText(), operation.getName()));
        });

Then make sure it's called at the appropriate times using JavaFX listeners

    // Update the visibility of each node when either the list of nodes or the filter text changes
    this.getChildren().addListener(filterOperations);
    this.textField.textProperty().addListener(filterOperations);

src/main/java/edu/wpi/grip/ui/OperationView.java
ThomasJClark: Also, keep in mind that  `String::contains` will probably not be our final search function.  Assuming we eventually have a more advanced fuzzy search:

1. The proposed design doesn't allow us to sort by relevance to a search (so if I type "eige", eigenvalue related operations show up before "addWeighted")
2. The proposed design allow us to search with more advanced options (search by name, by description, by category, using regex) without adding more methods to let `OperationView` access the values in the views inside of the `PaletteView`.

src/main/java/edu/wpi/grip/ui/PaletteView.java
```diff
@@ -49,6 +105,12 @@ public void clearOperations() {
 
     @Subscribe
     public void onOperationAdded(OperationAddedEvent event) {
-        this.operations.getChildren().add(new OperationView(this.eventBus, event.getOperation()));
+        OperationView view = new OperationView(this.eventBus, event.getOperation());
+        this.operations.getChildren().add(view);
+        FXCollections.sort(this.operations.getChildren(),
```
ThomasJClark: If we're going to sort all operations by name, we have to add categories so the groups of related operations don't get scrambled up

src/main/java/edu/wpi/grip/ui/PaletteView.java
```diff
@@ -37,9 +52,50 @@ public PaletteView(EventBus eventBus) {
...
+
+        // Update the visibility of each node when either the list of nodes or the filter text changes
+        this.operations.getChildren().addListener(filterOperations);
+        this.operationSearch.textProperty().addListener(filterOperations);
+
         this.eventBus.register(this);
     }
 
+    @Override
+    public void initialize(URL location, ResourceBundle resources) {
```
ThomasJClark: Why?

src/main/java/edu/wpi/grip/ui/PaletteView.java
JLLeitschuh: Not in this pull request.

JLLeitschuh: If you need a threshold then you can search on threshold.

ThomasJClark: Ok, then don't alphabetize the operations yet

JLLeitschuh: Brad wanted it this way.

src/main/java/edu/wpi/grip/ui/PaletteView.java
JLLeitschuh: Ooops

src/main/java/edu/wpi/grip/ui/PaletteView.java
ThomasJClark: Yeah, instead of adding the search feature.  This change also adds a search feature.

Either way, alphabetizing every operation makes it difficult to browse them because the related ones are all scrambled around.  If you think the generated ones should be alphabetized, maybe we can just make the generator code generate them in alphabetical order.

src/main/resources/edu/wpi/grip/ui/Palette.fxml
```diff
@@ -1,17 +1,16 @@
...
-         fillWidth="true" styleClass="palette">
+<?import javafx.scene.layout.VBox?>
+<?import org.controlsfx.control.textfield.CustomTextField?>
+<fx:root fillWidth="true" styleClass="palette" type="javafx.scene.layout.VBox" xmlns="http://javafx.com/javafx/8.0.40"
+         xmlns:fx="http://javafx.com/fxml/1">
     <children>
         <Label styleClass="pane-title" text="Operation Palette"/>
-        <ScrollPane hbarPolicy="NEVER" vbarPolicy="AS_NEEDED" fitToWidth="true" fitToHeight="true"
-                    fx:id="palettePane">
+        <CustomTextField fx:id="operationSearch"></CustomTextField>
```
ThomasJClark: Recommendation:

    <CustomTextField fx:id="operationSearch" promptText="search"/>

So it's clear what the text field does

src/main/java/edu/wpi/grip/ui/PaletteView.java
JLLeitschuh: It absolutely does. I can't keep track of what operation is where. Alphabetical is a pattern that most human beings can recognize as good way to organize data.

src/main/resources/edu/wpi/grip/ui/Palette.fxml
JLLeitschuh: Good call.

src/main/java/edu/wpi/grip/ui/PaletteView.java
```diff
@@ -37,9 +49,46 @@ public PaletteView(EventBus eventBus) {
...
+        this.operationSearch.textProperty().addListener(filterOperations);
+
         this.eventBus.register(this);
     }
 
+
+    // XXX: This is the only way to do this unfortunately.
+    // This is where this came from: https://bitbucket.org/controlsfx/controlsfx/issues/330/making-textfieldssetupclearbuttonfield
+    private void setupClearButtonField(CustomTextField customTextField) throws Exception {
+        Method m = TextFields.class.getDeclaredMethod("setupClearButtonField", TextField.class, ObjectProperty.class);
```
ThomasJClark: I actually didn't even notice the clear button at first.  Would it be better to do something like this?  It's infinitely more cleaner than using reflection, and I like it better because it shows an actual button that says "clear".

Palette.fxml

        <CustomTextField fx:id="search" promptText="Search">
            <right>
                <Button fx:id="clearSearch" text="clear"/>
            </right>
        </CustomTextField>

PaletteView.java

    @FXML
    private TextInputControl search;

    @FXML
    private Button clearSearch;

...

        this.clearSearch.setOnMouseClicked(event-> this.search.clear());
        this.clearSearch.visibleProperty().bind(this.search.textProperty().isEmpty().not());

JLLeitschuh: No, this is pretty common practice to put a clear `x` button in the right side of an input box.

PR #100, opened 2015-11-08
Adds Socket Label Coloring when input invalid

JLLeitschuh:
Indicates to the user what sockets need to be fixed in order to make the operation run correctly.


JLLeitschuh:
Woot! #100

PR #104, opened 2015-11-09
Fixes allocating large chunks of memory within operations

JLLeitschuh:


src/main/java/edu/wpi/grip/core/operations/composite/ThresholdOperation.java
```diff
@@ -0,0 +1,33 @@
+package edu.wpi.grip.core.operations.composite;
+
+import edu.wpi.grip.core.Operation;
+import org.bytedeco.javacpp.opencv_core.Mat;
+import org.bytedeco.javacpp.opencv_core.Scalar;
+
+import java.util.Optional;
+
+public abstract class ThresholdOperation implements Operation {
```
ThomasJClark: This should probably also implement `getIcon()` so all of the threshold operations have icons and not just RGB

src/main/java/edu/wpi/grip/core/Operation.java
```diff
@@ -42,12 +42,27 @@
     OutputSocket<?>[] createOutputSockets(EventBus eventBus);
 
     /**
+     * Override this to provide persistent per step data
```
ThomasJClark: *per-step

PR #124, opened 2015-11-11
Update sources

ThomasJClark:
- Add support for Axis cameras (and other IP cameras)
- Change the appearance of the sourceview based on PDR feedback (the shadows *are* kind of ugly)

### DIFF
--- a/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
+++ b/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
@@ -81,7 +81,7 @@ public class CameraSource implements Source {
     /**
      * Starts the video capture from the
      *
-     * @param deviceNumber The index of the Webcam device that should be attached to
+     * @param grabber A JavaCV {@link FrameGrabber} instance to capture from
      */
     private synchronized void startVideo(FrameGrabber grabber) throws IOException {
         final OpenCVFrameConverter.ToMat convertToMat = new OpenCVFrameConverter.ToMat();
@@ -94,7 +94,7 @@ public class CameraSource implements Source {
         try {
             grabber.start();
         } catch (FrameGrabber.Exception e) {
-            throw new IOException("A problem occurred trying to start the frame grabber", e);
+            throw new IOException("A problem occurred trying to start the frame grabber for " + this.name, e);
         }
 
         // Store the grabber only once it has been started in the case that there is an exception.
--- a/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
+++ b/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
@@ -9,12 +9,10 @@ import edu.wpi.grip.core.Source;
 import edu.wpi.grip.core.events.FatalErrorEvent;
 import edu.wpi.grip.core.events.SourceRemovedEvent;
 import org.bytedeco.javacpp.opencv_core.Mat;
-import org.bytedeco.javacv.Frame;
-import org.bytedeco.javacv.FrameGrabber;
-import org.bytedeco.javacv.OpenCVFrameConverter;
-import org.bytedeco.javacv.OpenCVFrameGrabber;
+import org.bytedeco.javacv.*;
 
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


src/main/java/edu/wpi/grip/core/sources/CameraSource.java
```diff
@@ -71,9 +83,8 @@ public String getName() {
      *
      * @param deviceNumber The index of the Webcam device that should be attached to
      */
-    private synchronized void startVideo(final int deviceNumber) throws IOException {
+    private synchronized void startVideo(FrameGrabber grabber) throws IOException {
```
JLLeitschuh: I like this.

src/main/java/edu/wpi/grip/core/sources/CameraSource.java
```diff
@@ -83,7 +94,7 @@ private synchronized void startVideo(final int deviceNumber) throws IOException
         try {
             grabber.start();
         } catch (FrameGrabber.Exception e) {
-            throw new IOException("A problem occurred trying to start the frame grabber for device number " + deviceNumber, e);
+            throw new IOException("A problem occurred trying to start the frame grabber", e);
```
JLLeitschuh: This should indicate what device has caused the issue. Either the URL or the device number?

ThomasJClark: There's no common interface to get that for any frame grabber, and I think it's OK if we don't explicitly say that information

src/main/java/edu/wpi/grip/ui/pipeline/AddSourceView.java
```diff
@@ -86,11 +85,41 @@ public AddSourceView(EventBus eventBus) {
...
+            });
+        });
+
+        addButton("Add IP\nCamera", getClass().getResource("/edu/wpi/grip/ui/icons/add-webcam.png"), mouseEvent -> {
+            final Parent root = this.getScene().getRoot();
+
+            // Show a dialog for the user to pick a camera URL
+            final Dialog<ButtonType> dialog = new Dialog<>();
+            final TextField cameraAddress = new TextField();
+            cameraAddress.setPromptText("Ex: http://10.1.90.11/mjpg/video.mjpg");
```
JLLeitschuh: Text field should have validation to ensure the output is a valid URL.

src/main/java/edu/wpi/grip/core/sources/CameraSource.java
JLLeitschuh: No, but you have the information from above: `"IP Camera " + new URL(address).getHost()` and the device number. I think this will be helpful in issue messages that get posted.

src/main/java/edu/wpi/grip/ui/pipeline/AddSourceView.java
```diff
@@ -86,11 +87,55 @@ public AddSourceView(EventBus eventBus) {
...
+            // If the user clicks OK, add a new camera source
             dialog.showAndWait().filter(Predicate.isEqual(ButtonType.OK)).ifPresent(result -> {
                 loadSourceExecutor.execute(() -> {
                     try {
-                        final WebcamSource source = new WebcamSource(eventBus, cameraIndex.getValue());
+                        final CameraSource source = new CameraSource(eventBus, cameraIndex.getValue());
+                        eventBus.post(new SourceAddedEvent(source));
+                    } catch (IOException e) {
+                        eventBus.post(new FatalErrorEvent(e));
+                        e.printStackTrace();
```
JLLeitschuh: Print the stack trace first in case the event handling has an exception.

JLLeitschuh:
Other than that this looks good to me.

src/main/java/edu/wpi/grip/ui/pipeline/AddSourceView.java
ThomasJClark: eventBus.post never ever ever ever throw an exception

(also this code was written by you, I just copied those lines for the IP camera source)

PR #131, opened 2015-11-12
Adds appVeyor.yml file to run builds with

JLLeitschuh:


ThomasJClark:
Can you make it so we can get the GRIP.exe installer?

JLLeitschuh:
Working on that

PR #132, opened 2015-11-12
Adds travis deploy task for release tags

JLLeitschuh:


ThomasJClark:
Will this actually make a release on GitHub, or just produce a file we can upload?  And is it triggered whenever there's a push, or only on tagged releases?

ThomasJClark:
Also, can you do the same for the AppVeyor thing?  I assume it's basically the same

PR #140, opened 2015-11-13
Update README.md

ThomasJClark:
- That link doesn't actually lead to the javadocs, and I don't think people are gonna come here looking for javadocs anyways.
- That common problem shouldn't actually happen anymore, and it probably wasn't even common to begin with.
- Update the logo
- very early state -> early state :+1:
- Add a status badge for AppVeyor

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,31 +1,27 @@
-![logo](https://cloud.githubusercontent.com/assets/3964980/7551665/c133ce00-f65f-11e4-8d65-f4f122880b1c.png)
+![logo](https://cloud.githubusercontent.com/assets/3964980/11156885/6fa1967a-8a1c-11e5-8c78-e552ffba31c0.png)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,31 +1,27 @@
...
+## Building GRIP From Source
 
-## Docs
-Javadocs for the project can be found [here](http://WPIRoboticsProjects.github.io/GRIP)
+To build and run, use the included Gradle wrapper script.
 
-## Common Problems
-### UnsatisfiedLinkError
-The opencv library for linux requires version 4.8 or higher of gcc in order for `GOMP_4.0` to be present.
-Check the version of gcc you have installed using `gcc -v` and update if necessary.
```
JLLeitschuh: If you are going to remove this then please add a DEVELOPER.md and put this there.

PR #143, opened 2015-11-14
Improve operation searching

### LINKED ISSUE #113: search does not search step description
The function to search the palette list of steps does not search the step descriptions.

ThomasJClark:
Searching is now "fuzzy" - it will ignore differences in whitespace,
capitalization, and small typos.  It also takes into account both the
name of an operation and its description.
The visual glitch where the bottom of a filtered operation list gets get
off if the last operation's text overflows has been fixed (we just
needed to set the ScrollPane to always grow vertically)

Typing while the palette is focused will automatically focus the search
box.

closes #113

### DIFF
--- a/src/main/java/edu/wpi/grip/ui/PaletteView.java
+++ b/src/main/java/edu/wpi/grip/ui/PaletteView.java
@@ -4,6 +4,7 @@ import com.google.common.eventbus.EventBus;
 import com.google.common.eventbus.Subscribe;
 import edu.wpi.grip.core.Operation;
 import edu.wpi.grip.core.events.OperationAddedEvent;
+import edu.wpi.grip.ui.util.SearchUtility;
 import javafx.beans.InvalidationListener;
 import javafx.beans.property.ObjectProperty;
 import javafx.fxml.FXML;
@@ -14,6 +15,7 @@ import org.controlsfx.control.textfield.CustomTextField;
 import org.controlsfx.control.textfield.TextFields;
 
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 
 import static com.google.common.base.Preconditions.checkNotNull;
@@ -33,7 +35,6 @@ public class PaletteView extends VBox {
     @FXML
     private CustomTextField operationSearch;
 
-
     public PaletteView(EventBus eventBus) {
         checkNotNull(eventBus);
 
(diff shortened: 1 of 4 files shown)


src/main/java/edu/wpi/grip/ui/PaletteView.java
```diff
@@ -71,24 +71,20 @@ public PaletteView(EventBus eventBus) {
...
-        // TODO: Make this use a fuzzy search
-        final String lowerCaseSearchString = searchText.toLowerCase();
-        return lowerCaseSearchString.isEmpty() || name.toLowerCase().contains(lowerCaseSearchString);
+    private void setupClearButtonField(CustomTextField customTextField) {
+        try {
+            final Method m = TextFields.class.getDeclaredMethod("setupClearButtonField", TextField.class, ObjectProperty.class);
+            m.setAccessible(true);
+            m.invoke(null, customTextField, customTextField.rightProperty());
+        } catch (IllegalAccessException | InvocationTargetException | NoSuchMethodException e) {
+            e.printStackTrace();
```
JLLeitschuh: Rethrow with illegal state

ThomasJClark: Not being able to show a clear button that isn't guaranteed to be available to us isn't an illegal state IMO.  Logging a stack trace and letting the textfield otherwise keep working is a good way to gracefully handle that error.

src/main/java/edu/wpi/grip/ui/util/SearchUtility.java
```diff
@@ -0,0 +1,39 @@
...
+     * @return true if query if approximately a substring of text, not taking into account capitalization, whitespace,
+     * or minor typos
+     */
+    public static boolean fuzzyContains(String text, String query) {
+        if (query.isEmpty()) return true;
+        if (text.isEmpty()) return false;
+
+        // Normalize the capitalization and whitespace (which are currently both pretty inconsistent among operations)
+        text = text.toUpperCase().replaceAll("[^a-zA-Z]", "");
+        query = query.toUpperCase().replaceAll("[^a-zA-Z]", "");
```
JLLeitschuh: Why are you removing all of the space characters?
Why are you not keeping numbers around?
What about a quoted query? `"Threshold"` should prioritize over fuzzy search.

ThomasJClark: See the comment right before this line. The operations are currently very inconsistent about whitespace and other non-alpha characters. If you search "bitwiseAnd", "bitwise and", or "bitwise_and", it should return the same results.

I'm not sure that a quote syntax is a good idea.  How would people know to use that syntax?  And what situation would the regular fuzzy search not work?

PR #144, opened 2015-11-14
Add IDE plugins to buildscript

JacisNonsense:
Adds the ability to run ```gradlew idea``` and ```gradlew eclipse``` to setup a development environments with all dependencies linked and ready-to-go.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,11 @@ gradle-app.setting
 # IntelliJ
 out/
 
+# Eclipse
+.classpath
+.project
+.settings/
+
 # mpeltonen/sbt-idea plugin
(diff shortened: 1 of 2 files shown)


ThomasJClark:
At least with `gradlew idea`, this doesn't seem to include the generated code as part of the build, so I get `Error:(13, 30) java: package edu.wpi.grip.generated does not exist`

The generated code is still generated, it's just not added to the source sets in the IntelliJ project.  I assume the gradle idea plugin has some option to let you add source directories

JacisNonsense:
All generated SourceSets should automatically be added to IDEs now @ThomasJClark

PR #147, opened 2015-11-16
Splits Project into Core and ApplicationUI

JLLeitschuh:
This allows us to build the project in headless mode


ThomasJClark:
I don't know if this is a good approach or not, but let's put headless mode off until serialization is working so we can work on the tasks we have due this week. We have a schedule to stick to!

JLLeitschuh:
This is how gradle is designed. This is how it is supposed to be implemented.

ThomasJClark:
I didn't say there's anything wrong with it. Just that if it's gonna block all the stuff we want to get done now, and its not gonna work until we have serialization, we should wait. The way headless mode works is pretty important, and we should meet to come up with something.

core/src/main/java/edu/wpi/grip/core/operations/PythonScriptOperation.java
```diff
@@ -30,16 +33,16 @@
...
  * and returns the values for the outputs.  It can return a single value if there's one output, or a sequence type for
  * any number of values.
- *
+ * <p>
  * <pre>{@code
  * def perform(a, b):
  * return a + b
  * }</pre>
- *
+ * <p>
```
JLLeitschuh: Inserted by autoformat. This was not me

PR #149, opened 2015-11-16
[RESUBMIT] Splits Project into Core and ApplicationUI

JLLeitschuh:
This allows us to build the project in headless mode


JLLeitschuh:
Brad wants this done. Just chatted with him. He imparted to me how important getting this to work was.

.travis.yml
```diff
@@ -20,7 +20,7 @@ before_install:
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
 
-install: ./gradlew assemble --stacktrace
+install: ./gradlew :applicationUi:assemble --stacktrace
```
ThomasJClark: This should be "ui", not "applicationUi" to match everthing else (such as the package name)

README.md
```diff
@@ -24,4 +24,5 @@ Check out the release notes and [the wiki](https://github.com/WPIRoboticsProject
 
 To build and run, use the included Gradle wrapper script.
 
-    ./gradlew run
+ - With UI `./gradlew :applicationUi:run`
+ - Headless `./gradlew :core:run`
```
ThomasJClark: Does headless mode do anything yet?  If not, we should not include it in the readme

applicationUi/src/main/java/edu/wpi/grip/ui/pipeline/AddSourceView.java
```diff
@@ -3,8 +3,8 @@
 import com.google.common.eventbus.EventBus;
 import edu.wpi.grip.core.events.FatalErrorEvent;
 import edu.wpi.grip.core.events.SourceAddedEvent;
-import edu.wpi.grip.core.sources.ImageFileSource;
 import edu.wpi.grip.core.sources.CameraSource;
+import edu.wpi.grip.core.sources.ImageFileSource;
```
ThomasJClark: Please get rid of this commit noise.  This will just make merging stuff more frustrating.

applicationUi/src/main/java/edu/wpi/grip/ui/preview/PointSizeSocketPreviewView.java
```diff
@@ -9,7 +9,8 @@
 import javafx.scene.layout.GridPane;
 import org.bytedeco.javacpp.IntPointer;
 
-import static org.bytedeco.javacpp.opencv_core.*;
+import static org.bytedeco.javacpp.opencv_core.Point;
```
ThomasJClark: Get rid of this commit noise

build.gradle
```diff
@@ -1,97 +1,118 @@
-apply plugin: 'java'
 apply plugin: 'idea'
 apply plugin: 'eclipse'
-apply plugin: 'application'
 apply plugin: 'jacoco'
-apply from: 'http://dl.bintray.com/shemnon/javafx-gradle/8.1.1/javafx.plugin'
 
-mainClassName = "edu.wpi.grip.Main"
-version = '0.1.0'
+allprojects {
```
ThomasJClark: Is there a reason to reorder these statements?  If not, you should revert the order

buildSrc/src/main/groovy/edu/wpi/gripgenerator/CodeGenerator.groovy
```diff
@@ -26,8 +26,7 @@ class CodeGenerator extends DefaultTask {
     @TaskAction
     def runAction() {
         //Get the target directory
-        LinkedHashSet destSet = dest;
-        String targetDirectoryString = destSet.getAt(0).toString()
+        String targetDirectoryString = dest;
```
ThomasJClark: What does this change mean?

core/src/main/java/edu/wpi/grip/Main.java
```diff
@@ -11,46 +12,42 @@
...
+    static {
+        eventBus = new EventBus((exception, context) -> {
+            Main.eventBus.post(new FatalErrorEvent(exception));
+        });
+        core = new Core(eventBus, new Pipeline(eventBus), () -> loadAllOperations());
+    }
 
-    @FXML
-    private ScrollPane bottomPane;
+    public static class Core {
```
ThomasJClark: This class does not need to exist.  Just make a new pipeline in the main function of the UI and in the main function of headless mode

buildSrc/src/main/groovy/edu/wpi/gripgenerator/CodeGenerator.groovy
JLLeitschuh: It's now being passed a string from the gradle build file instead. Not a list of files

core/src/main/java/edu/wpi/grip/Main.java
```diff
@@ -11,46 +12,42 @@
...
+            this.eventBus = eventBus;
+            this.pipeline = pipeline;
+            this.loadOperations = loadOperations;
         }
+    }
 
-        final PreviewsView previewsView = new PreviewsView(eventBus);
-        final PaletteView paletteView = new PaletteView(eventBus);
-        final PipelineView pipelineView = new PipelineView(eventBus, new Pipeline(eventBus));
+    protected static Core getCore() {
```
ThomasJClark: Singletons aren't cool.  This will prevent us from ever having multiple projects open

build.gradle
JLLeitschuh: Order?

ThomasJClark: yeah.  does that need clarification?

core/src/main/java/edu/wpi/grip/Main.java
JLLeitschuh: The UI shouldn't create these things. They are specifically from the core.

core/src/main/java/edu/wpi/grip/Main.java
```diff
@@ -11,46 +12,42 @@
...
+        return core;
+    }
 
-        this.topPane.getItems().addAll(previewsView, paletteView);
-        this.bottomPane.setContent(pipelineView);
-        pipelineView.prefHeightProperty().bind(this.bottomPane.heightProperty());
+    public static void main(final String... args) throws NoSuchMethodException, InvocationTargetException, IllegalAccessException {
```
ThomasJClark: So, headless mode just prints "Running in headless mode"?  This doesn't seem very usefull

core/src/main/java/edu/wpi/grip/Main.java
ThomasJClark: Ok, I disagree with that, but if you think we should refactor things like that, please do it in a separate PR.  This PR should just create a separate module for the UI and core.

Since this separation is apparently very important to do right now, it's best not to include controversion design changes along with it that will keep it in review for a long time.

build.gradle
JLLeitschuh: All projects is at the top because it defines what rules apply to all subprojects of this one.

core/src/main/java/edu/wpi/grip/Main.java
JLLeitschuh: Okay. That makes sense. I'll have to think about how to fix that.

core/src/main/java/edu/wpi/grip/Main.java
JLLeitschuh: At the moment yep!
I mean I could put in a demo script to see it working perhaps.

ThomasJClark: It should really just open a project file and run it.  That's the use case for headless mode

core/src/main/java/edu/wpi/grip/Main.java
ThomasJClark: Don't have a `Core` object, just have the main classes create what objects they need.

We might want to have projects share a palette, but they would definitely not share a pipeline.  I'm not sure if they'd share an event bus or not.  There's no justification to group these into a single object, and it'll just get in the way of future development

core/src/main/java/edu/wpi/grip/Main.java
JLLeitschuh: This has to be split so the core has access to these variables.

JLLeitschuh:
Now if you run the demo it uses the axis camera.

PR #156, opened 2015-11-18
Make Dependencies Platform-Dependent

JacisNonsense:
Prevents the JavaCPP OpenCV presets being fetched for all platforms, fetching only what is required for the User's machine.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,12 +1,3 @@
-buildscript {
-  repositories {
-    mavenCentral()
-  }
-  dependencies {
-    classpath 'com.google.gradle:osdetector-gradle-plugin:1.4.0'
-  }
-}
-
 plugins {
     id 'com.github.johnrengelman.shadow' version '1.2.2'
     id 'java'
@@ -14,6 +5,7 @@ plugins {
     id 'eclipse'
     id 'application'
     id 'jacoco'
+    id 'com.google.osdetector' version '1.4.0'
 }
 
 apply from: 'http://dl.bintray.com/shemnon/javafx-gradle/8.1.1/javafx.plugin'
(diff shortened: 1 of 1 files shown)


ThomasJClark:
Looks good, just need to test this on each platform. @JLLeitschuh can you confirm this works on OS X?  Travis and AppVeyor will tell us about linux and windows.

One change I would suggest in the meantime is using the `plugins { ... }` syntax for importing the plugin.  So, if you just add `id "com.google.osdetector" version "1.4.0"` inside of the plugins block, you don't need to explicitly add any buildscript dependencies.

JacisNonsense:
@ThomasJClark for some unknown reason Gradle will fail when putting the identical plugin id and version number inside of the `plugins { }` block, but works fine when using `apply plugin`

ThomasJClark:
This works for me:

    plugins {
        id 'com.github.johnrengelman.shadow' version '1.2.2'
        id 'java'
        id 'idea'
        id 'eclipse'
        id 'application'
        id 'jacoco'
        id "com.google.osdetector" version "1.4.0"
    }

build.gradle
```diff
@@ -5,9 +5,11 @@ plugins {
     id 'eclipse'
     id 'application'
     id 'jacoco'
+    id 'com.google.osdetector' version '1.4.0'
 }
 
 apply from: 'http://dl.bintray.com/shemnon/javafx-gradle/8.1.1/javafx.plugin'
+apply plugin: 'com.google.osdetector'
```
ThomasJClark: I think we can leave this out since it's specified in the `plugins` thing

JacisNonsense: Whoops, fair point. Missed that

JLLeitschuh:
Not on OSX:
```
FAILURE: Build failed with an exception.

* What went wrong:
Could not resolve all dependencies for configuration ':compile'.
> Could not find opencv-osx-x86_64.jar (org.bytedeco.javacpp-presets:opencv:3.0.0-1.1).
  Searched in the following locations:
      https://repo1.maven.org/maven2/org/bytedeco/javacpp-presets/opencv/3.0.0-1.1/opencv-3.0.0-1.1-osx-x86_64.jar

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED
```

JLLeitschuh:
That works now on OSX

PR #157, opened 2015-11-18
Appveyor now runs tests on Windows

JLLeitschuh:


ThomasJClark:
It doesn't look like AppVeyor is collecting the results of the test.

https://ci.appveyor.com/project/ThomasJClark/grip/build/0.1.0-110/tests

JLLeitschuh:
Its not supposed to collect the test results, its just supposed to run the tests. I mean I can look into that if you want. Its not hugely important.

JLLeitschuh:
Okay, it is possible, I'll do this later http://www.appveyor.com/docs/running-tests#test-results

JLLeitschuh:
The above link is actually more complicated than it seems. I don't have time to figure it out. I added a quick fix to tell you what tests failed.

PR #158, opened 2015-11-18
Serialization

ThomasJClark:


src/main/java/edu/wpi/grip/core/Palette.java
```diff
@@ -0,0 +1,47 @@
...
+ * The palette is a library of operations that can be added as steps in the {@link Pipeline}
+ */
+public class Palette {
+
+    private final EventBus eventBus;
+    private final Map<String, Operation> operations = new HashMap<>();
+
+    public Palette(EventBus eventBus) {
+        checkNotNull(eventBus);
+        this.eventBus = eventBus;
```
JLLeitschuh: This could be `this.eventBus = checkNotNull(eventBus, "The Event Bus can not be null");`

src/test/java/edu/wpi/grip/core/serialization/SerializationTest.java
```diff
@@ -17,8 +17,8 @@
 import static org.junit.Assert.assertEquals;
 
 public class SerializationTest {
-    private final EventBus eventBus = new EventBus();
-    private final Serialization serialization = new Serialization(eventBus);
+/*    private final EventBus eventBus = new EventBus();
```
JLLeitschuh: Can you fix the tests instead of just commenting all of them out?

src/test/java/edu/wpi/grip/core/PaletteTest.java
```diff
@@ -0,0 +1,62 @@
...
+import org.junit.Test;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Optional;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Created by tom on 11/15/15.
```
JLLeitschuh: Atribution

src/main/java/edu/wpi/grip/core/Pipeline.java
```diff
@@ -41,6 +44,21 @@ public void register() {
     }
 
     /**
+     * Remove everything in the pipeline
+     */
+    public void clear() {
+        this.steps.stream()
+                .map(StepRemovedEvent::new)
+                .collect(Collectors.toList())
```
JLLeitschuh: Do you really need to collect them first?

src/main/java/edu/wpi/grip/core/Pipeline.java
```diff
@@ -41,6 +44,21 @@ public void register() {
...
+                .forEach(this.eventBus::post);
+
+        this.sources.stream()
+                .map(SourceRemovedEvent::new)
+                .collect(Collectors.toList())
```
JLLeitschuh: Ditto

src/test/java/edu/wpi/grip/core/serialization/SerializationTest.java
ThomasJClark: Um, no?  I'm almost completely re-writing the serialization stuff, so these tests probably won't even compile.  When this is ready for review, all amend it with real tests.

src/main/java/edu/wpi/grip/core/Pipeline.java
ThomasJClark: yes

src/main/java/edu/wpi/grip/core/Step.java
```diff
@@ -12,6 +13,7 @@
  * A step is an instance of an operation in a pipeline.  A step contains a list of input and output sockets, and it
  * runs the operation whenever one of the input sockets changes.
  */
+@XStreamAlias(value = "grip:Step")
```
JLLeitschuh: You can access static final variables in annotations. Don't know if you want to do that here.

src/test/java/edu/wpi/grip/core/serialization/SerializationTest.java
JLLeitschuh: Oh, this isn't ready. Sorry.

src/main/java/edu/wpi/grip/core/Step.java
ThomasJClark: no, probably not

src/main/java/edu/wpi/grip/core/Pipeline.java
JLLeitschuh: Why are you still collecting them first then?

src/main/java/edu/wpi/grip/core/serialization/SocketConverter.java
```diff
@@ -0,0 +1,170 @@
...
+            });
+
+            // Save whether or not output sockets are published and previewed
+            if (socket.getDirection() == Socket.Direction.OUTPUT) {
+                writer.addAttribute(PUBLISHED_ATTRIBUTE, String.valueOf(((OutputSocket) socket).isPublished()));
+                writer.addAttribute(PREVIEWED_ATTRIBUTE, String.valueOf(((OutputSocket) socket).isPreviewed()));
+            }
+
+            // Save the value of input sockets that could possibly have been set with the GUI
+            if (socket.getDirection() == Socket.Direction.INPUT
```
JLLeitschuh: .equals

src/main/java/edu/wpi/grip/core/serialization/SocketConverter.java
```diff
@@ -0,0 +1,170 @@
...
+                    && socket.getConnections().isEmpty()
+                    && socket.getSocketHint().getView() != SocketHint.View.NONE) {
```
JLLeitschuh: .equals

src/main/java/edu/wpi/grip/core/serialization/SocketConverter.java
```diff
@@ -0,0 +1,170 @@
...
+                } else {
+                    throw new IllegalArgumentException("Unexpected node in socket: " + reader.getNodeName());
+                }
+                reader.moveUp();
+            }
+
+            reader.moveUp();
+
+            return socket;
+        } catch (Exception e) {
```
JLLeitschuh: Don't catch the generic exception

src/main/java/edu/wpi/grip/core/Pipeline.java
ThomasJClark: Sending a `StepRemovedEvent` causes `this.steps` to be modified, but bad things happen when streams modify their source.   So, it has to be a `forEach` on a separate list, not on a stream of that list.

src/test/java/edu/wpi/grip/core/serialization/ProjectTest.java
```diff
@@ -172,11 +173,10 @@ public void testPerformSerializedPipeline() throws Exception {
...
     @SuppressWarnings("unchecked")
     public void testPerformSerializedPipelineWithMats() throws Exception {
-        Pipeline pipeline1 = new Pipeline(eventBus);
-        eventBus.post(new StepAddedEvent(new Step(eventBus, new AddOperation())));
-        Pipeline pipeline2 = serializeAndDeserialize(pipeline1);
+        eventBus.post(new StepAddedEvent(new Step(eventBus, opencvAddOperation)));
+        serializeAndDeserialize();
 
-        Step step1 = pipeline2.getSteps().get(0);
+        Step step1 = pipeline.getSteps().get(0);
```
JLLeitschuh: Could be final

JLLeitschuh: Nvm, this is a test, I don't care all that much

src/main/java/edu/wpi/grip/core/serialization/SocketConverter.java
ThomasJClark: For enums, it doesn't matter.

src/main/java/edu/wpi/grip/core/serialization/SocketConverter.java
ThomasJClark: This is intentionally so that any exception is rethrown as a `ConversionException`.

src/main/java/edu/wpi/grip/core/serialization/Project.java
```diff
@@ -0,0 +1,76 @@
...
+        this.pipeline = pipeline;
+
+        this.xstream.registerConverter(new StepConverter(eventBus, palette));
+        this.xstream.registerConverter(new SourceConverter(eventBus, xstream.getMapper()));
+        this.xstream.registerConverter(new SocketConverter(xstream.getMapper(), pipeline));
+
+        this.xstream.registerConverter(new ConnectionConverter(eventBus));
+        this.xstream.registerConverter(new InjectedObjectConverter<>(eventBus), XStream.PRIORITY_VERY_HIGH);
+        this.xstream.registerConverter(new PythonScriptOperationConverter(), XStream.PRIORITY_VERY_HIGH);
+        this.xstream.registerConverter(new MatConverter());
```
JLLeitschuh: What about size, point and scalar?

ThomasJClark: Size, point, and scalar are never serialized.

Neither is `Mat`.  That was dead code from the previous serialization implementation, which I've removed now.

src/main/java/edu/wpi/grip/core/serialization/ConnectionConverter.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Override
+    public Object unmarshal(HierarchicalStreamReader reader, UnmarshallingContext context) {
+        OutputSocket<?> outputSocket = null;
+        InputSocket<?> inputSocket = null;
+
+        // Read the child nodes of this connection. An output socket and an input socket should be defined.
+        while (reader.hasMoreChildren()) {
+            Socket<?> tmp = (Socket<?>) context.convertAnother(null, Socket.class);
```
JLLeitschuh: Could make this final if you want.

src/main/java/edu/wpi/grip/core/Pipeline.java
JLLeitschuh: You should leave a comment about why you are doing this this way then. Use the above comment.

src/main/java/edu/wpi/grip/core/serialization/SocketConverter.java
JLLeitschuh: There should be a better way than this but okay.

src/main/java/edu/wpi/grip/core/serialization/SocketConverter.java
```diff
@@ -0,0 +1,170 @@
...
+    private Class<?> getDeserializedType(Socket socket) {
+        final Class<?> socketType = socket.getSocketHint().getType();
+
+        if (!socketType.isInterface() && !Modifier.isAbstract(socketType.getModifiers())) {
+            return socketType;
+        } else if (socketType.equals(List.class)) {
+            return ArrayList.class;
+        } else if (socketType.equals(Number.class)) {
+            return Double.class;
+        }
```
JLLeitschuh: This feels wrong. Hard coding it so that every time you use a list it converts it to an array list seems like it could create unexpected behavior for future developers.

src/main/java/edu/wpi/grip/core/Socket.java
```diff
@@ -96,19 +98,36 @@ public T getValue() {
...
+     * @see #getSource()
      */
     public Optional<Step> getStep() {
         return step;
     }
 
+    /**
+     * @param source The source that this socket is part of, if it's in a source.
+     */
+    protected void setSource(Optional<Source> source) {
```
JLLeitschuh: Is there some common interface between Source and Step we could use so that we don't need to have two optional fields?
That would let socket always have an "owner"

src/main/java/edu/wpi/grip/core/Source.java
```diff
@@ -1,19 +1,48 @@
...
+import java.io.IOException;
+import java.util.Optional;
+import java.util.Properties;
+
 /**
- * Common interface for an input into the pipeline.
+ * Base class for an input into the pipeline.
  */
-public interface Source {
+public abstract class Source {
```
JLLeitschuh: I like this

src/main/java/edu/wpi/grip/core/serialization/ConnectionConverter.java
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        if (inputSocket == null || outputSocket == null) {
+            throw new ConversionException("Input and/or output was not specified in connection");
+        }
+
+        // Send the new connection as an event, so the GUI and other classes can listen for new connections even if regardless
+        // of if they came from a serialized file or not.
+        this.eventBus.post(new ConnectionAddedEvent(new Connection(this.eventBus, outputSocket, inputSocket)));
+        return null;
```
JLLeitschuh: Why are you returning null to the marshal/unmarshal methods?

JLLeitschuh:
What happened to the xstream converter for Mats? No longer necessary? Why not?

src/main/java/edu/wpi/grip/core/serialization/SocketConverter.java
ThomasJClark: Nope.  If a future developer creates a socket hint with type `List`, it shouldn't be unexpected that that socket gets assigned a value of type `ArrayList`, since `ArrayList` implements `List`.

Also, it doesn't convert socket values.  If you read the comments, it converts the *serialized* data into an ArrayList.  That means that if you have a socket value like this:

    <value>
        <double>1</double>
        <double>2</double>
        <double>3</double>
    </value>

... that we should store 1, 2, and three in an ArrayList.  There is literally no way to avoid choosing a concrete list class to use.

src/main/java/edu/wpi/grip/core/Socket.java
ThomasJClark: No, there isn't currently one.

ThomasJClark:
They're not necessary because we don't have any reason to ever serialize mats.

src/main/java/edu/wpi/grip/core/serialization/ConnectionConverter.java
ThomasJClark: Yeah, because we add things using events, not by returning it and having XStream directly inject it into the pipeline.  This is so we actually notify the GUI that something's changing in the pipeline.

Issue #129, opened 2015-11-11
Set values on step when focus is lost in a input box


JLLeitschuh:
This is more intuitive than having to press the enter key in a field to commit the value to the step.

PR #160, opened 2015-11-20
Text field of spinner now updates socket value when focus is lost.

PaulaRudy:
Would close https://trello.com/c/ItVmvdMv (trello card 74) and issue #129


JLLeitschuh:
This works. Merging

PR #162, opened 2015-11-22
Cleaned up the step descriptions.

PaulaRudy:


src/main/java/edu/wpi/grip/core/operations/composite/BlurOperation.java
```diff
@@ -48,7 +48,7 @@ public String getName() {
 
     @Override
     public String getDescription() {
-        return "Soften the details of an image to remove noise";
+        return "Soften the details of a Mat to remove noise.";
```
ThomasJClark: I would leave this as-is.  It's much clearer to the user to say "image" than "Mat".

ThomasJClark:
For an operation called *X*, a useful description should say more than "Perform *X*".  We should have a concise way of saying *what* it's doing, not just restating the name of the operation.

PaulaRudy:
"Image" was changed to "Mat" in order to be more accurate and consistent with the other descriptions. Unfortunately, many descriptions cannot contain a more thorough explanation without greatly increasing the length to the point where the layout is affected.

ThomasJClark:
> "Image" was changed to "Mat" in order to be more accurate and consistent with the other descriptions. 

Ah, okay.  Consistency is a good idea, but without a doubt it would be better to change the onces that say "Mat" into "image" than the other way around.  I think the Mat descriptions are left over from copy/pasted javadocs.

> Unfortunately, many descriptions cannot contain a more thorough explanation without greatly increasing the length to the point where the layout is affected.

We could definitely at least come up with a concise summary of an operation in those cases. For example, erosion could be descripted as "Shrink the light areas of an image".

PaulaRudy:
>Ah, okay. Consistency is a good idea, but without a doubt it would be better to change the onces that say "Mat" into "image" than the other way around. I think the Mat descriptions are left over from copy/pasted javadocs.

Can do.


>We could definitely at least come up with a concise summary of an operation in those cases. For example, erosion could be descripted as "Shrink the light areas of an image".

Ok, I'll take another look.

PR #170, opened 2015-11-29
Added a new class to handle non fatal errors.

PaulaRudy:
This is associated with trello card 86: https://trello.com/c/GbrSIOPr

### DIFF
--- a/src/main/java/edu/wpi/grip/core/events/NotificationEvent.java
+++ b/src/main/java/edu/wpi/grip/core/events/NotificationEvent.java
@@ -1,0 +1,31 @@
+package edu.wpi.grip.core.events;
+
+import java.util.Optional;
+
+/**
+ * Should be posted when a non fatal user cause error occurs.
+ * This is used to communicate to the main window that a
+ * error should be displayed unobtrusively to the user.
+ * The message that is displayed is the message in the Throwable.
+ */
+public class NotificationEvent {
+    private final Throwable exception;
+    private final Optional<String> detailedDescription;
+
+    public NotificationEvent(Throwable exception){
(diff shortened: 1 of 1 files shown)


src/main/java/edu/wpi/grip/core/events/NotificationEvent.java
```diff
@@ -0,0 +1,31 @@
...
+public class NotificationEvent {
+    private final Throwable exception;
+    private final Optional<String> detailedDescription;
+
+    public NotificationEvent(Throwable exception){
+        this(exception, null);
+    }
+
+    public NotificationEvent(Throwable exception, String detailedDescription) {
+        this.exception = exception;
```
JLLeitschuh: Use `checkNotNull(exception, "The exception can not be null");` from googles api

JLLeitschuh:
Also, look at the FXML changes I added to the main window:
Can you implement those too: https://github.com/WPIRoboticsProjects/GRIP/pull/84/files
Also, the event handing in the main window.

PaulaRudy:
Shouldn't you do this then? Since it's really just re-implementing something you already did?

PaulaRudy:
I mean, if I'm really just copy-pasting this, shouldn't you just be getting the old pull request up to date and we can merge that instead?

JLLeitschuh:
That PR is very very out of date. I don't have the time to bring it up to date. (Thats why I asked you to do it). If I was redoing it I would be copying and pasting and creating a new PR anyways.
You just need to implement the event handler inside of the controller.

PaulaRudy:
Ahhh Ok

PR #176, opened 2015-12-02
Fix ImageConverter incorrect assertion

JLLeitschuh:


src/main/java/edu/wpi/grip/ui/util/ImageConverter.java
```diff
@@ -34,12 +35,16 @@ public Image convert(Mat mat) {
...
         final int channels = mat.channels();
+        final int type = mat.type();
 
         assert channels == 3 || channels == 1 :
                 "Only 3-channel BGR images or single-channel grayscale images can be converted";
 
-        assert mat.type() == CV_8U || mat.type() == CV_8S :
-                "Only images with 8 bits per channel can be previewed";
+        // Note: If more channels are supported than this will need to be updated
+        assert Arrays.asList(
```
ThomasJClark: So, I know streams are cool and all, but could you just rewrite this to use `||`?  So it's more readable.

src/main/java/edu/wpi/grip/ui/util/ImageConverter.java
```diff
@@ -34,12 +35,16 @@ public Image convert(Mat mat) {
...
+                CV_8UC1, CV_8UC3,
+                CV_8SC1, CV_8SC3
+        ).stream().anyMatch(e -> e == type) : "Only images with 8 bits per channel can be previewed";
```
ThomasJClark: I think you can just do `mat.depth() == CV_8U || mat.depth() == CV_8S`

PR #177, opened 2015-12-03
Socket and SocketHint Value to Optional refactor

JLLeitschuh:


src/main/java/edu/wpi/grip/core/SocketHint.java
```diff
@@ -15,60 +19,31 @@
...
-     *                             property does not make sense for all types and is left unspecified for some.
-     * @param initialValueSupplier A function that returns an initial value for newly-created sockets.
+     *                             property does not make sense for all types and is left unspecified for some
      * @param publishable          If this isn't true, the socket won't be able to be set as published, and the GUI
      *                             should also not show a button to publish it.  This is false by default.
      */
-    public SocketHint(String identifier, Class<T> type, Supplier<T> initialValueSupplier, View view, T[] domain,
-                      boolean publishable) {
-        this.identifier = identifier;
+    private SocketHint(Class<T> type, String identifier, Optional<Supplier<T>> initialValueSupplier, View view, Optional<T[]> domain, boolean publishable) {
```
ThomasJClark: Why switch the order of `type` and `identifier`?

JLLeitschuh: Because it is required in the constructor for the builder, the rest are not.

src/main/java/edu/wpi/grip/ui/util/ImageConverter.java
```diff
@@ -37,7 +37,7 @@ public Image convert(Mat mat) {
 
         assert channels == 3 || channels == 1 :
                 "Only 3-channel BGR images or single-channel grayscale images can be converted";
-
+        System.out.println(mat.type());
```
JLLeitschuh: :+1:

src/main/java/edu/wpi/grip/core/operations/Operations.java
```diff
@@ -0,0 +1,31 @@
...
+
+import com.google.common.eventbus.EventBus;
+import edu.wpi.grip.core.events.OperationAddedEvent;
+import edu.wpi.grip.core.operations.composite.*;
+import edu.wpi.grip.core.operations.opencv.MatFieldAccessor;
+import edu.wpi.grip.core.operations.opencv.MinMaxLoc;
+import edu.wpi.grip.core.operations.opencv.NewPointOperation;
+import edu.wpi.grip.core.operations.opencv.NewSizeOperation;
+
+public class Operations {
```
JLLeitschuh: :+1:

src/main/java/edu/wpi/grip/core/Step.java
```diff
@@ -78,13 +78,25 @@ protected Pipeline getPipeline() {
         return this.pipeline;
     }
 
+    private synchronized void runPerformIfPossible() {
+        for (InputSocket<?> inputSocket : inputSockets) {
```
ThomasJClark: Definitely comment this to explain why we do this (I know the answer, but I don't think it's immediately obvious)

src/main/java/edu/wpi/grip/core/Step.java
```diff
@@ -78,13 +78,25 @@ protected Pipeline getPipeline() {
...
+            if (!inputSocket.getValue().isPresent()) {
+                for (OutputSocket<?> outputSocket : outputSockets) {
+                    outputSocket.resetValueToInitial();
+                }
+                return;
```
ThomasJClark: Should this just return?  Or at least print something

src/main/java/edu/wpi/grip/core/SocketHints.java
```diff
@@ -0,0 +1,105 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.function.Supplier;
+
+
+/**
+ * Create default socket hints
+ */
+public final class SocketHints {
```
ThomasJClark: Would it be possible to have a `createEnumSocketHint` factory method?

src/main/java/edu/wpi/grip/core/SocketHints.java
```diff
@@ -0,0 +1,105 @@
...
+    public static SocketHint<Size> createSizeSocketHint(final String identifier, final boolean withDefault) {
+        return createObjectSocketHintBuilder(identifier, Size.class, Size::new, withDefault).build();
+    }
+
+
+    public static SocketHint<Point> createPointSocketHint(final String identifier, final boolean withDefault) {
+        return createObjectSocketHintBuilder(identifier, Point.class, Point::new, withDefault).build();
+    }
+
+    public static SocketHint<Number> createNumberSliderSocketHint(final String identifier, final Number number,
```
ThomasJClark: Since the domain for sliders has to be two values, it would be safer to make this take two numbers instead of an array

src/main/java/edu/wpi/grip/core/SocketHints.java
```diff
@@ -0,0 +1,105 @@
...
+                                                            final Number[] domain) {
+        return createNumberSocketHintBuilder(identifier, number).domain(domain).view(SocketHint.View.SLIDER).build();
+    }
+
+    public static SocketHint<Number> createNumberSpinnerSocketHint(final String identifier, final Number number,
```
ThomasJClark: Same as [:arrow_up:](https://github.com/WPIRoboticsProjects/GRIP/pull/177/files#r46576528)

src/main/java/edu/wpi/grip/core/SocketHints.java
```diff
@@ -0,0 +1,105 @@
...
+                .initialValueSupplier(() -> Arrays.asList(low, high))
+                .domain(new List[]{Arrays.asList(low, high)})
+                .view(SocketHint.View.RANGE)
+                .build();
+    }
+
+    private static <T> SocketHint.Builder<T> createObjectSocketHintBuilder(final String identifier,
+                                                                   final Class<T> type,
+                                                                   final Supplier<T> supplier,
+                                                                   final boolean withDefault) {
```
ThomasJClark: Instead of having `withDefault`, would it work to just make the supplier an optional parameter, and assume there's no default if it's not specified?

src/main/java/edu/wpi/grip/core/Step.java
JLLeitschuh: Good call. Thanks

src/main/java/edu/wpi/grip/core/Step.java
JLLeitschuh: This will eventually be replaced with the event posting but okay. I can take care of that.

src/main/java/edu/wpi/grip/core/SocketHints.java
JLLeitschuh: Okay, makes sense.

src/main/java/edu/wpi/grip/core/SocketHints.java
JLLeitschuh: This is private and is used internally.

src/main/java/edu/wpi/grip/core/SocketHints.java
JLLeitschuh: What do you mean? Can you provide an example?

ThomasJClark: Something so that [this](https://github.com/WPIRoboticsProjects/GRIP/pull/177/files#diff-35295a6e1bb5368122de3caef3564decR34) could be replaced with `SocketHints.createEnumSocketHint("Type", Type.class)`.

ThomasJClark:
We should have a `SocketHints.createEnumSocketHint` method also

JLLeitschuh:
> We should have a SocketHints.createEnumSocketHint method also

Sure, I'll do that in a separate PR though. This is blocking a lot of stuff I need to do, that won't hold up anything.

ThomasJClark:
Why a separate PR just for enums?  Every other type is in the one

JLLeitschuh:
Resolved

src/main/java/edu/wpi/grip/core/SocketHints.java
```diff
@@ -0,0 +1,185 @@
...
+
+        public static SocketHint<Number> createNumberSocketHint(final String identifier, Number defaultValue) {
+            return new SocketHint.Builder(Number.class)
+                    .identifier(identifier)
+                    .initialValue(defaultValue)
+                    .publishable(true).build();
+        }
+    }
+
+    public static <T extends Enum<T>> SocketHint<T> createEnumSocketHint(final String identifier,
```
ThomasJClark: Seems kind of redundant to require the values to be explicitly passed in, since it's for enums.  Would we ever want to make a drop-down thing with only some of an enum's values?

JLLeitschuh: You can't access static methods on generics.

ThomasJClark: I don't see why you need to

    ... createEnumSocketHint(String id, Class<T> type) {
            return new SocketHint.Builder()
                    .identifier(id)
                    .domain(type.getEnumConstants())
                    .initialValue(type.getEnumConstants()[0])
                    .view(SocketHint.View.SELECT)
                    .build();
    }

(And probably a version with an explicit default value too, but you get the point)

That way, we can really easily make a socket hint automatically for any enum:

    private final SocketHint<Whatever> whateverHint = SocketHints.createEnumSocketHint("something", Whatever.class);

JLLeitschuh: Okay, resolve it. Look now.

ThomasJClark:
I get `No enum constant edu.wpi.grip.core.operations.composite.BlurOperation.Type.` or something similar whenever I load a project (this happens before the last enum commit also)

JLLeitschuh:
Works fine for me.

ThomasJClark:
Can you post the project you're loading?

JLLeitschuh:
Oh, I see the problem now.

JLLeitschuh:
Again, can we leave this enum thing to another PR.

JLLeitschuh:
Okay, so this is a problem with serialization. Not the socket hints.

JLLeitschuh:
Fixed serialization too.

JLLeitschuh:
Pull request against this pull request: https://github.com/ThomasJClark/GRIP/pull/1

src/main/java/edu/wpi/grip/core/SocketHints.java
PaulaRudy: >Would we ever want to make a drop-down thing with only some of an enum's values?

Actually, yes. In [threshold(http://docs.opencv.org/3.0.0/d7/d1b/group__imgproc__misc.html#gae8a4a146d1ca78c626a53577199e9c57), some enums require different input values.

src/main/java/edu/wpi/grip/core/Step.java
```diff
@@ -78,13 +79,50 @@ protected Pipeline getPipeline() {
...
+        }
+    }
+
+    /**
+     * The {@link Operation#perform} method should only be called if all {@link InputSocket#getValue()} are not empty.
+     * If one input is invalid then the perform method will not run and all output sockets will be assigned to their
+     * default values.
+     */
+    private synchronized void runPerformIfPossible() {
+        try {
```
ThomasJClark: This could be rewritten like this to be much clearer and not use exceptions for control flow:

    for (InputSocket<?> socket : this.inputSockets) {
        if (!socket.getValue().isPresent()) {
            this.resetOutputSockets();
            return;
        }
    }

src/main/java/edu/wpi/grip/ui/pipeline/input/CheckboxInputSocketView.java
```diff
@@ -27,10 +29,14 @@ public CheckboxInputSocketView(EventBus eventBus, InputSocket<Boolean> socket) {
         this.getIdentifier().setContentDisplay(ContentDisplay.RIGHT);
     }
 
+    private void assignSocketValue(final Optional<Boolean> value) {
+        this.checkBox.setSelected(value.isPresent() && value.get());
```
ThomasJClark: This method does not assign the socket value, so it shouldn't be named like this.  Better yet, this line could just not be abstracted into a separate function, since it's one line and it's only used twice

src/main/java/edu/wpi/grip/ui/pipeline/input/NumberSpinnerInputSocketView.java
```diff
@@ -29,7 +29,7 @@ public NumberSpinnerInputSocketView(EventBus eventBus, InputSocket<Number> socke
         super(eventBus, socket);
 
 
-        Object[] domain = socket.getSocketHint().getDomain();
+        Object[] domain = socket.getSocketHint().getDomain().orElseThrow(() -> new IllegalStateException("No domain was supplied"));
         if (domain == null) {
```
ThomasJClark: `domain` cannot be `null`, since you got the value from an `Optional`

src/main/java/edu/wpi/grip/ui/pipeline/input/RangeInputSocketView.java
```diff
@@ -27,8 +27,8 @@
     public RangeInputSocketView(EventBus eventBus, InputSocket<List<Number>> socket) {
         super(eventBus, socket);
 
-        final Object[] domain = socket.getSocketHint().getDomain();
-        final List<Number> value = socket.getValue();
+        final Object[] domain = socket.getSocketHint().getDomain().get();
+        final List<Number> value = socket.getValue().get();
 
         checkArgument(domain != null && domain.length == 1 && domain[0] instanceof List,
```
ThomasJClark: `domain` can't be `null`

src/main/java/edu/wpi/grip/ui/pipeline/input/SelectInputSocketView.java
```diff
@@ -27,14 +27,14 @@
     public SelectInputSocketView(EventBus eventBus, InputSocket<T> socket) {
         super(eventBus, socket);
 
-        final Object[] domain = socket.getSocketHint().getDomain();
+        final Object[] domain = socket.getSocketHint().getDomain().get();
         checkNotNull(domain);
```
ThomasJClark: `domain` can't be `null`

src/main/java/edu/wpi/grip/ui/pipeline/input/SliderInputSocketView.java
```diff
@@ -25,7 +25,7 @@
     public SliderInputSocketView(EventBus eventBus, InputSocket<Number> socket) {
         super(eventBus, socket);
 
-        final Object[] domain = socket.getSocketHint().getDomain();
+        final Object[] domain = socket.getSocketHint().getDomain().get();
 
         checkArgument(domain != null
```
ThomasJClark: `domain` can't be `null`

src/main/java/edu/wpi/grip/ui/pipeline/input/NumberSpinnerInputSocketView.java
```diff
@@ -29,7 +29,7 @@ public NumberSpinnerInputSocketView(EventBus eventBus, InputSocket<Number> socke
         super(eventBus, socket);
 
 
-        Object[] domain = socket.getSocketHint().getDomain();
+        Object[] domain = socket.getSocketHint().getDomain().orElseThrow(() -> new IllegalStateException("No domain was supplied"));
```
ThomasJClark: Don't throw an exception if the domain isn't specified - spinners don't need a domain, so this will cause some operations to crash

src/main/java/edu/wpi/grip/ui/pipeline/input/RangeInputSocketView.java
JLLeitschuh: I don't see the problem here? `.get()` will throw a `NoSuchElementException`

src/main/java/edu/wpi/grip/ui/pipeline/input/SelectInputSocketView.java
JLLeitschuh: So remove this line then?

src/main/java/edu/wpi/grip/core/SocketHints.java
```diff
@@ -0,0 +1,177 @@
...
+
+/**
+ * Commonly used {@link SocketHint SocketHints}
+ */
+public final class SocketHints {
+
+    /**
+     * Provides some default domains
+     */
+    public enum Domain {
```
ThomasJClark: This class is not necessary, and is actually IMO harmful.

If you want a domain of positive integers, you can use `(..., 1, Integer.MAX_VALUE);`.  This is more concise than using this enum, and also more consistent - it makes it so you use the same semantics to set a range of 1 to 100 as 1 to `Integer.MAX_VALUE`, which I think would cause less confusion among people writing python scripts

src/main/java/edu/wpi/grip/ui/pipeline/input/SliderInputSocketView.java
JLLeitschuh: Okay.

src/main/java/edu/wpi/grip/core/SocketHints.java
```diff
@@ -0,0 +1,177 @@
...
+        INTEGERS(Integer.MIN_VALUE, Integer.MAX_VALUE),
+        POSITIVE_INTEGERS(0d, Integer.MAX_VALUE),
```
ThomasJClark: 0 isn't positive.

This is another disadvantage of this - someone might use `POSITIVE_INTEGERS` and expect it to start at 1.  Even if it's fixed, I'm sure plenty of people might be confused at why it doesn't start at 0.

src/main/java/edu/wpi/grip/ui/pipeline/input/RangeInputSocketView.java
ThomasJClark: That's exactly the problem.  `domain` will never be `null`.

src/main/java/edu/wpi/grip/ui/pipeline/input/SelectInputSocketView.java
ThomasJClark: yes

src/main/java/edu/wpi/grip/core/Step.java
JLLeitschuh: That's exactly what I had originally but you didn't like it.

ThomasJClark: Link?  I don't remember that.

src/main/java/edu/wpi/grip/core/SocketHints.java
JLLeitschuh: Would UNSIGNED_INTEGERS make more sense?

ThomasJClark: I think `UNSIGNED_INTEGERS` would imply it goes up to 4 billion, which it doesn't.

ThomasJClark:
> Okay, so this is a problem with serialization. Not the socket hints.

If you've found an problem with serialization, can you open an issue and assign it to me?

src/main/java/edu/wpi/grip/core/Step.java
JLLeitschuh: Second outdated diff from the top

JLLeitschuh:
It was a problem with serializing optional values. I fixed it.

src/main/java/edu/wpi/grip/core/SocketHints.java
JLLeitschuh: @PaulaRudy What do you think? I wrote it because I thought that it would make it cleaner and easier to create domains.

src/main/java/edu/wpi/grip/core/Step.java
JLLeitschuh: The Exception Witness only takes Exceptions. I need to have a Exception to pass to it.

src/main/java/edu/wpi/grip/core/SocketHints.java
PaulaRudy: I'm gonna go with Tom on this one. It adds the potential for confusion without giving enough convenience to make it worth it, IMO.

Issue #155, opened 2015-11-18
Objects should not call `eventBus.register(this)` in constructors


JLLeitschuh:
# Problem
Registering an object to the event bus inside of a constructor call can result in a subscriber being called prior to all of the fields in the class being fully constructed. This can lead to null pointer exceptions for fields that are final and should have been initialized in the classes constructor.

[Relavent Gitter Discussion](https://gitter.im/WPIRoboticsProjects/GRIP?at=564baf1609446b3b7461702a) Discovery of problem and discussion of how to fix it.

[Guava Google Groups Discussion with Dev](https://groups.google.com/forum/#!topic/guava-discuss/4ZJhhpmMvKQ)

> ## Don't publish the "this" reference during construction
One of the mistakes that can introduce a data race into your class is to expose the this reference to another thread before the constructor has completed. Sometimes the reference is explicit, such as directly storing this in a static field or collection, but other times it can be implicit, such as when you publish a reference to an instance of a non-static inner class in a constructor. Constructors are not ordinary methods -- they have special semantics for initialization safety. An object is assumed to be in a predictable, consistent state after the constructor has completed, and publishing a reference to an incompletely constructed object is dangerous. Listing 2 shows an example of introducing this sort of race condition into a constructor. It may look harmless, but it contains the seeds of serious concurrency problems.
> \-[Brian Goetz](http://www.ibm.com/developerworks/library/j-jtp0618/)

# Resolution
This can be resolved by ensuring that constructors never call a register method

JLLeitschuh:
Resolution, this will be handled with dependency injection.

ThomasJClark:
Now that our exception dialog at least doesn't stack overflow when there's a problem with the GUI thread, here's a full stacktrace for the fabled `NullPointerException`.

	java.lang.NullPointerException
		at javafx.scene.Scene$ScenePulseListener.synchronizeSceneNodes(Scene.java:2289)
		at javafx.scene.Scene$ScenePulseListener.pulse(Scene.java:2419)
		at com.sun.javafx.tk.Toolkit.lambda$runPulse$30(Toolkit.java:314)
		at com.sun.javafx.tk.Toolkit$$Lambda$310/747188721.run(Unknown Source)
		at java.security.AccessController.doPrivileged(Native Method)
		at com.sun.javafx.tk.Toolkit.runPulse(Toolkit.java:313)
		at com.sun.javafx.tk.Toolkit.firePulse(Toolkit.java:340)
[...]

PR #179, opened 2015-12-04
Improves unexpected throwable handling

JLLeitschuh:


src/main/java/edu/wpi/grip/core/sources/CameraSource.java
```diff
@@ -173,13 +173,13 @@ private synchronized void startVideo(FrameGrabber grabber) throws IOException {
                     // TODO Pass Exception to the UI.
                     System.err.println("Webcam Frame Grabber Thread crashed with uncaught exception:");
                     exception.printStackTrace();
-                    eventBus.post(new FatalErrorEvent(exception));
+                    eventBus.post(new UnexpectedThrowableEvent(exception));
                     try {
                         stopVideo();
                     } catch (TimeoutException e) {
                         System.err.println("Webcam Frame Grabber could not be stopped!");
                         e.printStackTrace();
```
ThomasJClark: If we're gonna start actually handling errors, don't print the stack trace here.  We'll have something listening for stack traces to log

src/main/java/edu/wpi/grip/ui/Main.java
```diff
@@ -43,18 +42,22 @@ public void start(Stage stage) {
         stage.show();
     }
 
+    private void triggerUnexpectedThrowableEvent(UnexpectedThrowableEvent event) {
```
ThomasJClark: Why is this wrapped in a method?

src/main/java/edu/wpi/grip/ui/Main.java
```diff
@@ -62,5 +65,11 @@ public final void onFatalErrorEvent(FatalErrorEvent error) {
                 }
             }
         });
+
+        if(event.isFatal()) {
+            System.err.println("Original fatal exception");
+            event.getThrowable().printStackTrace();
```
ThomasJClark: We already print the stack trace [here](https://github.com/WPIRoboticsProjects/GRIP/pull/179/files#diff-d5c3d322c479e7a4fc933fa793c49bcdR52), right?

src/main/java/edu/wpi/grip/ui/pipeline/AddSourceView.java
```diff
@@ -52,7 +52,7 @@ public AddSourceView(EventBus eventBus) {
                     eventBus.post(new SourceAddedEvent(new ImageFileSource(eventBus, file)));
                 } catch (IOException e) {
                     e.printStackTrace();
```
ThomasJClark: Remove this `printStackTrace()`s, since we have a handler that prints them

src/main/java/edu/wpi/grip/core/events/UnexpectedThrowableEvent.java
```diff
@@ -0,0 +1,37 @@
...
+public final class UnexpectedThrowableEvent {
+    private final Throwable throwable;
+    private final boolean fatal;
+
+    /**
+     * @param throwable The throwable that was caught.
+     * @param fatal     True if this cause the application to quit forcibly.
+     *                  If the throwable is an {@link Error} than this is automatically true. Defaults to false.
+     */
+    public UnexpectedThrowableEvent(Throwable throwable, boolean fatal) {
```
ThomasJClark: I'm assuming `fatal` is true if we send this from a `Thread.UncaughtExceptionHandler`?  In that case, you should add handling to any threads we create

src/main/java/edu/wpi/grip/core/events/UnexpectedThrowableEvent.java
```diff
@@ -0,0 +1,37 @@
+package edu.wpi.grip.core.events;
+
+/**
+ * Event should be thrown when Unexpected Throwable ends up in an {@link Thread#uncaughtExceptionHandler}.
```
ThomasJClark: Based on your usage of this event, this this isn't true

src/main/java/edu/wpi/grip/ui/Main.java
```diff
@@ -31,8 +31,7 @@ public void start(Stage stage) {
          * Any exceptions thrown by the UI will be caught here and an exception dialog will be displayed
          */
         Thread.currentThread().setUncaughtExceptionHandler((thread, throwable) -> {
-            this.eventBus.post(new FatalErrorEvent(throwable));
-
+            this.eventBus.post(new UnexpectedThrowableEvent(throwable));
```
ThomasJClark: This should be fatal

src/main/java/edu/wpi/grip/core/events/UnexpectedThrowableEvent.java
JLLeitschuh: I plan to. In a different PR. I'm trying to go for small pull requests that are easy to review and merge.

src/main/java/edu/wpi/grip/core/events/UnexpectedThrowableEvent.java
JLLeitschuh: I was going to fix that in a separate PR

src/main/java/edu/wpi/grip/ui/Main.java
JLLeitschuh: Why?
JavaFX hits null pointers all the time. I'm working on fixing that. Some of these things can be easily resolved by simply closing and reopening a preview for example.

src/main/java/edu/wpi/grip/ui/pipeline/AddSourceView.java
JLLeitschuh: Not what this PR is about.

src/main/java/edu/wpi/grip/ui/Main.java
JLLeitschuh: I want to make it entirely clear what the original stack trace is just in case more come later.

src/main/java/edu/wpi/grip/core/sources/CameraSource.java
JLLeitschuh: Events should not be used to log. Logging is a separate thing that adds overhead.

src/main/java/edu/wpi/grip/core/events/UnexpectedThrowableEvent.java
ThomasJClark: Fix the comment or change it so the comment is true?

src/main/java/edu/wpi/grip/ui/pipeline/AddSourceView.java
ThomasJClark: I would argue that fixing the double or triple stack trace printed would definitely "improve unexpected throwable handling"

src/main/java/edu/wpi/grip/ui/Main.java
ThomasJClark: I've never seen this.  Can you open an issue for it?

src/main/java/edu/wpi/grip/ui/Main.java
ThomasJClark: What?  This just makes it print the same stack trace twice

src/main/java/edu/wpi/grip/core/sources/CameraSource.java
ThomasJClark: :disappointed: 

https://github.com/WPIRoboticsProjects/GRIP/pull/179/files#diff-d5c3d322c479e7a4fc933fa793c49bcdR52

JLLeitschuh: Logging things that are expected parts of the application. We should definitely use events for reporting errors to the UI

src/main/java/edu/wpi/grip/ui/Main.java
JLLeitschuh: Because I can't access the `eventBus` from within a lambda being created inside of a constructor for a field that hasn't been assigned yet.

src/main/java/edu/wpi/grip/ui/Main.java
JLLeitschuh: If your console doesn't have infinite scroll on and another exception is thrown from anywhere inside of `PlatformImpl.runAndWait(...` (IE sun's code not our code) then it will be lost.

src/main/java/edu/wpi/grip/ui/pipeline/AddSourceView.java
JLLeitschuh: Fine.

src/main/java/edu/wpi/grip/core/events/UnexpectedThrowableEvent.java
JLLeitschuh: Nope, as soon as I'm done with this I'm going to be replacing its invalid usages with something else. In the meantime it stays where it is to handle exceptions we get.

src/main/java/edu/wpi/grip/ui/Main.java
JLLeitschuh: Its open: #155

src/main/java/edu/wpi/grip/core/events/UnexpectedThrowableEvent.java
ThomasJClark: Nope as in neither?

src/main/java/edu/wpi/grip/core/sources/CameraSource.java
ThomasJClark: I'm having difficulty interpreting this comment.  Could you clarify/rephrase that?

ThomasJClark: - You're logging an exception then sending the exception in an event
- The event handler also logs the exception
- You said "Events should not be used to log", but that's what you're doing

src/main/java/edu/wpi/grip/core/events/UnexpectedThrowableEvent.java
JLLeitschuh: Yes. I'm going to remove it. Just not right now. That is coming in another PR.

src/main/java/edu/wpi/grip/core/sources/CameraSource.java
JLLeitschuh: I'll remove this stack trace.
In the future however, logging should be done in place where the exception occurs.

ui/src/main/java/edu/wpi/grip/ui/ExceptionAlert.java
```diff
@@ -62,17 +64,20 @@
      *                  the issue website.
      * @see <a href="http://code.makery.ch/blog/javafx-dialogs-official/">Inspiration</a>
      */
-    public ExceptionAlert(final Parent root, final Throwable throwable, final HostServices services) {
+    public ExceptionAlert(final Parent root, final Throwable throwable, Optional<String> message, boolean isFatal, final HostServices services) {
```
ThomasJClark: `Optional<String> message` should just be an optional parameter

ui/src/main/java/edu/wpi/grip/ui/ExceptionAlert.java
```diff
@@ -203,10 +199,9 @@ private String generateSystemInfoMessage() {
      * @return The fully constructed issue text.
      */
     private String issueText() {
-        return new StringBuilder(ISSUE_PROMPT_QUESTION)
-                .append("\n\n\n\n")
-                .append(systemInfoMessage)
-                .append(exceptionMessage).toString();
+        final StringBuilder builder = new StringBuilder(ISSUE_PROMPT_QUESTION).append("\n\n\n\n")
```
ThomasJClark: Use string concatenation

ThomasJClark: is this event used?

ui/src/main/java/edu/wpi/grip/ui/ExceptionAlert.java
```diff
@@ -203,10 +199,12 @@ private String generateSystemInfoMessage() {
      * @return The fully constructed issue text.
      */
     private String issueText() {
-        return new StringBuilder(ISSUE_PROMPT_QUESTION)
-                .append("\n\n\n\n")
-                .append(systemInfoMessage)
-                .append(exceptionMessage).toString();
+        return ISSUE_PROMPT_QUESTION
+                + "\n\n\n\n"
+                + getAdditionalInfoMessage()
```
JLLeitschuh: Undo this change....

PR #180, opened 2015-12-04
Adds intellij settings for ensuring same runConfig

JLLeitschuh:
Basically ensures that everyone is using `-ea` param when running grip.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ gradle-app.setting
 
 ## Directory-based project format:
 .idea/
+!.idea/runConfigurations/
 # if you remove the above rule, at least ignore the following:
 
 # User-specific stuff:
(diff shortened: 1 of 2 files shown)


ThomasJClark:
We should absolutely not add IDE files to the repo.  This will break IntelliJ if we every change the main class package (such as for headless mode or refactoring or whatever), and it also means developers using IntelliJ/Eclipse/Netbeans/command line will have different behavior, which is definitely bad.

As for assertions, isn't there some way to enable them in gradle?  I don't see this as a big issue (especially as long as Travis catchs assertions), but it would be okay to add gradle settings to enable assertions.

JLLeitschuh:
1. Do not just close PR's without a discussion. Its rude and dismissive and its upsetting.
2. This files is designed to be shared amongst teammates.

ThomasJClark:
> I was planning on adding the assertion flag to gradle in a separate PR.

That would be good

PR #181, opened 2015-12-04
Add contour-related operations and examples

ThomasJClark:


src/main/java/edu/wpi/grip/core/operations/composite/FindContourOperation.java
```diff
@@ -0,0 +1,96 @@
...
+import static org.bytedeco.javacpp.opencv_core.*;
+import static org.bytedeco.javacpp.opencv_imgproc.*;
+
+/**
+ * An {@link Operation} that, given a binary image, produces a list of contours of all of the shapes in the image
+ */
+public class FindContourOperation implements Operation {
+
+    public static class Contours {
+        int rows = -1, cols = -1;
```
ThomasJClark: Dear annoying bot,

Before you say anything, yes these are package local.  This is intentional, because they are accessed by other classes in the package (there are several operations that accept or produce contours).

JLLeitschuh: Make them protected and make the class final then.

src/main/java/edu/wpi/grip/core/operations/composite/FilterContoursOperation.java
```diff
@@ -0,0 +1,110 @@
...
+            if (bb.width() < minWidth || bb.width() > maxWidth) continue;
+            if (bb.height() < minHeight || bb.height() > maxHeight) continue;
+
+            outputContours.put(filteredContourCount++, contour);
+        }
+
+        outputContours.resize(filteredContourCount);
+
+        final OutputSocket<Contours> outputSocket = (OutputSocket<Contours>) outputs[0];
+        outputSocket.getValue().rows = inputSocket.getValue().rows();
```
JLLeitschuh: This is not javascript. Please don't do this.

ThomasJClark: What?

src/main/java/edu/wpi/grip/core/operations/composite/FilterContoursOperation.java
```diff
@@ -0,0 +1,110 @@
...
+    private final SocketHint<Number> minPerimeterHint = new SocketHint<>("Min Perimeter", Number.class, 0,
+            SocketHint.View.SPINNER);
+    private final SocketHint<Number> minWidthHint = new SocketHint<>("Min Width", Number.class, 0,
+            SocketHint.View.SPINNER);
+    private final SocketHint<Number> maxWidthHint = new SocketHint<>("Max Width", Number.class, Integer.MAX_VALUE,
+            SocketHint.View.SPINNER);
+    private final SocketHint<Number> minHeightHint = new SocketHint<>("Min Height", Number.class, 0,
+            SocketHint.View.SPINNER);
+    private final SocketHint<Number> maxHeightHint = new SocketHint<>("Max Height", Number.class, Integer.MAX_VALUE,
+            SocketHint.View.SPINNER);
```
JLLeitschuh: #177 will break this. We should wait on that to be merged.

src/main/java/edu/wpi/grip/ui/preview/ContoursSocketPreviewView.java
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * @param eventBus The EventBus used by the application
+     * @param socket   An output socket to preview
+     */
+    public ContoursSocketPreviewView(EventBus eventBus, OutputSocket<Contours> socket) {
+        super(eventBus, socket);
+
+        this.setContent(new VBox(this.imageView, this.infoLabel));
+        this.render();
```
JLLeitschuh: You are exposing the `this` object to another thread before the object is fully constructed.

ThomasJClark: That's not true - it's fully constructed by this line (we spoke about this, remember?)

src/main/java/edu/wpi/grip/core/operations/composite/FindContourOperation.java
```diff
@@ -0,0 +1,96 @@
...
+    @Override
+    @SuppressWarnings("unchecked")
+    public void perform(InputSocket<?>[] inputs, OutputSocket<?>[] outputs, Optional<?> data) {
+        final Mat input = ((InputSocket<Mat>) inputs[0]).getValue();
+        final Mat tmp = ((Optional<Mat>) data).get();
+        final boolean externalOnly = ((InputSocket<Boolean>) inputs[1]).getValue();
+        final OutputSocket<Contours> contoursSocket = (OutputSocket<Contours>) outputs[0];
+        final Contours contours = contoursSocket.getValue();
+
+        if (input.empty()) {
```
JLLeitschuh: Not needed if merging after #177

src/main/java/edu/wpi/grip/core/operations/composite/FilterContoursOperation.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Override
+    public OutputSocket<?>[] createOutputSockets(EventBus eventBus) {
+        return new OutputSocket<?>[]{new OutputSocket<>(eventBus, contoursHint)};
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public void perform(InputSocket<?>[] inputs, OutputSocket<?>[] outputs) {
```
JLLeitschuh: Deprecated

src/main/java/edu/wpi/grip/core/operations/composite/FilterContoursOperation.java
ThomasJClark: Or wait on this to be merged?  This is probably ready (pending review), #177 is still in progress

src/main/java/edu/wpi/grip/core/operations/composite/FilterContoursOperation.java
ThomasJClark: That definitely should not be deprecated, I'll probably fix that in a future PR

src/main/java/edu/wpi/grip/core/operations/composite/FilterContoursOperation.java
JLLeitschuh: What's left to be finished with #177? It's done.

src/main/java/edu/wpi/grip/ui/preview/ContoursSocketPreviewView.java
JLLeitschuh: >  Synchronization ensures that changes made by one thread become visible to other threads in a predictable manner. On some architectures, in the absence of synchronization, different threads may see memory operations appear to have been executed in a different order than they actually were executed.
> [Brian Goetz](http://www.ibm.com/developerworks/library/j-jtp0618/)


> **What is meant by reordering?**
There are a number of cases in which accesses to program variables (object instance fields, class static fields, and array elements) may appear to execute in a different order than was specified by the program. The compiler is free to take liberties with the ordering of instructions in the name of optimization. Processors may execute instructions out of order under certain circumstances. Data may be moved between registers, processor caches, and main memory in different order than specified by the program.
> \- [Jeremy Manson and Brian Goetz](https://www.cs.umd.edu/~pugh/java/memoryModel/jsr-133-faq.html)

src/main/java/edu/wpi/grip/core/operations/composite/FilterContoursOperation.java
```diff
@@ -0,0 +1,110 @@
...
+import com.google.common.eventbus.EventBus;
+import edu.wpi.grip.core.InputSocket;
+import edu.wpi.grip.core.Operation;
+import edu.wpi.grip.core.OutputSocket;
+import edu.wpi.grip.core.SocketHint;
+
+import java.io.InputStream;
+import java.util.Optional;
+
+import static edu.wpi.grip.core.operations.composite.FindContourOperation.Contours;
```
JLLeitschuh: This should be its own independent class if it's being used this way. Not an inner class.

ThomasJClark: I made it an inner class because it has like two fields and is only used as the type of contour-related `SocketHint`s

src/main/java/edu/wpi/grip/core/operations/composite/FindContourOperation.java
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * An {@link Operation} that, given a binary image, produces a list of contours of all of the shapes in the image
+ */
+public class FindContourOperation implements Operation {
+
+    public final static class Contours {
+        protected int rows = -1, cols = -1;
+        protected MatVector contours = new MatVector();
+
```
JLLeitschuh: Give this class a constructor like any normal object. Let the java memory model clean it up like any normal object.
This isn't a large object.
If you don't want to keep allocating MatVectors then use the operations, createData() method to allocate a single MatVector to use as the output. The rest of these `Contour` objects can be flushed by the jvm.

ThomasJClark: > Give this class a constructor like any normal object.

The constructor would be empty, which is the same as the implicit constructor.

> Let the java memory model clean it up like any normal object.

What am I doing to stop that?

> If you don't want to keep allocating MatVectors then use the operations, createData() method to allocate a single MatVector to use as the output. The rest of these Contour objects can be flushed (ie. deleted) by the jvm.

I don't care about allocating MatVectors.  They don't take up a lot of space.

src/main/java/edu/wpi/grip/core/operations/composite/FilterContoursOperation.java
```diff
@@ -0,0 +1,110 @@
...
+ */
+public class FilterContoursOperation implements Operation {
+
+    private final SocketHint<Contours> contoursHint = new SocketHint<Contours>("Contours", Contours.class, Contours::new);
+    private final SocketHint<Number> minAreaHint = new SocketHint<>("Min Area", Number.class, 0, SocketHint.View.SPINNER);
+    private final SocketHint<Number> minPerimeterHint = new SocketHint<>("Min Perimeter", Number.class, 0,
+            SocketHint.View.SPINNER);
+    private final SocketHint<Number> minWidthHint = new SocketHint<>("Min Width", Number.class, 0,
+            SocketHint.View.SPINNER);
+    private final SocketHint<Number> maxWidthHint = new SocketHint<>("Max Width", Number.class, Integer.MAX_VALUE,
```
JLLeitschuh: What happens when `Min Width` is greater than `Max Width`

ThomasJClark: No contours would be returned.

src/main/java/edu/wpi/grip/core/operations/composite/FilterContoursOperation.java
JLLeitschuh: It's used by two operations and has its own preview view. Other people may want to use it in their python scripts. It should be its own class.

src/main/java/edu/wpi/grip/core/operations/composite/FilterContoursOperation.java
JLLeitschuh: Modify variables without a setter.
You shouldn't even be doing setting here. This should be a constructor call.

src/main/java/edu/wpi/grip/ui/preview/ContoursSocketPreviewView.java
ThomasJClark: Again, we talked about this a while ago.  We're doing this for now because it probably won't actually cause any problems, but in the long term we're going to use dependency injection

src/main/java/edu/wpi/grip/core/operations/composite/FilterContoursOperation.java
ThomasJClark: ok

src/main/java/edu/wpi/grip/core/operations/composite/ContoursReport.java
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * The output of {@link FindContourOperation}.  This stores a list of contours (which is basically a list of points) in
+ * OpenCV objects, as well as the width and height of the image that the contours are from, to give context to the
+ * points.
+ */
+public final class ContoursReport {
+    protected int rows = -1, cols = -1;
+    protected MatVector contours = new MatVector();
+
```
JLLeitschuh: Give this its own constructor.

src/main/java/edu/wpi/grip/core/operations/composite/ContoursReport.java
```diff
@@ -0,0 +1,25 @@
...
+import static org.bytedeco.javacpp.opencv_core.*;
+
+/**
+ * The output of {@link FindContourOperation}.  This stores a list of contours (which is basically a list of points) in
+ * OpenCV objects, as well as the width and height of the image that the contours are from, to give context to the
+ * points.
+ */
+public final class ContoursReport {
+    protected int rows = -1, cols = -1;
+    protected MatVector contours = new MatVector();
```
JLLeitschuh: Private and final

ThomasJClark: :-1: These aren't final, this is a mutable object

src/main/java/edu/wpi/grip/core/operations/composite/FilterContoursOperation.java
JLLeitschuh: This should be a range. That's why we have that type of socket hint view.

ThomasJClark: What would the min and max be?

JLLeitschuh: 0 and MAX_VALUE like you have right there.

src/main/java/edu/wpi/grip/core/operations/composite/ContoursReport.java
JLLeitschuh: Why? It doesn't need to be a mutable object.

JLLeitschuh: Using immutables prevents race time conditions. They should be used as much as possible in a multithreaded application.

src/main/java/edu/wpi/grip/core/operations/composite/FilterContoursOperation.java
ThomasJClark: :-1: You realize `Integer.MAX_VALUE` is like 2 billion right?  It would be impossible to select a reasonable range (which I think would probably be around 1,000-100,000)

src/main/java/edu/wpi/grip/core/operations/composite/ContoursReport.java
JLLeitschuh: > The real problem is that programmers have spent far too much time worrying about efficiency in the wrong places and at the wrong times; premature optimization is the root of all evil (or at least most of it) in programming.
> \-Donald Knuth

This is premature optimization.

ThomasJClark: That quote has exactly no relevance to this change, but I can update it to be immutable

src/main/java/edu/wpi/grip/core/operations/composite/FilterContoursOperation.java
JLLeitschuh: Max value should be dynamically set from the maximum possible size of a contour. You can get that from the size of the original matrix right?

src/main/java/edu/wpi/grip/core/operations/composite/ConvexHullsOperation.java
```diff
@@ -0,0 +1,65 @@
...
+    private final SocketHint<ContoursReport> contoursHint = new SocketHint<ContoursReport>("Contours", ContoursReport.class, ContoursReport::new);
+
+    @Override
+    public String getName() {
+        return "Convex Hulls";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Compute the convex hulls of contours";
```
JLLeitschuh: Add a period to the end.

src/main/java/edu/wpi/grip/ui/MainWindowView.java
```diff
@@ -76,6 +76,9 @@ public MainWindowView(final EventBus eventBus) {
         eventBus.post(new OperationAddedEvent(new RGBThresholdOperation()));
         eventBus.post(new OperationAddedEvent(new HSVThresholdOperation()));
         eventBus.post(new OperationAddedEvent(new HSLThresholdOperation()));
+        eventBus.post(new OperationAddedEvent(new FindContourOperation()));
+        eventBus.post(new OperationAddedEvent(new ConvexHullsOperation()));
+        eventBus.post(new OperationAddedEvent(new FilterContoursOperation()));
```
JLLeitschuh: This is going to end up being moved to another file in #172

src/main/java/edu/wpi/grip/ui/preview/ContoursSocketPreviewView.java
```diff
@@ -0,0 +1,85 @@
...
+import static org.bytedeco.javacpp.opencv_core.*;
+import static org.bytedeco.javacpp.opencv_imgproc.drawContours;
+
+import edu.wpi.grip.core.operations.composite.ContoursReport;
+
+/**
+ * A preview view for displaying contours.  This view shows each contour as a different-colored outline (so they can be
+ * individually distinguished), as well as a count of the total number of contours found.
+ */
+public class ContoursSocketPreviewView extends SocketPreviewView<ContoursReport> {
```
JLLeitschuh: Finalize this object if you don't think it will be extended.

src/main/java/edu/wpi/grip/ui/preview/ContoursSocketPreviewView.java
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * @param eventBus The EventBus used by the application
+     * @param socket   An output socket to preview
+     */
+    public ContoursSocketPreviewView(EventBus eventBus, OutputSocket<ContoursReport> socket) {
+        super(eventBus, socket);
+
+        this.setContent(new VBox(this.imageView, this.infoLabel));
+        this.render();
```
JLLeitschuh: > Synchronization ensures that changes made by one thread become visible to other threads in a predictable manner. On some architectures, in the absence of synchronization, different threads may see memory operations appear to have been executed in a different order than they actually were executed.
\- Brian Goetz

Another:

> **What is meant by reordering?**
There are a number of cases in which accesses to program variables (object instance fields, class static fields, and array elements) may appear to execute in a different order than was specified by the program. The compiler is free to take liberties with the ordering of instructions in the name of optimization. Processors may execute instructions out of order under certain circumstances. Data may be moved between registers, processor caches, and main memory in different order than specified by the program.
\- Jeremy Manson and Brian Goetz

Also, just so we're clear, I'm talking about calling `Platform.runLater()` in a constructor.

src/main/java/edu/wpi/grip/core/operations/composite/FilterContoursOperation.java
ThomasJClark: We can't set the max value dynamically

src/main/java/edu/wpi/grip/core/operations/composite/ContoursReport.java
ThomasJClark: > Give this its own constructor.

I looked into this.  I seemed like a good idea to make this an immutable value-based type, but it turned out to make everything else (in particular operations that produce a contour report) way more complex.  Plus, we don't currently do that for any other `...Report` classes, so it would be confusing and inconsistent to do that.

src/main/java/edu/wpi/grip/ui/MainWindowView.java
ThomasJClark: ok

src/main/java/edu/wpi/grip/ui/preview/ContoursSocketPreviewView.java
ThomasJClark: ok

src/main/java/edu/wpi/grip/core/operations/composite/ContoursReport.java
JLLeitschuh: This class is currently impossible to use in a python script because of the variable scoping.

ThomasJClark: Because of what?

src/main/java/edu/wpi/grip/core/operations/composite/FilterContoursOperation.java
JLLeitschuh: Right, that's unfortunate.

src/main/java/edu/wpi/grip/core/operations/composite/ContoursReport.java
JLLeitschuh: How does it make it more complex?

src/main/java/edu/wpi/grip/core/operations/composite/FilterContoursOperation.java
ThomasJClark: I disagree, that would be super confusing if the slider moved when you connected a different-sized image to it

src/main/java/edu/wpi/grip/core/operations/composite/ContoursReport.java
JLLeitschuh: If the python script isn't declared in the same package as this file then the `rows` and `cols` values can't be assigned.

src/main/java/edu/wpi/grip/core/operations/composite/ContoursReport.java
ThomasJClark: It's just more lines of meaningless boilerplate code everywhere that we use `ContoursReport`

src/main/java/edu/wpi/grip/core/operations/composite/ContoursReport.java
ThomasJClark: Ok.  I'm not completely clear on what you mean by variable scoping, but the problem with that is that Jython sees that we have getters but not setters and assumes that `rows` and `cols` aren't supposed to be modified.

Since they *are* supposed to be modified, I'll just add setters.  It's also probably better to do that just in case we decide to move all of the report classes to a separate package.

JLLeitschuh: I think the report classes should be moved to a separate package.

ThomasJClark: ok, fixed

ThomasJClark: > I think the report classes should be moved to a separate package.

Yeah, I kind of agree we should overall revamp the package structure a little.  `operations` should probably not be in `core`, `Source` should be in the `sources` package, etc...

src/main/java/edu/wpi/grip/core/operations/composite/ContoursReport.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public MatVector getContours() {
+        return this.contours;
+    }
+
+    public void setRows(int rows) {
+        this.rows = rows;
+    }
+
+    public void setCols(int cols) {
```
JLLeitschuh: These setters are no longer necessary as this object should not be modified after being created. It really should be immutable.

src/main/java/edu/wpi/grip/core/operations/composite/FilterContoursOperation.java
JLLeitschuh: No, the slider would dynamically resize because of the image you attached. The slider wouldn't move.

src/main/java/edu/wpi/grip/core/operations/composite/ConvexHullsOperation.java
```diff
@@ -0,0 +1,65 @@
...
+        final ContoursReport inputContours = inputSocket.getValue();
+        final ContoursReport outputContours = outputSocket.getValue();
+        outputContours.getContours().resize(inputContours.getContours().size());
+
+        for (int i = 0; i < inputContours.getContours().size(); i++) {
+            convexHull(inputContours.getContours().get(i), outputContours.getContours().get(i));
+        }
+
+        outputContours.setRows(inputContours.getRows());
+        outputContours.setCols(inputContours.getCols());
```
JLLeitschuh: This could be done with a copy constructor.

JLLeitschuh:
Why does find contours not support all of the `cv::RetrievalModes` & `cv::ContourApproximationModes`?

JLLeitschuh:
Find contours needs to come with an epilepsy warning or you need to fix those colors flashing like that.

JLLeitschuh:
Pull request against this pull request: https://github.com/ThomasJClark/GRIP/pull/1

src/main/java/edu/wpi/grip/core/operations/composite/ContoursReport.java
JLLeitschuh: Solved in ThomasJClark#1

src/main/java/edu/wpi/grip/core/operations/composite/ConvexHullsOperation.java
JLLeitschuh: Solved in ThomasJClark#1 please merge.

JLLeitschuh:
Can you make it so that it doesn't flash with so many colors. Or add the ability to make the whole thing white until you check a box or something.
I'm not kidding when I say that that operation is epilepsy problem. I had a friend with a concussion looking at my screen when I had that up and was in physical pain after viewing it.

Its a danger. Please resolve.

JLLeitschuh:
Please add the label when this is up to date and you actually want me to look at it again.

core/src/main/java/edu/wpi/grip/core/operations/composite/ContoursReport.java
```diff
@@ -0,0 +1,47 @@
...
+        return this.contours;
+    }
+
+    public void setRows(int rows) {
+        this.rows = rows;
+    }
+
+    public void setCols(int cols) {
+        this.cols = cols;
+    }
```
JLLeitschuh: Its quite easy to make this immutable. You should do that. See ThomasJClark#1

ThomasJClark:
> Why does find contours not support all of the cv::RetrievalModes & cv::ContourApproximationModes?

Because that's a low-level detail, and this is meant to be a high-level operation.  The other retrieval modes are also not useful for this application.

core/src/main/java/edu/wpi/grip/core/operations/composite/ContoursReport.java
ThomasJClark: Yeah, but I'm not gonna do that.  Trust me, I've tried doing it immutably, and I really prefer this way.

ThomasJClark:
> Can you make it so that it doesn't flash with so many colors. Or add the ability to make the whole thing white until you check a box or something.
I'm not kidding when I say that that operation is epilepsy problem. I had a friend with a concussion looking at my screen when I had that up and was in physical pain after viewing it.

> Its a danger. Please resolve.

No, I have no expertise in verifying epilepsy-safe visuals, and I don't support making the UI harder to use to get around that.  Including a warning in the release notes/download page might be a good idea though.

core/src/main/java/edu/wpi/grip/core/operations/composite/FindContoursOperation.java
```diff
@@ -0,0 +1,81 @@
...
+        final ContoursReport contours = contoursSocket.getValue().get();
+
+        if (input.empty()) {
+            return;
+        }
+
+        // findContours modifies its input, so we pass it  a temporary copy of the input image
+        input.copyTo(tmp);
+        findContours(tmp, contours.getContours(), externalOnly ? CV_RETR_EXTERNAL : CV_RETR_LIST,
+                CV_CHAIN_APPROX_TC89_KCOS);
```
JLLeitschuh: Can you explain why you only chose to support these options? What do the others do and why aren't they appropriate for GRIP?

ui/src/main/java/edu/wpi/grip/ui/preview/ContoursSocketPreviewView.java
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * @param eventBus The EventBus used by the application
+     * @param socket   An output socket to preview
+     */
+    public ContoursSocketPreviewView(EventBus eventBus, OutputSocket<ContoursReport> socket) {
+        super(eventBus, socket);
+
+        this.setContent(new VBox(this.imageView, this.infoLabel));
+        this.render();
```
JLLeitschuh: Calling render in the constructor exposes fields in the object to other threads during construction.
Please don't do that.

> Final fields are only frozen when the constructor is finished, and aggressive reordering within the constructor is allowed as long as very simple readability rules are followed.
 \- [Ben Manes](https://groups.google.com/d/msg/guava-discuss/4ZJhhpmMvKQ/LXomj57WAgAJ)

core/src/main/java/edu/wpi/grip/core/operations/composite/FindContoursOperation.java
ThomasJClark: Sure.   First of all, keep in mind that this operation is meant to cover like 95% of use cases, and probably all FRC use cases, but that doesn't mean it has to expose every single option.

- `CV_CHAIN_APPROX_TC89_KCOS` works well and I don't think the average user would care to try to discover the difference.  This is in contrast to something like `BlurOperation`, where each method looks visually different.
- `CV_RETR_EXTERNAL` and `CV_RETR_LIST` return flat lists of contours, which is great because we can process them through a pipeline.  `CV_RETR_TREE` and `CV_RETR_CCOMP` are similar, but they establish a hierarchy of contours, which we can't do anything with.  The only situation where you would use them would be in a Python script anyways.

As usually, if people demand these features and we find a good way to include them, we can add them later.

ui/src/main/java/edu/wpi/grip/ui/preview/ContoursSocketPreviewView.java
ThomasJClark: I know, we already talked about how this will be fixed

ui/src/main/java/edu/wpi/grip/ui/preview/ContoursSocketPreviewView.java
```diff
@@ -0,0 +1,84 @@
...
+            if (!contours.getContours().isNull() && contours.getRows() > 0 && contours.getCols() > 0) {
+                // Allocate a completely black OpenCV Mat to draw the contours onto.  We can easily render contours
+                // by using OpenCV's drawContours function and converting the Mat into a JavaFX Image.
+                this.tmp.create(contours.getRows(), contours.getCols(), CV_8UC3);
+                bitwise_xor(tmp, tmp, tmp);
+
+                numContours = contours.getContours().size();
+
+                for (int i = 0; i < numContours; i++) {
+                    drawContours(this.tmp, contours.getContours(), i, CONTOUR_COLORS[i % CONTOUR_COLORS.length]);
```
JLLeitschuh: Make the color a scale of the contours size.
This just makes it flash and could be epilepsy inducing.

core/src/main/java/edu/wpi/grip/core/operations/composite/ContoursReport.java
JLLeitschuh: What is wrong with how I handled it? Enlighten me to the problems you faced.

core/src/main/java/edu/wpi/grip/core/operations/composite/FindContoursOperation.java
JLLeitschuh: Wonderful! That's really helpful and constructive. It would be great if you left a comment explaining that in the code.

ui/src/main/java/edu/wpi/grip/ui/preview/ContoursSocketPreviewView.java
JLLeitschuh: How does dependency injection solve this problem? By totally removing the constructor?

JLLeitschuh:
> I have no expertise in verifying epilepsy-safe visuals, and I don't support making the UI harder to use to get around that. Including a warning in the release notes/download page might be a good idea though.

I do though. I spent my entire summer reading though documentation about accessibility and issues like this. This is not safe. http://webaim.org/articles/seizure/

ui/src/main/java/edu/wpi/grip/ui/preview/ContoursSocketPreviewView.java
ThomasJClark: With automatic dependency injection, the framework constructs an object and initializes any injected fields.  We don't have any access to the object until after it's fully initialized, so this problem goes away.

ThomasJClark:
Ok.  Regarding epilepsy safety, I would shelf that for not.  It's something you can perhaps work on across-the-board later, and can be solved now with a simple warning.

ui/src/main/java/edu/wpi/grip/ui/preview/ContoursSocketPreviewView.java
JLLeitschuh: Regardless, this would be more useful if the color assigned was relative to the size of the contour it was drawing. Why don't you do that.

ThomasJClark:
Can you stop removing the ready for review label?

JLLeitschuh:
I'm removing it and adding the `waiting response` after I'm done reviewing to let you know I'm waiting for a response and done responding.

ThomasJClark:
Ok.  Can you just leave that label up?  Since it's in review

ui/src/main/java/edu/wpi/grip/ui/preview/ContoursSocketPreviewView.java
JLLeitschuh: Fine, :ok:

ui/src/main/java/edu/wpi/grip/ui/preview/ContoursSocketPreviewView.java
ThomasJClark: I don't understand why that would be useful

JLLeitschuh: Because currently the contours flash arbitrarily. The colors should actually represent something.
This just looks bad and, again, epilepsy.

JLLeitschuh:
I am actually waiting on a response. Can you only remove the `waiting on response` label once you've actually responded?

ThomasJClark:
A response on what?  fyi I'm not going to do anything about the epilepsy comments.

JLLeitschuh:
> Epilepsy Facts.
1 in 26 Americans will develop epilepsy in their lifetime. An estimated 3 million Americans and 65 million people worldwide currently live with epilepsy. Each year at least 200,000 people are diagnosed with epilepsy.

Do you want to make this project  unusable to 1/26 people?

Waiting on:
- https://github.com/WPIRoboticsProjects/GRIP/pull/181#discussion-diff-46750279R75
- https://github.com/WPIRoboticsProjects/GRIP/pull/181#discussion-diff-46750195R38

core/src/main/java/edu/wpi/grip/core/operations/composite/FindContoursOperation.java
ThomasJClark: sure

ThomasJClark:
ok, done

JLLeitschuh:
Oh, and waiting on this: https://github.com/WPIRoboticsProjects/GRIP/pull/181#discussion-diff-46750543R70

ThomasJClark:
> Oh, and waiting on this: #181 (diff)

I already responded about that one.  I'm not going to change it.

ui/src/main/java/edu/wpi/grip/ui/preview/ContoursSocketPreviewView.java
JLLeitschuh: Do you not think this will be confusing?

ThomasJClark: No, I don't.

OpenCV returns contorus in order from top left to bottom right, and we color them according to the order that OpenCV returns them.  If you give it a really noise stream of images, the contours will change a lot, and it'll be hard to track them, and you'll be able to tell that.  If we color by size, you wouldn't be able to distinguish the individual contours, and it would be difficult for users to distinguish how well their pipeline is working and what needs to change in it.

As as simple example, this:

a

... is more informative than this...

b

Although in practice, coloring all contours with a size range the same will probably have this effect on noisy images:

c

JLLeitschuh: Fine, then perhaps they shouldn't be colored if there are more than like 20 contours

ThomasJClark: No!  Making them not colored makes it *harder* to distinguish between them.

JLLeitschuh: If there are more than 50-70 it is impossible to distinguish between them anyways because they are all flashing so much.

PR #186, opened 2015-12-04
Fix TextAreaSocketPreviewView after Optional Refactor

JLLeitschuh:


ui/src/main/java/edu/wpi/grip/ui/preview/TextAreaSocketPreviewView.java
```diff
@@ -35,7 +35,7 @@ public void onSocketChanged(SocketChangedEvent event) {
         final Socket socket = event.getSocket();
 
         if (socket == this.getSocket()) {
-            this.text.setText(socket.getValue().toString());
+            this.text.setText(socket.getValue().get().toString());
```
ThomasJClark: This should better handle empty sockets, even if we think that won't ever happen.

ui/src/main/java/edu/wpi/grip/ui/preview/TextAreaSocketPreviewView.java
```diff
@@ -24,7 +24,7 @@ public TextAreaSocketPreviewView(EventBus eventBus, OutputSocket<T> socket) {
 
         this.setStyle("-fx-pref-width: 20em;");
 
-        this.text = new TextArea(socket.getValue().toString());
+        this.text = new TextArea(socket.getValue().orElse((T)"EMPTY!").toString());
```
JLLeitschuh: Yes, this does work. I tested it.

JLLeitschuh: Use is present.

PR #194, opened 2015-12-07
Add initial headless mode support

ThomasJClark:


build.gradle
```diff
@@ -60,9 +60,18 @@ project(":core") {
...
+            dirs 'libs'
+        }
+    }
+
     dependencies {
         compile group: 'org.bytedeco', name: 'javacv', version: '1.1'
         compile group: 'org.bytedeco.javacpp-presets', name: 'opencv', version: '3.0.0-1.1'
+        compile group: 'org.bytedeco.javacpp-presets', name: 'opencv', version: '3.0.0-1.1', classifier: os
+        compile group: 'org.bytedeco.javacpp-presets', name: 'opencv-3.0.0-1.1', classifier: "linux-frc"
+        compile group: 'org.bytedeco.javacpp-presets', name: 'videoinput', version: '0.200-1.1', classifier: os
```
JLLeitschuh: Add a flag

JLLeitschuh:
```
if (project.hasProperty('coreOnly')) {
  compile group: 'org.bytedeco.javacpp-presets', name: 'videoinput', version: '0.200-1.1', classifier: os
} else {
  testCompile group: 'org.bytedeco.javacpp-presets', name: 'videoinput', version: '0.200-1.1', classifier: os
}
```

core/src/main/java/edu/wpi/grip/core/Main.java
```diff
@@ -1,6 +1,38 @@
...
-    public static void main(String[] args) {
+    public static void main(String[] args) throws Exception {
+        if (args.length != 1) {
+            System.err.println("Usage: GRIP.jar project.grip");
+            return;
+        }
+
+        final String projectPath = args[0];
+
+        final EventBus eventBus = new EventBus();
```
JLLeitschuh: This should at least print the stack trace. The built in logger for the event bus sucks.

core/src/main/java/edu/wpi/grip/core/sources/IPCameraFrameGrabber.java
```diff
@@ -0,0 +1,224 @@
+/*
```
JLLeitschuh: This class isn't even used anywhere?

JLLeitschuh: Also, IPCameraFrameGrabber isn't a source, It should probably go in a different package.

core/src/main/java/edu/wpi/grip/core/Main.java
```diff
@@ -1,6 +1,38 @@
...
+        final Pipeline pipeline = new Pipeline(eventBus);
+        final Palette palette = new Palette(eventBus);
+        final Project project = new Project(eventBus, pipeline, palette);
```
JLLeitschuh: We should have one way of creating these objects and the UI should take advantage of it.
Do something like `main.setup()`.
That way if we change any of the constructors we don't need to worry about it.

As for dependency injection, lets take care of that when it actually gets implemented.

core/src/main/java/edu/wpi/grip/core/sources/IPCameraFrameGrabber.java
JLLeitschuh: Can you add very clear comments on the lines you've added/changed to be completely clear what is different.
Do it inline like `// Our change`

core/src/main/java/edu/wpi/grip/core/sources/IPCameraFrameGrabber.java
```diff
@@ -0,0 +1,224 @@
...
+        BufferedImage bi = ImageIO.read(new ByteArrayInputStream(readImage()));
+        return bi;
+    }
+
+    byte[] readImage() throws IOException {
+        byte[] buffer = new byte[4096];// MTU or JPG Frame Size?
+        int n = -1;
+
+        StringBuffer sb = new StringBuffer();
+        int total = 0;
```
JLLeitschuh: This variable is no longer used.

core/src/main/java/edu/wpi/grip/core/sources/IPCameraFrameGrabber.java
ThomasJClark: > This class isn't even used anywhere?

That's not true

core/src/main/java/edu/wpi/grip/core/sources/IPCameraFrameGrabber.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    public BufferedImage grabBufferedImage() throws IOException {
+        BufferedImage bi = ImageIO.read(new ByteArrayInputStream(readImage()));
+        return bi;
+    }
+
+    byte[] readImage() throws IOException {
+        byte[] buffer = new byte[4096];// MTU or JPG Frame Size?
+        int n = -1;
```
JLLeitschuh: This variable is no longer used.

core/src/main/java/edu/wpi/grip/core/sources/IPCameraFrameGrabber.java
```diff
@@ -0,0 +1,224 @@
...
+        }
+    }
+
+    public BufferedImage grabBufferedImage() throws IOException {
+        BufferedImage bi = ImageIO.read(new ByteArrayInputStream(readImage()));
+        return bi;
+    }
+
+    byte[] readImage() throws IOException {
+        byte[] buffer = new byte[4096];// MTU or JPG Frame Size?
```
JLLeitschuh: This variable is no longer used.

core/src/main/java/edu/wpi/grip/core/Main.java
ThomasJClark: > As for dependency injection, lets take care of that when it actually gets implemented.
 Add a line note

This is the answer to the first part of your comment

build.gradle
```diff
@@ -60,15 +60,22 @@ project(":core") {
...
+            dirs 'libs'
+        }
+    }
+
     dependencies {
         compile group: 'org.bytedeco', name: 'javacv', version: '1.1'
         compile group: 'org.bytedeco.javacpp-presets', name: 'opencv', version: '3.0.0-1.1'
+        compile group: 'org.bytedeco.javacpp-presets', name: 'opencv', version: '3.0.0-1.1', classifier: os
+        compile group: 'org.bytedeco.javacpp-presets', name: 'opencv-3.0.0-1.1', classifier: 'linux-frc'
+        compile group: 'org.bytedeco.javacpp-presets', name: 'videoinput', version: '0.200-1.1', classifier: os
```
JLLeitschuh: Still need to fix this.

ThomasJClark: :-1: this works fine and doesn't have a huge impact on file size

core/src/main/java/edu/wpi/grip/core/sources/IPCameraFrameGrabber.java
JLLeitschuh: I don't see any changes in the CameraSource. How is this one being imported? Doesn't the bytdeco import take precedence?

ThomasJClark: no

ThomasJClark: this one does

core/src/main/java/edu/wpi/grip/core/Main.java
```diff
@@ -1,6 +1,38 @@
...
+        final EventBus eventBus = new EventBus((exception, context) -> exception.printStackTrace());
+        final Pipeline pipeline = new Pipeline(eventBus);
+        final Palette palette = new Palette(eventBus);
+        final Project project = new Project(eventBus, pipeline, palette);
+
+        Operations.addOperations(eventBus);
+        CVOperations.addOperations(eventBus);
+
+        // Open a project from a .grip file specified on the command line
+        project.open(new File(projectPath));
```
JLLeitschuh: Either add dependency injection first or make one way to construct GRIP.
It should probably be wrapped an object like `GripInstance`

PR #197, opened 2015-12-09
Preview order functionality complete

PaulaRudy:
Preview order now reflects position in the pipeline or in the list of sources. Source previews are always displayed at the beginning. Previews now react to steps being moved in the pipeline.


PaulaRudy:
So at this point functionality is complete. I've tested it by hand and everything works. I would like to pull out some methods from the onPreviewOrderChanged method and add some tests, but as of this moment everything works fine.

JLLeitschuh:
Uhhhhh. Yea, you added all of the generated files to the repository.
Can you undo that please?

PaulaRudy:
All fixed! Thanks for catching that.

ui/src/main/java/edu/wpi/grip/ui/pipeline/PipelineView.java
```diff
@@ -154,10 +154,12 @@ private OutputSocketView findOutputSocketView(OutputSocket socket) {
     }
 
     /**
+     * This function is public because it is used by {@link edu.wpi.grip.ui.preview.PreviewsView#onSocketPreviewChanged}
+     * to order the list of previews.
+     *
```
JLLeitschuh: This comment isn't really necessary and, as @ThomasJClark said before, Intellij does this for you by default.

ui/src/main/java/edu/wpi/grip/ui/pipeline/PipelineView.java
```diff
@@ -168,9 +170,12 @@ private SourceView findSourceView(Source source) {
...
```
JLLeitschuh: Same as above.

ui/src/main/java/edu/wpi/grip/ui/preview/PreviewsView.java
```diff
@@ -48,24 +58,196 @@ public PreviewsView(EventBus eventBus) {
...
+                        boolean zeroReached = false;//We will set this to true if we reach the beginning of the list of previews (there are no source previews open)
+                        while ((!zeroReached) &&
+                                ((nextSocketInDirection.getStep().isPresent())
+                                        && (nextSocketInDirection.getStep().get() == this.previewedSockets.get(newLocation).getStep().get()))) { //While we haven't reached the beginning of the list of previews, the socket at this location is a socket from a step, and it is the SAME step as the step of the socket at the new location...
+                            count++;
+                            if ((newLocation - count) > 0)//If we haven't reached the beginning of the list of open previews...
+                                nextSocketInDirection = this.previewedSockets.get(newLocation - count);//Grab the next previewed socket to examine in the direction we are moving
+                            else
+                                zeroReached = true;//Mark that we've reached the beginning of the list of previews so we know to stop looking for more
```
JLLeitschuh: So as a general practice always wrap your if statement code in blocks `{ ... }` OR put it on the same line
`else zeroReached = true;`

I enforce this because merging code can create bugs that are really hard to spot.
See the major apple security bug that was caused by having two gotos after an if statement: https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/

ui/src/main/java/edu/wpi/grip/ui/pipeline/PipelineView.java
ThomasJClark: @JLLeitschuh :+1:

ui/src/main/java/edu/wpi/grip/ui/pipeline/PipelineView.java
ThomasJClark: @JLLeitschuh :+1:

ui/src/main/java/edu/wpi/grip/ui/preview/PreviewsView.java
```diff
@@ -48,24 +58,196 @@ public PreviewsView(EventBus eventBus) {
     }
 
     @Subscribe
+    /**
+     * This function is called when a step moves in the pipeline to adjust the positions of any open previews it has
+     * to reflect the new order of the pipeline.
+     */
+    public synchronized void onPreviewOrderChanged(StepMovedEvent event) {
```
ThomasJClark: So, just curious, but is there a reason why you didn't use some pre-defined sorting function?  Seems like it would be nice to not have all of this manual keeping track of positions and stacks and stuff.

    ObservableList<SocketPreviewView<?>> previews = FXCollections.observableArrayList(
            (Collection) previewBox.getChildren());

    Collections.sort(previews, (a, b) -> {
        if (...) {
            return -1; // a goes before b
        } else {
            return -1; // b goes before a
        }
    });

    previewBox.getChildren().setAll(previews);

ui/src/main/java/edu/wpi/grip/ui/preview/PreviewsView.java
PaulaRudy: Fair enough :)

ui/src/main/java/edu/wpi/grip/ui/pipeline/PipelineView.java
PaulaRudy: Ok. Will fix.

ui/src/main/java/edu/wpi/grip/ui/pipeline/PipelineView.java
PaulaRudy: Ok. Will fix.

ui/src/main/java/edu/wpi/grip/ui/preview/PreviewsView.java
PaulaRudy: To be honest, I'd rather just leave it as is and come back to this later- at this point, logging seems more important to me. Is this something that you think is critical before this can be merged? I can always come back to this later. But I do think you're right about this method being cleaner. I avoided this in this instance because I forgot about lambdas (I learned java without them). At the very least, next time I will be sure to use this method. Thank you for bringing it up :)

ui/src/main/java/edu/wpi/grip/ui/preview/PreviewsView.java
```diff
@@ -28,12 +35,15 @@
 
     private final EventBus eventBus;
     private final List<OutputSocket<?>> previewedSockets;
+    private final PipelineView pipeline;//This is used to determine the order the previews are shown in.
 
-    public PreviewsView(EventBus eventBus) {
+    public PreviewsView(EventBus eventBus, PipelineView pipeline) {
```
JLLeitschuh: :+1: Do this please.

ui/src/main/java/edu/wpi/grip/ui/preview/PreviewsView.java
JLLeitschuh: I'm going to agree with @PaulaRudy on this one. There are more important things to be taken care of.
If we have problems in the future this can be refactored.

ui/src/main/java/edu/wpi/grip/ui/preview/PreviewsView.java
```diff
@@ -48,24 +58,203 @@ public PreviewsView(EventBus eventBus) {
     }
 
     @Subscribe
+    /**
+     * This function is called when a step moves in the pipeline to adjust the positions of any open previews it has
+     * to reflect the new order of the pipeline.
+     */
```
JLLeitschuh: Can you move the comment  to above the `@Subscribe` attribute. Its the code style we've been following everywhere else and here it looks out of place.

ui/src/main/java/edu/wpi/grip/ui/preview/PreviewsView.java
```diff
@@ -48,24 +58,203 @@ public PreviewsView(EventBus eventBus) {
...
+                this.previewedSockets.add(newLocation, current);//...add it to the correct location in the list of previews open
+                this.previewBox.getChildren().add(newLocation, SocketPreviewViewFactory.createPreviewView(this.eventBus, current));//...and display it in the correct location in the list of previews open
+            }
+        });
+    }
+
+    @Subscribe
+    /**
+     * This function is called when a preview button is pushed/triggered
+     */
```
JLLeitschuh: Same as above,
comments above the `@Subscribe`

ui/src/main/java/edu/wpi/grip/ui/preview/PreviewsView.java
```diff
@@ -48,24 +58,203 @@ public PreviewsView(EventBus eventBus) {
     }
 
     @Subscribe
+    /**
+     * This function is called when a step moves in the pipeline to adjust the positions of any open previews it has
+     * to reflect the new order of the pipeline.
+     */
+    public synchronized void onPreviewOrderChanged(StepMovedEvent event) {
```
JLLeitschuh: I do agree with this comment. The complexity is high. If we have problems we may have to refactor it.

ui/src/main/java/edu/wpi/grip/ui/preview/PreviewsView.java
PaulaRudy: Sure!

ui/src/main/java/edu/wpi/grip/ui/preview/PreviewsView.java
PaulaRudy: So this is an issue with the original code, not my changes. Is this still something you would like me to do? Shouldn't it be in a separate pull request?

JLLeitschuh: Just take care of it here. You might as well fix this while you're here.

PR #201, opened 2015-12-10
Fix Divide by Zero Error in CameraSource

### LINKED ISSUE #199: ArithmeticException: / by zero
What were the actions performed prior to this error appearing?

I added a webcam.  I'm using a USB webcam (Logitech 9000 Pro) on Windows 10 64-bit

Message: Camera Frame Grabber Thread crashed with uncaught exception

## System Info:

Property Name | Property 
 ----- | -----
javafx.version | 8.0.66
java.runtime.name | Java(TM) SE Runtime Environment
java.vm.version | 25.66-b17
java.vm.vendor | Oracle Corporation
java.vm.name | Java HotSpot(TM) 64-Bit Server VM
java.runtime.version | 1.8.0_66-b17
java.awt.graphicsenv | sun.awt.Win32GraphicsEnvironment
javafx.runtime.version | 8.0.66-b17
os.name | Windows 10
os.version | 10.0
os.arch | amd64
file.encoding | Cp1252
java.vm.info | mixed mode
java.version | 1.8.0_66
sun.arch.data.model | 64
sun.cpu.endian | little

## Stack Trace:
```java
java.lang.ArithmeticException: / by zero
	at edu.wpi.grip.core.sources.CameraSource.lambda$start$18(CameraSource.java:158)
	at java.lang.Thread.run(Thread.java:745)

```

JLLeitschuh:
Closes #199

### DIFF
--- a/core/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
+++ b/core/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
@@ -137,7 +137,7 @@ public final class CameraSource extends Source implements StartStoppable {
             }
 
             final Thread frameExecutor = new Thread(() -> {
-                long lastFrame = System.currentTimeMillis();
+                long lastFrame = System.nanoTime();
                 while (!Thread.interrupted()) {
                     final Frame videoFrame;
                     try {
@@ -154,8 +154,13 @@ public final class CameraSource extends Source implements StartStoppable {
 
                     frameMat.copyTo(frameOutputSocket.getValue().get());
                     frameOutputSocket.setValue(frameOutputSocket.getValue().get());
-                    long thisMoment = System.currentTimeMillis();
-                    frameRateOutputSocket.setValue(1000 / (thisMoment - lastFrame));
+
+                    /*
+                     *
+                     */
+                    final long thisMoment = System.nanoTime();
+                    final long elapsedTime = thisMoment - lastFrame;
+                    if (elapsedTime != 0) frameRateOutputSocket.setValue(1e9 / elapsedTime);
                     lastFrame = thisMoment;
                 }
             }, "Camera");


core/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
```diff
@@ -154,8 +154,16 @@ public void start() throws IOException, IllegalStateException {
...
-                    long thisMoment = System.currentTimeMillis();
-                    frameRateOutputSocket.setValue(1000 / (thisMoment - lastFrame));
+
+                    final long thisMoment = System.currentTimeMillis();
+                    final long elapsedTime = thisMoment - lastFrame;
+                    if (elapsedTime == 0) {
+                        // This can actually happen: https://github.com/WPIRoboticsProjects/GRIP/issues/199
+                        frameRateOutputSocket.setValue(Double.NaN);
+                    } else {
+                        assert elapsedTime > 0 : "Time has run backwards!";
```
ThomasJClark: You should not use `System.currentTimeMillis()` if going backwards is a problem.  `System.currentTimeMillis()` is non-monotonic, so backwards times are expected.

[`System.nanoTime()` is monotonically increasing](http://docs.oracle.com/javase/7/docs/api/java/lang/System.html#nanoTime()), so it would be a better choice

core/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
```diff
@@ -154,8 +154,16 @@ public void start() throws IOException, IllegalStateException {
...
                     frameMat.copyTo(frameOutputSocket.getValue().get());
                     frameOutputSocket.setValue(frameOutputSocket.getValue().get());
-                    long thisMoment = System.currentTimeMillis();
-                    frameRateOutputSocket.setValue(1000 / (thisMoment - lastFrame));
+
+                    final long thisMoment = System.currentTimeMillis();
+                    final long elapsedTime = thisMoment - lastFrame;
+                    if (elapsedTime == 0) {
+                        // This can actually happen: https://github.com/WPIRoboticsProjects/GRIP/issues/199
+                        frameRateOutputSocket.setValue(Double.NaN);
```
ThomasJClark: I think we should try not to have NaNs in the pipeline to avoid weird unexpected behavior.  A decent way to handle this would probably be to not change `frameRateOutputSocket` if essentially no time has elapsed, or to set it to 0.

These other options aren't technically mathematically correct, but I think it would be less suprising and less likely to break things.

core/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
JLLeitschuh: :+1: Okay.

core/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
JLLeitschuh: Brad offered a good suggestion. If for some reason the frame rate is zero then we should use the old value? Does this make sense?

ThomasJClark: If the frame rate is zero?  Or the elapsed time?

bradamiller: Actually, I suggested that if the frame time is zero, then check the source and see why that is the case and how it can happen.

-- 
Brad Miller


On December 10, 2015 at 2:01:03 PM, Thomas Clark (notifications@github.com) wrote:

In core/src/main/java/edu/wpi/grip/core/sources/CameraSource.java:

> @@ -154,8 +154,16 @@ public void start() throws IOException, IllegalStateException {
>   
>                      frameMat.copyTo(frameOutputSocket.getValue().get());
>                      frameOutputSocket.setValue(frameOutputSocket.getValue().get());
> -                    long thisMoment = System.currentTimeMillis();
> -                    frameRateOutputSocket.setValue(1000 / (thisMoment - lastFrame));
> +
> +                    final long thisMoment = System.currentTimeMillis();
> +                    final long elapsedTime = thisMoment - lastFrame;
> +                    if (elapsedTime == 0) {
> +                        // This can actually happen: https://github.com/WPIRoboticsProjects/GRIP/issues/199
> +                        frameRateOutputSocket.setValue(Double.NaN);
If the frame rate is zero? Or the elapsed time?

core/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
```diff
@@ -154,8 +154,21 @@ public void start() throws IOException, IllegalStateException {
...
+                    if (elapsedTime == 0) {
+                        /* This can happen when using system.currentTimeMillis():
+                         * https://github.com/WPIRoboticsProjects/GRIP/issues/199
+                         * As such we use system.nanoTime().
+                         * However, we add an assertion to ensure this bug doesn't reappear.
+                         */
+                        assert elapsedTime != 0 : "The time between grabbing two frames was less than 1 nanosecond";
+                        frameRateOutputSocket.setValue(frameRateOutputSocket.getValue().get());
+                    } else {
+                        assert elapsedTime > 0 : "Time has run backwards!";
```
ThomasJClark: Time does not run backwards with `System.nanoTime()` (and even if it did, an assertion would only catch that during development)

core/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
```diff
@@ -154,8 +154,21 @@ public void start() throws IOException, IllegalStateException {
...
+
+                    final long thisMoment = System.nanoTime();
+                    final long elapsedTime = thisMoment - lastFrame;
+                    if (elapsedTime == 0) {
+                        /* This can happen when using system.currentTimeMillis():
+                         * https://github.com/WPIRoboticsProjects/GRIP/issues/199
+                         * As such we use system.nanoTime().
+                         * However, we add an assertion to ensure this bug doesn't reappear.
+                         */
+                        assert elapsedTime != 0 : "The time between grabbing two frames was less than 1 nanosecond";
```
ThomasJClark: Why couldn't it occur with `System.nanoTime()`?  Monotonically increasing doesn't mean every call is greater than every previous call, it just never goes backwards.

This is also not an appropriate place to use an assertion, since we'd only see it during development.  So, if we're asserting something, it's a test to make sure it'll be like that before we ship with.  Since the thing you're asserting is a timing condition, passing one time does not mean it will always pass.

JLLeitschuh: What do you want me to do instead then? Leave it as a divide by zero error but using nano time instead? What do you think should be here?

core/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
JLLeitschuh: I guess it doesn't make sense to test for that.

core/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
ThomasJClark: if (elapsedTime != 0) frameRateOutputSocket.setValue(1e12 / elapsedTime);

ThomasJClark: And don't even check if time has run backwards

core/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
ThomasJClark: Right, but assertions are good because they'll tell us if we break something before we deploy.  `elapsedTime != 0` will tell us if an unlikely timing condition occurs, and since that would not happen consistently, the assertion would probably not catch it before deploying.

You're right that `assert elapsedTime > 0` should never happen, but I don't think that's even worth asserting.

PaulaRudy: I agree with Tom on this, but if it's already there I don't see any harm in keeping it.

core/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
PaulaRudy: So I looked up using assertions in production code. Apparently this is an ongoing debate, but most people seem to agree with Tom that it should be avoided- instead of an assert, you should be handling errors explicitly- IE If you want to keep the test in, you should have something to handle it when it fails and you should be avoiding using "assert" in general. If you'd like to test this condition, re-writing it so that you don't use assert and the error is handled would be OK with me. But I think the easier option would be to simply take it out....

JLLeitschuh: I'll take it out

core/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
```diff
@@ -154,8 +154,13 @@ public void start() throws IOException, IllegalStateException {
 
                     frameMat.copyTo(frameOutputSocket.getValue().get());
                     frameOutputSocket.setValue(frameOutputSocket.getValue().get());
-                    long thisMoment = System.currentTimeMillis();
-                    frameRateOutputSocket.setValue(1000 / (thisMoment - lastFrame));
+
+                    /*
+                     *
```
ThomasJClark: Probably get rid of the empty comment things

PR #204, opened 2015-12-11
Feat: Remote Deployment and Running

JLLeitschuh:
Adds the ability to deploy and run GRIP on a remote platform.
There is currently no GUI for it at the moment

Deployment Menu
Not saved yet dialog
Simple Deploy
Advanced Deploy
Durring Deploy

### DIFF
--- a/core/src/main/java/edu/wpi/grip/core/Main.java
+++ b/core/src/main/java/edu/wpi/grip/core/Main.java
@@ -26,12 +26,12 @@ public class Main {
     @Inject
     private Logger logger;
 
-    @SuppressWarnings("PMD.SignatureDeclareThrowsException")
-    public static void main(String[] args) throws Exception {
+    public static void main(String[] args) throws IOException, InterruptedException {
         final Injector injector = Guice.createInjector(new GRIPCoreModule());
         injector.getInstance(Main.class).start(args);
     }
 
+    @SuppressWarnings("PMD.SystemPrintln")
     public void start(String[] args) throws IOException, InterruptedException {
         if (args.length != 1) {
             System.err.println("Usage: GRIP.jar project.grip");
@@ -69,6 +69,9 @@ public class Main {
         // Open a project from a .grip file specified on the command line
         project.open(new File(projectPath));
 
+
+        // This is done in order to indicate to the user using the deployment UI that this is running
+        System.out.println("SUCCESS! The project is running in headless mode!");
         // There's nothing more to do in the main thread since we're in headless mode - sleep forever
         for (; ; ) {
             Thread.sleep(Integer.MAX_VALUE);
--- a/ui/src/main/java/edu/wpi/grip/ui/DeployerController.java
+++ b/ui/src/main/java/edu/wpi/grip/ui/DeployerController.java
@@ -1,0 +1,129 @@
+package edu.wpi.grip.ui;
+
+
+import com.google.common.base.Throwables;
+import com.google.common.eventbus.EventBus;
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+import edu.wpi.grip.ui.annotations.ParametrizedController;
+import edu.wpi.grip.ui.components.StartStoppableButton;
+import edu.wpi.grip.ui.deployment.DeploymentOptionsController;
+import edu.wpi.grip.ui.deployment.DeploymentOptionsControllersFactory;
+import edu.wpi.grip.ui.util.deployment.DeployedInstanceManager;
+import javafx.application.Platform;
+import javafx.fxml.FXML;
+import javafx.scene.control.Accordion;
+import javafx.scene.control.DialogPane;
+import javafx.scene.control.ProgressBar;
+import javafx.scene.control.TextArea;
+import javafx.scene.layout.HBox;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.PrintStream;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+@Singleton
+@ParametrizedController(url = "DeployerPane.fxml")
+public class DeployerController {
+
+    @FXML
+    private DialogPane root;
+
+    @FXML
+    private HBox controlsBox;
+
+    @FXML
+    private Accordion deploymentMethods;
+
+    @FXML
+    private TextArea stdOutStreamTextArea;
+
+    @FXML
+    private TextArea stdErrStreamTextArea;
+
+    @FXML
+    private ProgressBar progressIndicator;
+
+    private final EventBus eventBus;
+    private final StartStoppableButton.Factory startStopButtonFactory;
+    private final DeploymentOptionsControllersFactory optionsControllersFactory;
+
+    private class StreamToTextArea extends OutputStream {
+        private final TextArea outputArea;
+
+        public StreamToTextArea(TextArea outputArea) {
+            this.outputArea = outputArea;
+        }
+
+        public StreamToTextArea reset() {
+            outputArea.clear();
+            return this;
+        }
+
+        @Override
+        public void write(int i) throws IOException {
+            outputArea.appendText(String.valueOf((char) i));
+        }
+    }
+
+
+    public interface Factory {
+        DeployerController create();
+    }
+
+    @Inject
+    DeployerController(EventBus eventBus, StartStoppableButton.Factory startStopButtonFactory, DeploymentOptionsControllersFactory optionsControllersFactory) {
+        this.eventBus = eventBus;
+        this.startStopButtonFactory = startStopButtonFactory;
+        this.optionsControllersFactory = optionsControllersFactory;
+    }
+
+    @FXML
+    private void initialize() {
+        final Supplier<OutputStream> out = () ->
+                new PrintStream(new StreamToTextArea(stdOutStreamTextArea).reset(), false);
+        final Supplier<OutputStream> err = () ->
+                new PrintStream(new StreamToTextArea(stdErrStreamTextArea).reset(), false);
+        deploymentMethods.getPanes().addAll(
+                optionsControllersFactory
+                        .createControllers(this::onDeploy, out, err)
+                        .stream()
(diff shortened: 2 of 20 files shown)


ui/src/main/java/edu/wpi/grip/ui/util/deployment/HeadlessInstanceManager.java
```diff
@@ -0,0 +1,152 @@
...
+        // Ensure that the project isn't running from a previous instance.
+        runStop();
+        Thread launcher = new Thread(()-> {
+            try {
+                final Shell.Plain gripShell = new Shell.Plain(new Shell.Safe(details.createSSHShell()));
+                final String output =
+                        gripShell.exec("nohup " + deploymentCommands.getJARLaunchCommand(coreJar.getName(), projectFile.getName()) + " &");
+                // TODO: Do something with the output?
+            } catch (IOException e) {
+                throw new IllegalStateException("The program failed to start");
```
JLLeitschuh: Good catch.

ui/src/main/java/edu/wpi/grip/ui/util/deployment/SecureShellDetails.java
```diff
@@ -0,0 +1,97 @@
...
+            sshDirCommand += (":" + password.get());
+        }
+        sshDirCommand += ("@" + host + ":");
+        if (remoteDir.isPresent()) {
+            sshDirCommand += remoteDir.get();
+        }
+        return sshDirCommand;
+    }
+
+    public static final class FRCSecureShellDetails extends SecureShellDetails {
```
ThomasJClark: Default usernames and passwords is not a good reason to use inheritance.  Just pass the variables directly

ui/src/main/java/edu/wpi/grip/ui/util/deployment/HeadlessInstanceManager.java
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     *
+     * @param eventBus
+     * @param coreJar The jar with all of the core. This will be copied to the destination
+     * @param projectFile The project file to send over
+     * @param details The details regarding connecting to the secure shell
+     * @param deploymentCommands The commands required to start and stop GRIP
+     */
+    public HeadlessInstanceManager(EventBus eventBus, File coreJar, File projectFile, SecureShellDetails details, DeploymentCommands deploymentCommands) {
```
ThomasJClark: This class should be hardcoded to only do FRC.  Abstracting away certain parts of it will not make it any easier to port it to other platforms, since FRC is the only one where we have an exact set of commands that we know will work.

Of course, more functionality can be added later, but for now this is [premature encapsulation](http://www.adam-bien.com/roller/abien/entry/premature_encapsulation_is_the_root).

9/10

ui/src/test/java/edu/wpi/grip/ui/util/deployment/HeadlessInstanceManagerTest.java
```diff
@@ -0,0 +1,41 @@
...
+        SecureShellDetails details = new SecureShellDetails.FRCSecureShellDetails(191);
+        DeploymentCommands deploymentCommands = new DeploymentCommands.FCRDeploymentCommands();
+        File coreJar = new File(edu.wpi.grip.core.Main.class.getProtectionDomain().getCodeSource().getLocation().toURI().getPath());
+        File testSave = new File(HeadlessInstanceManager.class.getResource("LoadThreeStepsTestSave.grip").toURI());
+
+        headlessInstanceManager = new HeadlessInstanceManager(new EventBus(), coreJar, testSave, details, deploymentCommands);
+    }
+
+    @Test(timeout = 30000)
+    @Ignore("This requires a robot to test on")
```
ThomasJClark: This should not be an ignored unit test.  It's okay to write something like this for your own testing purposes, but this is useless to us as a unit test because we can't run it, and it makes it seem like our code is broken since we ignore it.

ui/src/main/java/edu/wpi/grip/ui/util/deployment/HeadlessInstanceManager.java
```diff
@@ -0,0 +1,155 @@
...
+import java.net.URLDecoder;
+import java.nio.file.Paths;
+import java.util.Optional;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Controls an instance of GRIP running on a remote device.
+ */
+public class HeadlessInstanceManager implements StartStoppable {
```
ThomasJClark: This class never runs in headless mode afaik, so a better name might be "DeploymentManager"

8/10

ThomasJClark:
Formatting 11/10

ui/src/main/java/edu/wpi/grip/ui/util/deployment/HeadlessInstanceManager.java
JLLeitschuh: We know it needs to be abstract in the future so I did it this way. I'm not going to de-factor it to make it more hard coded. That will just add more work in the future.

ThomasJClark: I disagree.  We don't know that we'll ever add more ways to deploy that look exactly like FRC but with different commands and passwords plugged in.  In fact, I *really* doubt it.  Even if we do add more automatic deploy buttons, they almost certainly won't perfectly match this same pattern, so it'll be even more work.

What's with your hesitation to "de-factor" things?  Not everything is worth wrapping in layers and layers of abstraction, and undoing some of that abstraction is definitely a good thing.

ui/src/test/java/edu/wpi/grip/ui/util/deployment/HeadlessInstanceManagerTest.java
JLLeitschuh: It's not useless. This is useful as a permanent test for when we want to test on the robo-rio.
This could probably be adapted to run with jenkins later.
I'm going to leave it here because I'm probably going to need to use this later.

ui/src/main/java/edu/wpi/grip/ui/util/deployment/HeadlessInstanceManager.java
JLLeitschuh: It doesn't just deploy though. It also starts and stops.

ui/src/main/java/edu/wpi/grip/ui/util/deployment/HeadlessInstanceManager.java
JLLeitschuh: This is also not really premature encapsulation, it's the single responsibility principle.

ui/src/main/java/edu/wpi/grip/ui/util/deployment/DeployedInstanceManager.java
```diff
@@ -0,0 +1,229 @@
...
+                scpFileToTarget(projectFile);
+                deferred.notify(100);
+                deferred.resolve(this);
+            } catch (IOException e) {
+                deferred.reject(e);
+            }
+        }, "SCP File Deployer");
+        executor.setDaemon(true);
+        executor.setUncaughtExceptionHandler((t, e) -> deferred.reject(e));
+        executor.run();
```
JLLeitschuh: Will do.

core/src/main/java/edu/wpi/grip/core/Main.java
```diff
@@ -30,6 +32,9 @@ public static void main(String[] args) throws Exception {
         // Open a project from a .grip file specified on the command line
         project.open(new File(projectPath));
 
+
+        // This is done in order to indicate to the user using the deployment UI that this is running
+        System.out.println("Running!");
```
JLLeitschuh: `"SUCCESS! The project is running in headless mode!"`

ui/src/main/java/edu/wpi/grip/ui/deployment/FRCDeploymentOptionsController.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @Override
+    protected void postInit() {
+        final Label label = new Label("Team Number");
+        final SpinnerValueFactory.IntegerSpinnerValueFactory spinnerValueFactory = new SpinnerValueFactory.IntegerSpinnerValueFactory(0, Integer.MAX_VALUE, 190);
+        this.teamNumberSpinner = new Spinner(spinnerValueFactory);
+        this.teamNumberSpinner.setEditable(true);
+        // Ensure the value entered is only a number
+        this.teamNumberSpinner.getEditor().textProperty().addListener((observable, oldValue, newValue) -> {
+            if (newValue.equals("")) {
```
JLLeitschuh: :+1:

ui/src/main/java/edu/wpi/grip/ui/DeployerController.java
```diff
@@ -0,0 +1,142 @@
...
+    private ProgressBar progressIndicator;
+
+    private final EventBus eventBus;
+    private final DeployedInstanceManager.Factory deployedInstanceManagerFactory;
+    private final DeploymentOptionsControllersFactory optionsControllersFactory;
+
+    private class StreamToTextArea extends OutputStream {
+        private final TextArea outputArea;
+
+        public StreamToTextArea(TextArea outputArea) {
```
JLLeitschuh: :+1:

ui/src/main/java/edu/wpi/grip/ui/DeployerController.java
```diff
@@ -0,0 +1,142 @@
...
+    private TextArea stdOutStreamTextArea;
+
+    @FXML
+    private TextArea stdErrStreamTextArea;
+
+    @FXML
+    private ProgressBar progressIndicator;
+
+    private final EventBus eventBus;
+    private final DeployedInstanceManager.Factory deployedInstanceManagerFactory;
```
JLLeitschuh: :+1:

PR #213, opened 2015-12-14
Fix Platform.runLater() usage

ThomasJClark:
Our code is not really written to handle asyncronous GUI updates in many
situations, such as updating the pipeline view when steps are
added/deleted/moved.  So, in these situations we should block the
calling thread until the GUI thread has finished performing the
specified action.

In addition, it is not necessary to call Platorm.runLater() from the GUI
thread.

### DIFF
--- a/ui/src/main/java/edu/wpi/grip/ui/pipeline/AddSourceView.java
+++ b/ui/src/main/java/edu/wpi/grip/ui/pipeline/AddSourceView.java
@@ -164,7 +164,7 @@ public class AddSourceView extends HBox {
             } catch (IOException e) {
                 // This will run it again with the new values retrieved by the supplier
                 failureCallback.accept(e);
-                Platform.runLater(() -> loadCamera(dialog, cameraSourceSupplier, failureCallback));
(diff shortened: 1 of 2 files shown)


ui/src/main/java/edu/wpi/grip/ui/pipeline/AddSourceView.java
```diff
@@ -164,7 +164,7 @@ private void loadCamera(Dialog<ButtonType> dialog, SupplierWithIO<CameraSource>
             } catch (IOException e) {
                 // This will run it again with the new values retrieved by the supplier
                 failureCallback.accept(e);
-                Platform.runLater(() -> loadCamera(dialog, cameraSourceSupplier, failureCallback));
+                loadCamera(dialog, cameraSourceSupplier, failureCallback);
```
JLLeitschuh: Did you actually test that this works?

ThomasJClark: Yes

PR #219, opened 2015-12-22
Exclude unused stuff from the jar

ThomasJClark:
This brings the size of the headless jar down almost 30%, which should both make the distribution smaller (and probably more likely to be downloaded by people), and make the deploy time quicker.

On my system (Linux x86_64), the core jar is now 34.3 MB instead of 48 MB.


JLLeitschuh:
Sure, this seems to be okay.

PR #221, opened 2015-12-27
Use the default path for Axis M1011 cameras if none is specified

### LINKED ISSUE #220: NullPointerException: null
What were the actions performed prior to this error appearing?
Axis M1011 camera was added

Message: Camera Frame Grabber Thread crashed with uncaught exception

## System Info:

Property Name | Property 
 ----- | -----
javafx.version | 8.0.66
java.runtime.name | Java(TM) SE Runtime Environment
java.vm.version | 25.66-b17
java.vm.vendor | Oracle Corporation
java.vm.name | Java HotSpot(TM) 64-Bit Server VM
java.runtime.version | 1.8.0_66-b17
java.awt.graphicsenv | sun.awt.Win32GraphicsEnvironment
javafx.runtime.version | 8.0.66-b17
os.name | Windows 7
os.version | 6.1
os.arch | amd64
file.encoding | Cp1252
java.vm.info | mixed mode
java.version | 1.8.0_66
sun.arch.data.model | 64
sun.cpu.endian | little

## Stack Trace:
```java
java.lang.NullPointerException
	at edu.wpi.grip.core.sources.IPCameraFrameGrabber.grab(IPCameraFrameGrabber.java:129)
	at edu.wpi.grip.core.sources.CameraSource.lambda$start$18(CameraSource.java:144)
	at java.lang.Thread.run(Thread.java:745)

```

ThomasJClark:
Most FRC teams are accustomed to using something like http://10.1.90.11 as the path to their camera, but our current implementation requires you to write out something like http://10.1.90.11/mjpg/video.mjpg.  We don't want to make GRIP *only* work with FRC-style setups, but this change makes it at least use that path by default if no path is included in the url.

This solves a problem that both  GitHub user yuryws and Brad have reported, so I guess we can't expect people to just figure out to add /mjpg/video.mjpg" manually.

Closes #220

### DIFF
--- a/core/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
+++ b/core/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
@@ -27,6 +27,13 @@ import static com.google.common.base.Preconditions.checkNotNull;
 @XStreamAlias(value = "grip:Camera")
 public final class CameraSource extends Source implements StartStoppable {
 
+    /**
+     * The path that Axis cameras stream MJPEG videos from.  Although any URL can be supplied
+     * {@link #CameraSource(EventBus, String)}, allowing this to work with basically any network video stream, this
+     * default path allows the Axis M1011 cameras used in FRC to work when only an IP address is supplied.
+     */
+    public final static String DEFAULT_IP_CAMERA_PATH = "/mjpg/video.mjpg";
+
     private final static String DEVICE_NUMBER_PROPERTY = "deviceNumber";
     private final static String ADDRESS_PROPERTY = "address";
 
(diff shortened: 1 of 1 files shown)


JLLeitschuh:
Are there any webcams that this will prevent from being used this way? IE, the way to get the webcam stream will be only with the IP address?

Also, this should not be in the core. This should be done in the UI.

ThomasJClark:
> Are there any webcams that this will prevent from being used this way? IE, the way to get the webcam stream will be only with the IP address?

I doubt it. IP cameras typically provide some sort of control panel thing as their index page. 

> Also, this should not be in the core. This should be done in the UI.

Why? That's a pretty unusual claim, I'd expect some kind of argument behind it.

PR #222, opened 2015-12-28
NetworkTables

ThomasJClark:
This change mainly adds a new operation called `NTPublishOperation`.  Unlike most operations, there can be multiple instances of this class (one for each type that can be published), although for now there's only one for `ContoursReport`.

We should be able to make any report object be publishable to NetworkTables by making it implement `NTPublishable` and adding one or more accessors like this:

    @NTValue(key = "foo")
    public double getFoo() {
        return this.foo;
    }

We could theoretically have a method in `NTPublishOperation` that report classes implement to return all publishable properties, but this has a few major disadvantages:

* There's no way to know what things can be published until we have an instance of the class, so we wouldn't be able to actually populate the checkboxes in the GUI.
* We have to compute all of the properties at once.  If a property like area or something involves some intensive calculations, we shouldn't do those calculations if the checkbox isn't checked.

So, instead, we use the `@NTValue` annotation to define publishable properties.  (We could also theoretically not have the `NTPublishable` interface at all, since it has no methods, but I thought it would be good to make sure only things explicitly marked as publishable are published)

The end result looks like this:

nt


ThomasJClark:
Besides adding more publishable things, the only thing left to do is figure out how to get the NetworkTables server.  Right now it's hardcoded as localhost, but I guess we need some quick and easy way to get the team number.

Comments are welcome, but I'm gonna try to figure this out asap.

core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
```diff
@@ -0,0 +1,113 @@
...
+
+    public NTPublishOperation(Class<T> type) {
+        NetworkTable.setClientMode();
+        NetworkTable.setIPAddress("localhost"); // TODO
+
+        this.table = NetworkTable.getTable("GRIP");
+        this.type = type;
+
+        // Any accessor method with an @NTValue annotation can be published to NetworkTables.
+        for (Method method : type.getDeclaredMethods())
```
JLLeitschuh: :+1:

core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
```diff
@@ -0,0 +1,113 @@
...
+ * An operation that publishes any type that implements {@link NTPublishable} to NetworkTables.
+ * <p>
+ * To be publishable, a type should have one or more accessor methods annotated with {@link NTValue}.  This is done
+ * with annotations instead of methods
+ */
+public class NTPublishOperation<T extends NTPublishable> implements Operation {
+
+    private final NetworkTable table;
+    private final Class<T> type;
+    private final List<Method> ntValueMethods = new ArrayList<>();
```
JLLeitschuh: Create this list inside the constructor with the correct size then assign it. That way the list's size is fixed and an array larger than necessary doesn't need to be allocated.

core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
```diff
@@ -0,0 +1,113 @@
...
+            for (Method method : ntValueMethods) {
+                String key = method.getAnnotation(NTValue.class).key();
+                if ((Boolean) inputs[i++].getValue().get()) {
+                    subtable.putValue(key, method.invoke(value));
+                } else {
+                    subtable.delete(key);
+                }
+            }
+        } catch (IllegalAccessException | InvocationTargetException e) {
+            throw new RuntimeException(e);
```
JLLeitschuh: Yea, don't do this. Something throw an illegal state exception with a message.

ui/src/main/java/edu/wpi/grip/ui/pipeline/input/TextFieldInputSocketView.java
```diff
@@ -0,0 +1,38 @@
...
+import edu.wpi.grip.core.InputSocket;
+import edu.wpi.grip.core.events.SocketChangedEvent;
+import javafx.scene.control.TextField;
+
+/**
+ * An {@link InputSocketView} that lets the user set the value of a number with a {@link javafx.scene.control.TextField}
+ */
+public class TextFieldInputSocketView extends InputSocketView<String> {
+
+    final TextField textField;
```
JLLeitschuh: :+1:

core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
```diff
@@ -0,0 +1,113 @@
...
+    private final NetworkTable table;
+    private final Class<T> type;
+    private final List<Method> ntValueMethods = new ArrayList<>();
+
+    public NTPublishOperation(Class<T> type) {
+        NetworkTable.setClientMode();
+        NetworkTable.setIPAddress("localhost"); // TODO
+
+        this.table = NetworkTable.getTable("GRIP");
+        this.type = type;
```
JLLeitschuh: Check not null

core/src/main/java/edu/wpi/grip/core/operations/networktables/NTValue.java
```diff
@@ -0,0 +1,25 @@
...
+ * object (such as {@link ContoursReport} to have arbitrarily many publishable values.  More values (like x, y, width,
+ * height, etc...) can be added simply by adding more accessors with this annotation.
+ * <p>
+ * One potential way of having multiple values without using annotations might be to make the interface method return
+ * a list, but this would prevent us from knowing how many values there are and what their names are without having
+ * an instance of the object being published.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+public @interface NTValue {
```
JLLeitschuh: Should this be inherited?

core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
```diff
@@ -0,0 +1,113 @@
...
+
+        // Any accessor method with an @NTValue annotation can be published to NetworkTables.
+        for (Method method : type.getDeclaredMethods())
+            if (method.getAnnotation(NTValue.class) != null)
+                ntValueMethods.add(method);
```
JLLeitschuh: Check that all methods annotated don't take arguments. This will just cause problems later.

JLLeitschuh:
I like the way you handled this. Can you add some tests for it too?

ui/src/main/java/edu/wpi/grip/ui/pipeline/input/TextFieldInputSocketView.java
ThomasJClark: k

core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
ThomasJClark: k

core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
ThomasJClark: k

core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
ThomasJClark: It doesn't matter how many array spaces are allocated.  There is currently only one instance of that array in the application, so there's no need to introduce complexity to save a couple of bytes.

core/src/main/java/edu/wpi/grip/core/operations/networktables/NTValue.java
ThomasJClark: ?

ThomasJClark: I don't see how that could apply to here?

core/src/main/java/edu/wpi/grip/core/operations/composite/ContoursReport.java
```diff
@@ -1,15 +1,26 @@
...
+import static org.bytedeco.javacpp.opencv_imgproc.boundingRect;
+import static org.bytedeco.javacpp.opencv_imgproc.contourArea;
 
 /**
  * The output of {@link FindContoursOperation}.  This stores a list of contours (which is basically a list of points) in
  * OpenCV objects, as well as the width and height of the image that the contours are from, to give context to the
  * points.
  */
-public final class ContoursReport {
+public final class ContoursReport implements NTPublishable {
```
JLLeitschuh: This would be easier if this class was immutable. Then this object wouldn't have to have to have thread locks on operations that should be simple reads.

core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
JLLeitschuh: But this list doesn't need to be dynamic. So why are you using something with variable size. You know the size at constructor time.

core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
```diff
@@ -0,0 +1,121 @@
...
+        final InputSocket<?>[] sockets = new InputSocket[2 + ntValueMethods.size()];
+        int i = 0;
+
+        // Create an input for the actual object being published
+        sockets[i++] = new InputSocket<>(eventBus,
+                new SocketHint.Builder<>(type).identifier("Value").initialValue(null).build());
+
+        // Create a string input for the key used by NetworkTables
+        sockets[i++] = new InputSocket<>(eventBus,
+                SocketHints.Inputs.createTextSocketHint("Subtable Name", "my" + type.getSimpleName()));
```
JLLeitschuh: Should there be a space there?

core/src/main/java/edu/wpi/grip/core/SocketHint.java
```diff
@@ -15,7 +15,7 @@
  * interface to learn how to talk to <code>Algorithm</code>s.
  */
 public class SocketHint<T> {
-    public enum View {NONE, SPINNER, SLIDER, RANGE, SELECT, CHECKBOX}
+    public enum View {NONE, TEXT, SLIDER, RANGE, SELECT, CHECKBOX}
```
JLLeitschuh: `TEXT` doesn't really make sense here. What about `FIELD`

core/src/main/java/edu/wpi/grip/core/operations/Operations.java
```diff
@@ -30,5 +31,6 @@ public static void addOperations(EventBus eventBus) {
         eventBus.post(new OperationAddedEvent(new NewPointOperation()));
         eventBus.post(new OperationAddedEvent(new NewSizeOperation()));
         eventBus.post(new OperationAddedEvent(new MatFieldAccessor()));
+        eventBus.post(new OperationAddedEvent(new NTPublishOperation<>(ContoursReport.class)));
```
JLLeitschuh: :+1:

core/src/main/java/edu/wpi/grip/core/SocketHint.java
ThomasJClark: It's `TEXT` because it's a text field.  It's just all of them are fields.

core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
ThomasJClark: no

JLLeitschuh: Okay

core/src/main/java/edu/wpi/grip/core/operations/composite/ContoursReport.java
```diff
@@ -44,4 +56,77 @@ public int getRows() {
...
+        boundingBoxes = Optional.empty();
+    }
+
+    /**
+     * Compute the bounding boxes of all contours (if they haven't already been computed).  Bounding boxes are used
+     * to compute several different properties, so it's probably not a good idea to compute them over and over again.
+     */
+    private List<Rect> computeBoundingBoxes() {
+        if (!boundingBoxes.isPresent()) {
+            List<Rect> bb = new ArrayList<>();
```
JLLeitschuh: Since this is being reallocated every time and this could be called upwards of 30 times per second this should be allocated with the size of `contours.size`

core/src/main/java/edu/wpi/grip/core/ProjectSettings.java
```diff
@@ -0,0 +1,55 @@
...
+                .add("teamNumber", teamNumber)
+                .toString();
+    }
+
+    @Override
+    public ProjectSettings clone() {
+        try {
+            return (ProjectSettings) super.clone();
+        } catch (CloneNotSupportedException impossible) {
+            return null;
```
JLLeitschuh: Why return null?
This should not just eat this exception.

ui/src/main/java/edu/wpi/grip/ui/ProjectSettingsEditor.java
```diff
@@ -0,0 +1,43 @@
...
+import javafx.scene.control.DialogPane;
+import javafx.scene.image.Image;
+import javafx.scene.image.ImageView;
+import org.controlsfx.control.PropertySheet;
+import org.controlsfx.property.BeanPropertyUtils;
+
+/**
+ * A JavaFX dialog that lets the user edit the project {@link ProjectSettings}
+ */
+public class ProjectSettingsEditor extends Dialog<ButtonType> {
```
JLLeitschuh: http://controlsfx.bitbucket.org/org/controlsfx/control/PropertySheet.html

ui/src/main/java/edu/wpi/grip/ui/ProjectSettingsEditor.java
```diff
@@ -0,0 +1,43 @@
...
+        DialogPane pane = getDialogPane();
+        pane.getButtonTypes().setAll(ButtonType.OK, ButtonType.CANCEL);
+        pane.setContent(propertySheet);
+        pane.styleProperty().bind(root.styleProperty());
+        pane.getStylesheets().addAll(root.getStylesheets());
+        pane.setPrefSize(DPIUtility.SETTINGS_DIALOG_SIZE, DPIUtility.SETTINGS_DIALOG_SIZE);
+
+        setTitle("Settings");
+        setHeaderText("Settings");
+        setGraphic(new ImageView(new Image(getClass().getResourceAsStream("icons/settings.png"))));
```
JLLeitschuh: http://fortawesome.github.io/Font-Awesome/icon/cog/

ui/src/main/resources/edu/wpi/grip/ui/MainWindow.fxml
```diff
@@ -96,6 +96,23 @@
...
+                    <graphic>
+                        <ImageView styleClass="menu-graphic">
+                            <fitWidth>
+                                <DPIUtility fx:constant="SMALL_ICON_SIZE"/>
+                            </fitWidth>
+                            <fitHeight>
+                                <DPIUtility fx:constant="SMALL_ICON_SIZE"/>
+                            </fitHeight>
+                            <image>
+                                <Image url="@icons/settings.png"/>
```
JLLeitschuh: You can use font awesome here.

core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
JLLeitschuh: The robot agrees.

core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
```diff
@@ -0,0 +1,139 @@
...
+
+    public NTPublishOperation(Class<T> type) {
+        this.table = NetworkTable.getTable("GRIP");
+        this.type = checkNotNull(type, "Type was null");
+
+        // Any accessor method with an @NTValue annotation can be published to NetworkTables.
+        for (Method method : type.getDeclaredMethods()) {
+            if (method.getAnnotation(NTValue.class) != null) {
+                if (method.getParameters().length > 0) {
+                    throw new IllegalArgumentException("@NTValue method must have 0 parameters: " + method);
```
JLLeitschuh: :+1:

core/src/main/java/edu/wpi/grip/core/ProjectSettings.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public void setNetworkProtocol(NetworkProtocol networkProtocol) {
+        this.networkProtocol = networkProtocol;
+    }
+
+    public NetworkProtocol getNetworkProtocol() {
+        return networkProtocol;
+    }
+
+    public void setTeamNumber(int teamNumber) {
```
JLLeitschuh: Shouldn't a team number of zero automatically assign no network tables protocol?

core/src/main/java/edu/wpi/grip/core/SocketHint.java
JLLeitschuh: Fine.

core/src/main/java/edu/wpi/grip/core/operations/Operations.java
```diff
@@ -30,5 +32,9 @@ public static void addOperations(EventBus eventBus) {
         eventBus.post(new OperationAddedEvent(new NewPointOperation()));
         eventBus.post(new OperationAddedEvent(new NewSizeOperation()));
         eventBus.post(new OperationAddedEvent(new MatFieldAccessor()));
+
+        Operation publishContours = new NTPublishOperation<>(ContoursReport.class);
+        eventBus.register(publishContours);
```
JLLeitschuh: Oh, okay. Not really a fan of registering operations on the event bus but sure, whatever.

ui/src/main/java/edu/wpi/grip/ui/MainWindowView.java
```diff
@@ -190,6 +189,18 @@ public boolean saveProjectAs() throws IOException {
     }
 
     @FXML
+    public void showProjectSettingsEditor() {
+        final ProjectSettings projectSettings = pipeline.getPipeline().getProjectSettings().clone();
```
JLLeitschuh: Explain why this is cloned? Also, why is this done instead of using a copy constructor?

core/src/main/java/edu/wpi/grip/core/events/ProjectSettingsChangedEvent.java
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * This event is posted after the {@link ProjectSettings} are changed so anything that relies on them can immediately
+ * update without restarting the application.
+ */
+public class ProjectSettingsChangedEvent {
+    private final ProjectSettings projectSettings;
+
+    public ProjectSettingsChangedEvent(ProjectSettings projectSettings) {
+        this.projectSettings = projectSettings;
```
JLLeitschuh: Check not null

core/src/main/java/edu/wpi/grip/core/operations/composite/ContoursReport.java
JLLeitschuh: Why is this event allocated into a list? This could simply be stored in an array.

core/src/main/java/edu/wpi/grip/core/Operation.java
```diff
@@ -61,7 +61,6 @@ default void perform(InputSocket<?>[] inputs, OutputSocket<?>[] outputs, Optiona
         perform(inputs, outputs);
     }
 
-    @Deprecated
```
JLLeitschuh: I think this method should actually be removed at some point. Why do you disagree?
Also, why is this part of this PR?

build.gradle
```diff
@@ -88,6 +91,7 @@ project(":core") {
         compile group: 'com.thoughtworks.xstream', name: 'xstream', version: '1.4.8'
         compile group: 'org.apache.commons', name: 'commons-lang3', version: '3.4'
         compile group: 'com.google.guava', name: 'guava', version: '18.0'
+        compile group: 'edu.wpi.first.wpilib.networktables.java', name: 'NetworkTables', version: '3.0.0-SNAPSHOT', classifier: 'desktop'
```
JLLeitschuh: What about the version for running on the robot? Does this also work?

core/src/main/java/edu/wpi/grip/core/ProjectSettings.java
```diff
@@ -0,0 +1,55 @@
...
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(this)
+                .add("networkProtocol", networkProtocol)
+                .add("teamNumber", teamNumber)
+                .toString();
+    }
+
+    @Override
+    public ProjectSettings clone() {
```
JLLeitschuh: If you're not going to use a copy constructor then you need to do this instead.

ui/src/main/java/edu/wpi/grip/ui/MainWindowView.java
```diff
@@ -190,6 +189,18 @@ public boolean saveProjectAs() throws IOException {
     }
 
     @FXML
+    public void showProjectSettingsEditor() {
```
JLLeitschuh: These methods don't technically need to be public.

ui/src/main/java/edu/wpi/grip/ui/ProjectSettingsEditor.java
```diff
@@ -0,0 +1,43 @@
...
+import javafx.scene.image.ImageView;
+import org.controlsfx.control.PropertySheet;
+import org.controlsfx.property.BeanPropertyUtils;
+
+/**
+ * A JavaFX dialog that lets the user edit the project {@link ProjectSettings}
+ */
+public class ProjectSettingsEditor extends Dialog<ButtonType> {
+
+    private final ProjectSettings projectSettings;
```
JLLeitschuh: :+1:

ui/src/main/java/edu/wpi/grip/ui/ProjectSettingsEditor.java
```diff
@@ -0,0 +1,43 @@
...
+    private final PropertySheet propertySheet;
```
JLLeitschuh: :+1:

core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
```diff
@@ -0,0 +1,139 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * An operation that publishes any type that implements {@link NTPublishable} to NetworkTables.
+ * <p>
+ * To be publishable, a type should have one or more accessor methods annotated with {@link NTValue}.  This is done
+ * with annotations instead of methods
+ */
+public class NTPublishOperation<T extends NTPublishable> implements Operation {
```
JLLeitschuh: Can you add a test for this class?

core/src/main/java/edu/wpi/grip/core/ProjectSettings.java
ThomasJClark: This exception does not actually happen

ui/src/main/resources/edu/wpi/grip/ui/MainWindow.fxml
ThomasJClark: How?  I couldn't figure out how to use it from FXML

core/src/main/java/edu/wpi/grip/core/ProjectSettings.java
ThomasJClark: I don't think so.  If you don't want to use NetworkTables, it makes more sense to choose `NONE` as your protocol then to know that 0 means not to use networktables

ui/src/main/java/edu/wpi/grip/ui/MainWindowView.java
ThomasJClark: So it copies all of the fields, even if we add more preferences.  I'd rather not have one more thing to remember to update every time we add a new setting

core/src/main/java/edu/wpi/grip/core/ProjectSettings.java
```diff
@@ -0,0 +1,55 @@
...
+     * What protocol to use to publish values.  Only NetworkTables is supported right now, but in the future we may
+     * make the publish operations work with other protocols like ROS.
+     */
+    public enum NetworkProtocol {
+        NETWORK_TABLES,
+        NONE
+    }
+
+    private NetworkProtocol networkProtocol = NetworkProtocol.NETWORK_TABLES;
+    private int teamNumber = 0;
```
JLLeitschuh: Can this be an `Optional<Integer>` or will that mess with the whole bean thing?

ui/src/main/resources/edu/wpi/grip/ui/MainWindow.fxml
JLLeitschuh: Something like:
```xml
<graphic>
        <Glyph fontFamily="FontAwesome" icon="GEAR"/>
    </graphic>
```
https://bitbucket.org/controlsfx/controlsfx/pull-requests/381/basic-refactoring-of-glyph-font-system-306/diff

core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
```diff
@@ -0,0 +1,139 @@
...
+            for (Method method : ntValueMethods) {
+                String key = method.getAnnotation(NTValue.class).key();
+                if ((Boolean) inputs[i++].getValue().get()) {
+                    subtable.putValue(key, method.invoke(value));
+                } else {
+                    subtable.delete(key);
+                }
+            }
+        } catch (IllegalAccessException | InvocationTargetException e) {
+            throw new RuntimeException(e);
```
JLLeitschuh: I still agree with the robot here.

ui/src/main/java/edu/wpi/grip/ui/pipeline/input/TextFieldInputSocketView.java
```diff
@@ -0,0 +1,38 @@
...
+            if (!textField.isFocused()) socket.setValue(textField.getText());
+        });
+
+        setContent(textField);
+    }
+
+    @Subscribe
+    public void updateTextFieldFromSocket(SocketChangedEvent event) {
+        if (event.getSocket() == getSocket()) {
+            textField.setText(getSocket().getValue().get());
```
JLLeitschuh:
```
Platform.runLater(() -> {
});
```
Only use GRIPPlatform if this needs to be run in a time sensitive manner.

ui/src/main/resources/edu/wpi/grip/ui/MainWindow.fxml
JLLeitschuh: Okay, I messed with it enough and finally figured it out:
```
<Glyph fontFamily="FontAwesome" icon="UPLOAD" style="-fx-text-fill: black;"/>
```
The key is the style otherwise it will be transparent (at least in the menus)

core/src/main/java/edu/wpi/grip/core/ProjectSettings.java
JLLeitschuh: We should really avoid the usage of the clone method. Its broken. http://stackoverflow.com/a/1106159/3708426

ThomasJClark: We're not using the clone method like that.  `clone()` actually works perfectly for our use case because we have a simple JavaBean object that we might add fields to in the future, and we want to make sure we don't introduce a bug by forgetting to copy those fields.

core/src/main/java/edu/wpi/grip/core/Operation.java
ThomasJClark: 1. Most operations don't use the data parameter, it was just a quick hack to let us cache some Mats.
2. Because I need to use this method in this PR

build.gradle
ThomasJClark: I don't know, I don't have access to a roborio

core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
ThomasJClark: Yeah, later

core/src/main/java/edu/wpi/grip/core/ProjectSettings.java
ThomasJClark: That will mess it up

ui/src/main/java/edu/wpi/grip/ui/pipeline/input/TextFieldInputSocketView.java
ThomasJClark: k

core/src/main/java/edu/wpi/grip/core/events/ProjectSettingsChangedEvent.java
ThomasJClark: k

core/src/main/java/edu/wpi/grip/core/ProjectSettings.java
JLLeitschuh: Okay

core/src/main/java/edu/wpi/grip/core/ProjectSettings.java
JLLeitschuh: Then make the clone method final.

core/src/main/java/edu/wpi/grip/core/ProjectSettings.java
JLLeitschuh: Then throw something. Don't just return null 9/10

core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * An operation that publishes any type that implements {@link NTPublishable} to NetworkTables.
+ * <p>
+ * To be publishable, a type should have one or more accessor methods annotated with {@link NTValue}.  This is done
+ * with annotations instead of methods
+ */
+public class NTPublishOperation<T extends NTPublishable> implements Operation {
+
+    static {
+        NetworkTable.setClientMode();
```
JLLeitschuh: Should this really be done here or in main?

Issue #225, opened 2015-12-29
ArrayIndexOutOfBoundsException


dangyogi:
What were the actions performed prior to this error appearing?

I had a webcam tied to an HSV Threshold and was viewing both in the top half of the screen.

The information below was captured by clicking the "Copy to Clipboard" button on the exception window and then pasting here.

Message: The UI Thread threw an uncaught exception

## System Info:

Property Name | Property 
 ----- | -----
javafx.version | 8.0.66
java.runtime.name | Java(TM) SE Runtime Environment
java.vm.version | 25.66-b17
java.vm.vendor | Oracle Corporation
java.vm.name | Java HotSpot(TM) 64-Bit Server VM
java.runtime.version | 1.8.0_66-b17
java.awt.graphicsenv | sun.awt.X11GraphicsEnvironment
javafx.runtime.version | 8.0.66-b17
os.name | Linux
os.version | 3.13.0-74-generic
os.arch | amd64
file.encoding | UTF-8
java.vm.info | mixed mode
java.version | 1.8.0_66
sun.arch.data.model | 64
sun.cpu.endian | little

## Stack Trace:
```java
java.lang.ArrayIndexOutOfBoundsException

```

JLLeitschuh:
Thanks.
What was the action you performed immediately preceding the exception appearing?
This may be related to #155

dangyogi:
Sorry, I don't really remember.  I'm just starting out with GRIP, so it was
pretty simple.  I had one webcam with the "eye" thing opened on top, and
that was linked to an HSV (I think, might have been HSL or RGB, but I'm
pretty sure HSV) threshold with its "eye" thing also open on top and I was
playing with the HSV threshold settings.  So two objects with two displays
on top.

But I'm kind of thinking too that it *may* have happened out of the blue
(not in response to me changing a setting).  "out of the blue" means that
the two "eye" views are open on top and the system is dynamically
processing images, but that I wasn't touching anything.

The webcam is a built-in laptop camera with a frame rate of ~30.  Not sure
what the resolution is.

I was surprised that the "Copy Error Info" button (or whatever it was
called) on the exception window didn't capture a stack backtrace to show
you where in the code the error happened.  Is there something that I can
set to turn on a backtrace?  (I'm not a java developer).

-Bruce

JLLeitschuh:
Yea, I'm confused as well. I've seen errors like that before and I'm not sure why the stack trace got lost.
[...]

PR #227, opened 2015-12-30
Prevents JIT from eating stack traces

JLLeitschuh:
This will inflate the jar a small ammount and
slow down throwing of errors but will allow us
to get stack traces where we normally would have
lost them.
Thanks to this post:
https://plumbr.eu/blog/java/on-a-quest-for-missing-stacktraces

Related #225

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,6 +33,10 @@ allprojects {
 
     version = '0.2.1'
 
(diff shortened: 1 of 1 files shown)


ThomasJClark:
What does this have to do with JIT? I'm not sure I understand.

PR #229, opened 2016-01-01
Add a button to respond to the ExceptionWitness

JLLeitschuh:
Finalized Popover
Popover closed

PR #230, opened 2016-01-01
Fixes the sink test throwing a class cast exception

JLLeitschuh:
I know that this test is being removed with dependency injection anyways but right now it seems to be breaking #229


ThomasJClark:
I don't care if the sink test doesn't work, we're getting rid of it.

> it seems to be breaking #229

I think you have it backwards.  #229 is breaking SinkTest, which is okay because SinkTest doesn't matter.

JLLeitschuh:
I can't merge #229 until we fix this.

JLLeitschuh:
Invalid as test was removed

PR #233, opened 2016-01-02
Updates to Gradle 2.10

JLLeitschuh:
Does what it says on the box.


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Tue Dec 08 15:02:34 EST 2015
+#Fri Jan 01 22:19:23 EST 2016
```
PaulaRudy: Shouldn't this not be here?

JLLeitschuh: This file gets updated every time the gradle version changes. This should be there (Unfortunately)

PR #259, opened 2016-01-04
Fix various deployment problems

JLLeitschuh:
- ARM Java does not support aop so we need to change what guice we are using
- Deploying can throw null pointers because writing text to the text box was
not in the UI thread.
- Logging was setup after the guice injector was loaded meaning problems with
injecting wouldn't be logged.
- The GRIPCore module assumed that the main file would be registered on the
event bus to log exceptions but it isn't until after the injector is created.
- The thread that manages starting GRIP on a remote device does not actually 
post a startstopped event if the shell crashes

### DIFF
--- a/core/src/main/java/edu/wpi/grip/core/GRIPCoreModule.java
+++ b/core/src/main/java/edu/wpi/grip/core/GRIPCoreModule.java
@@ -16,8 +16,8 @@ import edu.wpi.grip.core.sources.ImageFileSource;
 import edu.wpi.grip.core.sources.MultiImageFileSource;
 import edu.wpi.grip.core.util.ExceptionWitness;
 
-import java.util.logging.Level;
-import java.util.logging.Logger;
+import java.io.IOException;
+import java.util.logging.*;
 
 /**
  * A Guice {@link com.google.inject.Module} for GRIP's core package.  This is where instances of {@link Pipeline},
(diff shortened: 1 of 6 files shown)


core/src/main/java/edu/wpi/grip/core/GRIPCoreModule.java
```diff
@@ -29,6 +29,28 @@
     private final EventBus eventBus = new EventBus(this::onSubscriberException);
 
     public GRIPCoreModule() {
+        //Set up the global level logger. This handles IO for all loggers.
+        Logger globalLogger = LogManager.getLogManager().getLogger("");//This is our global logger
```
ThomasJClark: Not sure if this helps here, but Guice has built in bindings for `Logger`

https://github.com/google/guice/wiki/BuiltInBindings

core/src/main/java/edu/wpi/grip/core/Main.java
```diff
@@ -38,31 +39,10 @@ public void start(String[] args) throws IOException, InterruptedException {
         if (args.length != 1) {
             System.err.println("Usage: GRIP.jar project.grip");
             return;
+        } else {
+            System.out.println("Loading file " + args[0]);
```
ThomasJClark: Why not log?

core/src/main/java/edu/wpi/grip/core/GRIPCoreModule.java
JLLeitschuh: We still have to set it up.

core/src/main/java/edu/wpi/grip/core/Main.java
JLLeitschuh: Sure.

Issue #263, opened 2016-01-04
Saved file extension


AustinShalit:
Create a file extension that is automatically appended to the name of saved files to identify them as GRIP information.

.grip?

JLLeitschuh:
Sure, we thought about this but the only way to do that is to append it to the end of the file that the user has given us.
We should only do this if the file doesn't already exist.

AustinShalit:
Solved with #266

JLLeitschuh:
Thanks for closing!

PR #266, opened 2016-01-05
File extension

AustinShalit:
This should solve #263.  When saving a file that does not exist, the user is forced to save with the extension '.grip'.  The open file dialog limits the selection to '.grip' files with the option to allow all files.


JLLeitschuh:
Does force the user to append the .grip extension to the end or will it add it automatically if you don't?

AustinShalit:
On OS X it defaults to this:


If I click save the program saves a file called "Untitled.grip".
When I delete the .grip extension and click save the program saves a file called "Untitled.grip".

JLLeitschuh:
Cool! That's what we want!

JLLeitschuh:
Waiting on @JacisNonsense to test this out on a Windows machine before we merge it.

AustinShalit:
I just tested this on Windows 7 (dug an old laptop out of the closet).  Works identically to OS X.

Issue #112, opened 2015-11-10
Adaptive threshold step has formatting issues


PaulaRudy:
The step window for adaptiveThreshold has labels that are too long, causing them to be hard to read.
![screen shot 2015-11-10 at 1 30 08 pm](https://cloud.githubusercontent.com/assets/3857381/11071529/405042d2-87af-11e5-8aaa-14993ce0274a.png)

PR #278, opened 2016-01-05
Enum format fix

PaulaRudy:
Fixes it so enum "types" are not displayed in step gui element.This would close issue #112

### DIFF
--- a/ui/src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketController.java
+++ b/ui/src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketController.java
@@ -64,7 +64,7 @@ public class InputSocketController<T> implements Controller {
     @FXML
     public void initialize() {
         this.identifier.setText(this.socket.getSocketHint().getIdentifier());
-        if (!(this.socket.getSocketHint().getType().getSimpleName().contains("Enum"))) {
+        if (!(this.socket.getSocketHint().getType().isEnum())) {
             this.type.setText(this.socket.getSocketHint().getType().getSimpleName());
         }
         this.handle = socketHandleViewFactory.create(this.socket);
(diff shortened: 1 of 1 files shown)


ui/src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketController.java
```diff
@@ -64,7 +64,9 @@
     @FXML
     public void initialize() {
         this.identifier.setText(this.socket.getSocketHint().getIdentifier());
-        this.type.setText(this.socket.getSocketHint().getType().getSimpleName());
+        if (!(this.socket.getSocketHint().getType().getSimpleName().contains("Enum"))) {
```
ThomasJClark: This should be `this.socket.getSocketHint().getType().isEnum()`

ui/src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketController.java
```diff
@@ -64,7 +64,9 @@
...
+            this.type.setText(this.socket.getSocketHint().getType().getSimpleName());
+        }
         this.handle = socketHandleViewFactory.create(this.socket);
         root.add(this.handle, 0, 0);
```
ThomasJClark: Also don't add the handle if it's an enum

ui/src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketController.java
PaulaRudy: Yeah, that's probably a better option. I'll fix it.

ui/src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketController.java
PaulaRudy: Apparently this makes GRIP angry as it needs the handle for the XML. I agree that this might be something to fix later though.

Issue #280, opened 2016-01-06
"http://" before camera address


AustinShalit:
Teams are required to type http:// before typing the camera ip address.  Do we expect them to open any other kind of stream?  Can we fill the "http://" for them?

AustinShalit:
I will fix this

ThomasJClark:
Sounds good.

JLLeitschuh:
@nightpool Is this true?

ThomasJClark:
Even if it's not, we just make it only fill in the `http://` if nothing else is specified, so we don't break compatibility with other things

nightpool:
@JLLeitschuh uh can't think of any reason why not, but i mean it depends on how many kinds of protocols the underlying connection can support.

wouldn't the more elegent solution be to fill in `http://` if and only if they don't specify one?

AustinShalit:
#284 Solves this

AustinShalit:
thanks for closing

PR #284, opened 2016-01-06
Auto Fill Camera Prorocol

AustinShalit:
Automatically adds “http://“ to the textbox when adding an IP Camera if what is in the box is a valid IP address or the box ends in “.local”


AustinShalit:
Solves #280

Issue #269, opened 2016-01-05
Socket handles in sources don't fill in when connected


ThomasJClark:
This started happening after the dependency injection change was merged in.

![handle](https://cloud.githubusercontent.com/assets/3964980/12106849/a1ee7236-b331-11e5-8b76-8932f209828d.png)

JLLeitschuh:
I don't really know how those work. Can you take a crack at that one? You fixed this one time before didn't you?

ThomasJClark:
Ok, I'll look into it

AustinShalit:
I am going to try to tackle this.  Give me until morning

AustinShalit:
Solved with #285

AustinShalit:
@ThomasJClark Can this be closed?

ThomasJClark:
Yes.

Tip: If you put "Closes #269" in your PR message, it will automatically close this issue when it's merged

PR #285, opened 2016-01-06
Sources Socket Handles

AustinShalit:
This solves #269.  

Maybe this method should be rewritten to use the Source super class more.


ThomasJClark:
Nice catch, I totally forgot that `Socket` has `@Subscribe` methods.

Maybe it would be more future-proof to just have `eventBus.register(this);` in the constructor for `Socket`.  That way, we don't have to always remember to register new sockets

JLLeitschuh:
No, bad idea. See #155

AustinShalit:
How do we want to proceed?  I recommend we merge this PR for the 1.0.0 release and open a new issue to fix it for 1.0.1

JLLeitschuh:
The long term term fix will be that the sockets to be constructed using dependency injection.
This fix should work for now.

ThomasJClark:
> No, bad idea. See #155

That doesn't apply here because `Socket` subscribes to `ConnectionAddedEvent` and `ConnectionRemovedEvent`, which can't be sent before the socket is completely constructed (whoever posts  the event needs a reference to the socket to create the event)

JLLeitschuh:
Alternatively, we could make it so that the Connection calls `addConnection` on the `Socket` when `Connection` receives a `ConnectionAddedEvent` and then, that way, sockets don't even have to be registered on the event bus.

JLLeitschuh:
I think this is an appropriate way of handling this problem for the time being.
Do you agree @ThomasJClark or do you think it should be changed.

ThomasJClark:
> I think this is an appropriate way of handling this problem for the time being.

Does "this" mean the way this PR does it, or the way you proposed? (no on the first, yes on second)

JLLeitschuh:
Okay, fine I meant the first way.
@AustinShalit do you think you can resolve this the way I described?
Removing the subscribe methods from the socket and adding a `addConnection` and `removeConnection` method in their place?
If you can't get to this then just say so and I can take care of it.

AustinShalit:
@JLLeitschuh Yea.  I will take care of this

core/src/main/java/edu/wpi/grip/core/sources/MultiImageFileSource.java
```diff
@@ -34,6 +34,7 @@
 
     private final SocketHint<Mat> imageOutputHint = SocketHints.Inputs.createMatSocketHint("Image", true);
     private final OutputSocket<Mat> outputSocket;
+    private final EventBus eventBus;
```
JLLeitschuh: This should no longer be necessary

core/src/main/java/edu/wpi/grip/core/sources/MultiImageFileSource.java
```diff
@@ -88,6 +89,7 @@ private MultiImageFileSource(
             final String[] paths,
             final int index) {
         super(exceptionWitnessFactory);
+        this.eventBus = checkNotNull(eventBus, "Event Bus was null.");
```
JLLeitschuh: This too.

core/src/main/java/edu/wpi/grip/core/Socket.java
```diff
@@ -134,26 +134,25 @@ public Direction getDirection() {
...
 
-    @Subscribe
-    public void onConnectionRemoved(ConnectionRemovedEvent event) {
-        if (event.getConnection().getInputSocket() == this || event.getConnection().getOutputSocket() == this) {
-            this.connections.remove(event.getConnection());
+    /**
+     * @param connection The connection to remove from this socket.
+     */
+    public void removeConnection(Connection connection) {
+        this.connections.remove(connection);
```
JLLeitschuh: Add a `checkNotNull(connection, "Can not remove null connection")`

core/src/main/java/edu/wpi/grip/core/Socket.java
```diff
@@ -134,26 +134,25 @@ public Direction getDirection() {
         return ImmutableSet.copyOf(this.connections);
     }
 
+    /**
+     * @param connection The connection to add to this socket.
+     */
+    public void addConnection(Connection connection) {
+        this.connections.add(connection);
```
JLLeitschuh: `checkNotNull(connection, "Can not add null connection")`

core/src/main/java/edu/wpi/grip/core/Connection.java
```diff
@@ -64,14 +66,14 @@ public void removeConnection(StepRemovedEvent e) {
...
-                this.eventBus.post(new ConnectionRemovedEvent(this));
+                socket.removeConnection(this);
                 return;
             }
         }
 
         for (Socket socket : e.getStep().getInputSockets()) {
             if (socket == this.inputSocket || socket == this.outputSocket) {
-                this.eventBus.post(new ConnectionRemovedEvent(this));
+                socket.removeConnection(this);
```
JLLeitschuh: Yes! This is exactly right!

core/src/main/java/edu/wpi/grip/core/Connection.java
```diff
@@ -49,6 +49,8 @@
     public void onConnectionAdded(ConnectionAddedEvent event) {
         if (event.getConnection().equals(this)) {
             inputSocket.setValueOptional(outputSocket.getValue());
+            inputSocket.addConnection(this);
+            outputSocket.addConnection(this);
```
JLLeitschuh: Awesome!

core/src/main/java/edu/wpi/grip/core/sources/MultiImageFileSource.java
```diff
@@ -22,6 +21,7 @@
 import java.util.stream.Collectors;
 
 import static com.google.common.base.Preconditions.checkElementIndex;
+import static com.google.common.base.Preconditions.checkNotNull;
```
JLLeitschuh: This shouldn't need to be here.

PR #300, opened 2016-01-07
Speeds UP CI Builds

JLLeitschuh:
- Makes it so `:ui:assemble` only happens on deploy
  - Appveyor now completes in ~ 8 min instead of 12
  - Travis now completes in ~ 4 min instead of ~ 6 min
 - Moves JavaDoc generation to only happen on merge with master


JLLeitschuh:
Woot on getting the 300th issue!

JLLeitschuh:
@ThomasJClark Can you go and disable your Appveyor CI build of GRIP?

Issue #248, opened 2016-01-04
Deploying should stop the pipeline


JLLeitschuh:
This will prevent having to applications publishing network tables values to the same server.

PR #304, opened 2016-01-07
Created a "StopPipelineEvent" 

PaulaRudy:
to signal to the pipeline to stop executing when deployed. Currently, the event is posted from the DeployerController's "onDeploy" method, and the event is handled within the CameraSource file (this stops the camera, and if it encounters an error, logs it).

This will close issue #248


ui/src/main/java/edu/wpi/grip/ui/DeployerController.java
```diff
@@ -45,6 +48,8 @@
 
     private final StartStoppableButton.Factory startStopButtonFactory;
     private final DeploymentOptionsControllersFactory optionsControllersFactory;
+    private EventBus eventBus;
```
JLLeitschuh: This should be final

ui/src/main/java/edu/wpi/grip/ui/DeployerController.java
```diff
@@ -45,6 +48,8 @@
...
+    private final Logger logger = Logger.getLogger(DeployerController.class.getName());
```
JLLeitschuh: Remove the logger.

PR #307, opened 2016-01-08
Network tables is now only initialized when set as network protocol

### LINKED ISSUE #313: Network tables publishing does not flag exception when protocol is none
If the network protocol is `NONE` and there is a network tables publishing operation in the pipeline then the exception button should appear when you try to run the step. Otherwise, there is no indication to the user that the step isn't publishing.

JLLeitschuh:
If you have a NetworkTables operation in your pipeline and your
network protocol is not set correctly network tables operations
will now fail.

screenshot 2016-01-07 19 53 29

Closes #313

### DIFF
--- a/core/src/main/java/edu/wpi/grip/core/Main.java
+++ b/core/src/main/java/edu/wpi/grip/core/Main.java
@@ -27,6 +27,8 @@ public class Main {
     private EventBus eventBus;
     @Inject
     private Logger logger;
+    @Inject
+    private Operations operations;
 
     @SuppressWarnings("PMD.SystemPrintln")
     public static void main(String[] args) throws IOException, InterruptedException {
@@ -44,7 +46,7 @@ public class Main {
             logger.log(Level.INFO, "Loading file " + args[0]);
         }
 
-        Operations.addOperations(eventBus);
+        operations.addOperations(eventBus);
         CVOperations.addOperations(eventBus);
 
         final String projectPath = args[0];
--- a/core/src/main/java/edu/wpi/grip/core/operations/Operations.java
+++ b/core/src/main/java/edu/wpi/grip/core/operations/Operations.java
@@ -1,19 +1,27 @@
 package edu.wpi.grip.core.operations;
 
 import com.google.common.eventbus.EventBus;
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
 import edu.wpi.grip.core.events.OperationAddedEvent;
 import edu.wpi.grip.core.operations.composite.*;
+import edu.wpi.grip.core.operations.networktables.NTManager;
 import edu.wpi.grip.core.operations.networktables.NTPublishOperation;
 import edu.wpi.grip.core.operations.opencv.MatFieldAccessor;
 import edu.wpi.grip.core.operations.opencv.MinMaxLoc;
 import edu.wpi.grip.core.operations.opencv.NewPointOperation;
 import edu.wpi.grip.core.operations.opencv.NewSizeOperation;
 
-public final class Operations {
+@Singleton
+public class Operations {
 
-    private Operations() { /* no op */}
+    private final NTManager ntManager;
+    @Inject
+    Operations(NTManager ntManager) {
+        this.ntManager = ntManager;
+    }
 
-    public static void addOperations(EventBus eventBus) {
+    public void addOperations(EventBus eventBus) {
         // Add the default built-in operations to the palette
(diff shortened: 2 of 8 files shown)


ThomasJClark:
Does this really seem like the best way to handle this?  Is there a reason you can't just do a `ProjectSettingsChangedEvent` to make sure NetworkTables gets disabled?  If that doesn't work, you could even just do `NetworkTable::shutDown()` in the `NTPublishOperation` to negate the annoying auto-initialize thing

JLLeitschuh:
No, that won't work because there is no defined order of events when deserializing/initializing.
This is actually a safer, less hacky way to solve this problem.

ThomasJClark:
> No, that won't work because there is no defined order of events when deserializing/initializing.

Not true at all.  NetworkTables is only initialized when we call `getTable()` in the `NTPublishOperation` constructor, so calling `NetworkTables::shutdown()` immediately after cleanly stops it.  It doesn't matter at all the order that things are deserialized.

Just modify the `NTPublishOperation` like this:

```java
public NTPublishOperation(Class<T> type) {
    this.type = checkNotNull(type, "Type was null");

     // NetworkTables starts automatically when we create a table.  We don't want this behavior.
     table = NetworkTable.getTable("GRIP");
     NetworkTable.shutdown();
```

Much cleaner and more explicit, and now there aren't seemingly unnecessary changes in other classes that are secretly required to prevent this problem (such as injecting a `Provider<Pipeline>` instead of a `Pipeline` or using a `Supplier<ITable>` to provide a constant table)

JLLeitschuh:
Also, I really don't want an instance of Network Tables running in the tests. It could introduce a whole ton of instability.
This makes it impossible for network tables to unintentionally get started in a test.

ThomasJClark:
> Also, I really don't want an instance of Network Tables running in the tests. It could introduce a whole ton of instability.

Nope, again this is not the case.  NetworkTables is set to disabled *immediately* after it's enabled.  This is all before a single test runs.  It's entirely deterministic, there's no need for FUD about timing or concurrency or anything.

JLLeitschuh:
The `Operations` tests would enable network tables because they call perform.

ThomasJClark:
Nope, I've tried it.  Running the tests does not enable NetworkTables.  `NTPublishOperation::perform` would only called at all if we had a test that intentionally connects something to it.

JLLeitschuh:
To quote myself.
> Also, if network tables isn't the protocol defined then it would be unexpected and incorrect behavior to have the network tables step run successfully.

JLLeitschuh:
Do you disagree with this last statement?

ThomasJClark:
No, I think that seems reasonable.  You can also add a check for that regardless of how the initialization thing is solved.

JLLeitschuh:
How do you propose that I get the network tables state from within an operation???
There isn't a good way to do that.

ThomasJClark:
NetworkTables does that automatically and silently caches the variable until it's enabled.  We don't get a scary flashy icon like this, but tbh I think this is perfectly fine behavior.

JLLeitschuh:
Which variable?

ThomasJClark:
I think peter johnson said any variable we put in `Networktable::setValue()` is cached if NetworkTables isn't currently connected

ThomasJClark:
Are you gonna have time to update this tonight?  Because if not, I could take care if it right now.

JLLeitschuh:
I think this is a better way of fixing this. At least this way when network tables isn't publishing there is a UI element displaying why it isn't working.

If not that would cause confusion.

ThomasJClark:
The reason why I'm suggesting this is because there are a lot of code quality issues I have with this commit.  Instead of wasting a lot of time we don't have sorting them out, let's just use an easier fix that still completely solves the related issue.

JLLeitschuh:
Okay, then review the fix as is. I'll work on the other PR while you review this.

ThomasJClark:
Instead of me reviewing this and us spending the little time we have going back and forth, I'm opening a new pull request to resolve the NetworkTables log thing.  Let's not escalate this further than it has to go.

AustinShalit:
Code should only call a ```ProjectSettingsChangedEvent``` when the settings have changed.  


This code will still cause an exception to be reported to the console:
```
public NTPublishOperation(Class<T> type) {
    this.type = checkNotNull(type, "Type was null");

     // NetworkTables starts automatically when we create a table.  We don't want this behavior.
     table = NetworkTable.getTable("GRIP");
     NetworkTable.shutdown();
```

PR #308, opened 2016-01-08
Fix windows not shutting down when application stopped

### LINKED ISSUE #297: Process does not terminate when UI is closed
I just installed GRIP-0.3.0.exe on Windows 7. When I terminate the program, either by selecting File => Exit or by clicking on the red X in the upper-right corner of the window, the UI closes but the process does not terminate. I am forced to terminate the GRIP.exe process using Windows Task Manager.

JLLeitschuh:
Closes #297

### DIFF
--- a/ui/src/main/java/edu/wpi/grip/ui/Main.java
+++ b/ui/src/main/java/edu/wpi/grip/ui/Main.java
@@ -11,8 +11,8 @@ import edu.wpi.grip.core.events.UnexpectedThrowableEvent;
 import edu.wpi.grip.core.operations.Operations;
 import edu.wpi.grip.generated.CVOperations;
 import edu.wpi.grip.ui.util.DPIUtility;
-import edu.wpi.grip.ui.util.GRIPPlatform;
 import javafx.application.Application;
+import javafx.application.Platform;
 import javafx.fxml.FXMLLoader;
 import javafx.scene.Parent;
 import javafx.scene.Scene;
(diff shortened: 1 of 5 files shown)


ThomasJClark:
Cool, that was quick.  Could you explain in your PR comment what the problem turned out to be how this fixes it?

JLLeitschuh:
I'm not sure if it fixes it. I stopped having the hanging problem.
I basically threw everything I knew at trying to fix it.
 - Making sure that the uncaught thread handler didn't recursively call itself because of an exception.
 - Making sure that the thread will die if loggers fail to log.
 - Ensuring that if all else fails, if we get an error, then we try to log and leave as soon as possible.

I'm hoping someone with a windows machine can do an install and test it.

ThomasJClark:
ok

JLLeitschuh:
@ThomasJClark can we merge this and to a RC2 tonight to let people see if this works on windows?

ui/src/main/java/edu/wpi/grip/ui/Main.java
```diff
@@ -28,7 +30,9 @@
     @Inject
     private Palette palette;
```
JLLeitschuh: I don't know why the main file has the palette.
Whatever,

JLLeitschuh:
Can this be merged? (once I update)

ThomasJClark:
Is there any way we can get a build tested first?  I'd rather actually trying this first, but if there's no other option I guess we can merge and do a release.

JLLeitschuh:
I saw it on OSX but it was very, very, very rare.
Windows it happened every time.
I'm going to do a release. If you want anything else merged nows the time to do it.

PR #310, opened 2016-01-08
Makes deployer deploy to grip-core.jar

### LINKED ISSUE #305: Deployed GRIP jar should be called "grip.jar"
This makes it possible to run GRIP automatically from a C++ or Java robot program.  Currently, you'd have to know the exact version of GRIP that was deployed.

JLLeitschuh:
The deployer now coppies and renames the core grip jar to
`grip-core.jar` and SCPs that file over every time instead of
coping the original source code.

Closes #305

### DIFF
--- a/ui/src/main/java/edu/wpi/grip/ui/util/deployment/DeployedInstanceManager.java
+++ b/ui/src/main/java/edu/wpi/grip/ui/util/deployment/DeployedInstanceManager.java
@@ -25,6 +25,7 @@ import java.io.OutputStream;
 import java.net.InetAddress;
 import java.net.URISyntaxException;
 import java.net.URLDecoder;
+import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.Optional;
 import java.util.function.Supplier;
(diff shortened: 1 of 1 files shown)


JLLeitschuh:
Yes, I did test this on gears bot. It works.

ThomasJClark:
Why would we make another local copy of the jar?  We just want the copy on the roboRIO to be named grip.jar, not the local one

ThomasJClark:
Also, it should probably be grip.jar and not grip-core.jar

JLLeitschuh:
I can't do that with the ant SCP tool.
I have to copy it. Unless you have some other way you know how to SCP a file while changing its name.

JLLeitschuh:
By the way, I tried using some of these other params: https://ant.apache.org/manual/Tasks/scp.html
Didn't work.

ThomasJClark:
`localTofile` does't work?

JLLeitschuh:
No, throws a builder exception

ThomasJClark:
Ok, then `mv` the file with an ssh command after uploading it

JLLeitschuh:
Why? The temporary directory automatically gets cleaned up by java when we're done with it.
I've seen the `mv` command block before and never return on my computer.

Why is this any worse?

ThomasJClark:
Because you're using up extra space on the user's computer to store a file that's already on there.  It's probably not *that* much space, but it just seems like an incredibly hacky way to change an SCP target.

If you've really done every thing you can to investigate how to do this for real, I guess we can accept this workaround as a necessary evil.  Make sure you change the "-core" thing, though

JLLeitschuh:
Okay, fixed to remove the `-core`

PR #311, opened 2016-01-08
Don't automatically start up NetworkTables

ThomasJClark:
NetworkTables initializes itself automatically as soon as we create a table, which we do basically as soon as GRIP starts up.  This is kind of annoying because it creates a lot of noise in the logs when we're not actually using NetworkTables.

We can simply undo this by immediately calling `NetworkTables::shutdown()`, and it will only start back up again if we explicitly set the project settings to use NetworkTables.

### DIFF
--- a/core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
+++ b/core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
@@ -28,8 +28,11 @@ public class NTPublishOperation<T extends NTPublishable> implements Operation {
     private final List<Method> ntValueMethods = new ArrayList<>();
 
     public NTPublishOperation(Class<T> type) {
-        this.table = NetworkTable.getTable("GRIP");
         this.type = checkNotNull(type, "Type was null");
+
+        // NetworkTables automatically initializes itself when we get a table.  We don't want this behavior.
+        table = NetworkTable.getTable("GRIP");
+        NetworkTable.shutdown();
 
         // Any accessor method with an @NTValue annotation can be published to NetworkTables.
         for (Method method : type.getDeclaredMethods()) {


core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
```diff
@@ -28,9 +28,12 @@
     private final List<Method> ntValueMethods = new ArrayList<>();
 
     public NTPublishOperation(Class<T> type) {
-        this.table = NetworkTable.getTable("GRIP");
         this.type = checkNotNull(type, "Type was null");
 
+        // NetworkTables automatically initializes itself when we get a table.  We don't want this behavior.
+        table = NetworkTable.getTable("GRIP");
+        NetworkTable.shutdown();
```
JLLeitschuh: Try loading a project from a save file with network tables specified as network protocol.
Does it restore it running?

ThomasJClark: Yes.  When you load a project, a `ProjectSettingsChangedEvent` is sent, and `NTManager` hears this and calls `NetworkTables::initialize()`.

JLLeitschuh: But I don't get an error saying that network tables isn't working now though.
The step should inform me that its invalid.

This is what it looks like when I have a network protocol of NONE set.
There should be a warning on the step.
Not showing me a problem

This is an invalid fix.

JLLeitschuh:
Are you coming up with another solution or can I close this PR? It seems like its solving the problem in a way that won't be useful to a user.

core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
ThomasJClark: This is an "invalid fix"?  Did we even have that label before, or did you create it especially for this PR?

Anyway, the problem was that NetworkTables was logging messages even when disabled.  You personally complained about this a lot.  This change fixes that issue.  If you have opinions about changes we could possibly make to the user interface, feel free to open an issue for them.

ThomasJClark:
> Are you coming up with another solution or can I close this PR? It seems like its solving the problem in a way that won't be useful to a user.

No, this is solving a problem for the developer.  It prevents log files full of networktables messages, so we can debug easier and also help users with problems.  You were there for all of the discussions where we talked about this problem, and even if you weren't, you could have just read the PR description.

JLLeitschuh:
#307 is a better solution to the problem. Why not merge the solution that fixes two bugs with one stone?
Why do you seem so set on doing it your way when there are clear benefits to what I did?

ThomasJClark:
#307 is not a better solution to the problem.  It changes a lot of unrelated code in a fragile way to influence the order that things happen.  If someone, for example, didn't realize why a parameter for a constant was a `Provider` or `Supplier` and changed it back, it would break in an unexpected way.

This is a simple problem - NetworkTables automatically initializes when we call `getTable()`.  It can be solved a a simple way.

JLLeitschuh:
It does not fix #313

ThomasJClark:
We can deal with #313 later.  This PR has literally nothing to do with #313.

JLLeitschuh:
Okay, good, I fixed #313
This one will conflict with that PR.
Please merge the one that fixes #313 as it is a bug and should be fixed before kickoff.

JLLeitschuh:
1. This is a duplicate but doesn't do as much.
2. It doesn't solve an issue that is fairly straightforward to fix in #313

ThomasJClark:
This PR solves one problem and is ready to merge.

#307 attempts to solve the same problem and another entirely  unrelated one that we don't have a consensus on how to solve and is not ready to merge.

We should merge *this* PR,then figure out how to fix #313, which I think is is a very low priority issue.

I'd really rather not stoop to my own merge ultimatum, but we seriously need to stop wasting time on silly discussions like this.  Please either merge this pull request or tell me how it doesn't solve the NetworkTables logging problem, and don't waste any more time commenting here about the unrelated  #313 issue.

ThomasJClark:
Alright, the networktables logs are really starting to bother me as I work on fixing other bugs.  This is getting merged before the thing tonight unless there's a really good objection

PR #312, opened 2016-01-08
Perserve exception if copy fails

JLLeitschuh:
Thanks to PMD for catching this one


JLLeitschuh:
This is so simple of a change I'm just going to merge it.
There is absolutely no reason why anyone would have a problem with it.

PR #315, opened 2016-01-08
Replace Existing when copying to tempdir for deploy

JLLeitschuh:


JLLeitschuh:
This should be a really simple fix.
I'm going to merge it as soon as it passes so I can get this release candidate out the door.

Issue #252, opened 2016-01-04
Operations feed back to previous step


JLLeitschuh:
I loaded this this save file:
```xml
<grip:Pipeline>
  <sources>
    <grip:Camera>
      <property name="deviceNumber" value="0"/>
    </grip:Camera>
  </sources>
  <steps>
    <grip:Step name="resize">
      <grip:Input step="0" socket="0"/>
      <grip:Input step="0" socket="1"/>
      <grip:Input step="0" socket="2">
        <value>0.3</value>
      </grip:Input>
      <grip:Input step="0" socket="3">
        <value>0.3</value>
      </grip:Input>
      <grip:Input step="0" socket="4">
        <value>INTER_LINEAR</value>
      </grip:Input>
      <grip:Output step="0" socket="0" previewed="true"/>
    </grip:Step>
    <grip:Step name="Desaturate">
      <grip:Input step="1" socket="0"/>
      <grip:Output step="1" socket="0" previewed="true"/>
    </grip:Step>
    <grip:Step name="Find Lines">
      <grip:Input step="2" socket="0"/>
      <grip:Output step="2" socket="0" previewed="true"/>
    </grip:Step>
  </steps>
  <connections>
    <grip:Connection>
      <grip:Output step="0" socket="0" previewed="true"/>
      <grip:Input step="1" socket="0"/>
    </grip:Connection>
    <grip:Connection>
      <grip:Output step="1" socket="0" previewed="true"/>
      <grip:Input step="2" socket="0"/>
    </grip:Connection>
    <grip:Connection>
      <grip:Output source="0" socket="0" previewed="false"/>
      <grip:Input step="0" socket="0"/>
    </grip:Connection>
  </connections>
</grip:Pipeline>
```
And the pipeline looked like this:
![Pipeline Looks Wrong](https://dl.dropbox.com/s/rtld3lxv4y8b2ul/Screenshot%202016-01-04%2015.29.32.png)

Notice:
 - The steps don't have their previewed buttons toggled #253 
 - The first step that is just a simple resize is actually being desaturated as well (this issue)

The pipeline should look like this:
![How the pipeline should look](https://dl.dropbox.com/s/tdb6esdkg1pcrj4/Screenshot%202016-01-04%2015.31.06.png)

PaulaRudy:
Seems to be associated specifically with the desaturate step. The preview windows being open is an artifact of them being open at the time the file was saved. Here's the output of the spectacular error I managed to trrigger
[...]

PR #319, opened 2016-01-08
Implemented a workaround to fix issue # 252.

PaulaRudy:
Now all save files will save with the previews closed. This fixes 252 by ensuring the race condition caused when desaturate and another operation's previews are opened at the same time.

Closes issue #252

### DIFF
--- a/core/src/main/java/edu/wpi/grip/core/serialization/Project.java
+++ b/core/src/main/java/edu/wpi/grip/core/serialization/Project.java
@@ -76,7 +76,22 @@ public class Project {
         this.file = Optional.of(file);
     }
 
-    public void save(Writer writer) {
+    public void save(Writer writer) throws IOException {
         this.xstream.toXML(this.pipeline, writer);
+
+        //The following is a work-around to make sure the previews are all recorded as closed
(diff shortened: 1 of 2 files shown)


JLLeitschuh:
@ThomasJClark you did serialization, I'm going to leave this one to you to verify.

JLLeitschuh:
Actually, @PaulaRudy couldn't you just change line 121 of the `Socket Converter` so that set previewed was always set to false?

ThomasJClark:
Yeah, this is not okay. Do what @JLLeitschuh said,  and also open an issue to find the underlying problem.

ThomasJClark:
Also, I think there are tests for this functionality. If we decide to intentionally disable it for now, we should disable the tests also.

PaulaRudy:
So apparently this doesn't actually prevent the error all the time- I've discovered that when GRIP is slowed down (for whatever reason)  it's possible to trigger the bug by manually pressing the preview buttons. I have the changes y'all have prescribed implemented on my local branch, but I haven't pushed yet because I'm trying to see if I can fix the issue entirely. Do you think this is a good idea to implement in the mean time? It won't fix things, but it will help prevent them.

PR #322, opened 2016-01-08
Fix connections not positioning correctly when loading project

### LINKED ISSUE #321: Connections are positioned weirdly
In the dependency injection refactor, for some reason the transform invalidation and synchronization stuff that keeps the connections correctly positions was rewritten.  As a result of this, connections are incorrectly positioned when a new project is opened, as shown below.  In addition, connections flicker when they move, since that's done using `Platform.runLater` and some custom functions instead of `PlatformImpl.runAndWait`.

s

ThomasJClark:
Also connections flickering when moving.

Closes #321

Before:
before

After:
after

### DIFF
--- a/ui/src/main/java/edu/wpi/grip/ui/pipeline/PipelineController.java
+++ b/ui/src/main/java/edu/wpi/grip/ui/pipeline/PipelineController.java
@@ -3,15 +3,14 @@ package edu.wpi.grip.ui.pipeline;
 import com.google.common.eventbus.EventBus;
 import com.google.common.eventbus.Subscribe;
 import com.google.inject.Singleton;
+import com.sun.javafx.application.PlatformImpl;
 import edu.wpi.grip.core.*;
 import edu.wpi.grip.core.events.*;
 import edu.wpi.grip.ui.annotations.ParametrizedController;
 import edu.wpi.grip.ui.pipeline.input.InputSocketController;
 import edu.wpi.grip.ui.pipeline.source.SourceController;
 import edu.wpi.grip.ui.pipeline.source.SourceControllerFactory;
-import edu.wpi.grip.ui.util.GRIPPlatform;
 import edu.wpi.grip.ui.util.ControllerMap;
-import javafx.application.Platform;
 import javafx.beans.InvalidationListener;
 import javafx.beans.property.ReadOnlyObjectProperty;
 import javafx.collections.ObservableList;
@@ -56,8 +55,6 @@ public final class PipelineController {
     private StepController.Factory stepControllerFactory;
     @Inject
     private AddSourceView addSourceView;
-    @Inject
-    private GRIPPlatform platform;
 
     private ControllerMap<StepController, Node> stepsMapManager;
     private ControllerMap<SourceController, Node> sourceMapManager;
@@ -182,7 +179,7 @@ public final class PipelineController {
      * details of adding the connection.
      */
     private void addConnectionView(Connection connection) {
-        platform.runAsSoonAsPossible(() -> {
+        PlatformImpl.runAndWait(() -> {
             // Before adding a connection control, we have to look up the controls for both sockets in the connection so
             // we know where to position it.
             final OutputSocketController outputSocketView = findOutputSocketView(connection.getOutputSocket());
@@ -207,9 +204,7 @@ public final class PipelineController {
                     final double x2 = inputSocketBounds.getMinX() + inputSocketBounds.getWidth() / 2.0;
                     final double y2 = inputSocketBounds.getMinY() + inputSocketBounds.getHeight() / 2.0;
 
-                    // This can run whenever. Don't wait for it to complete.
-                    // This should be Platform.runLater
-                    Platform.runLater(() -> {
+                    PlatformImpl.runAndWait(() -> {
                         connectionView.inputHandleProperty().setValue(new Point2D(x1, y1));
                         connectionView.outputHandleProperty().setValue(new Point2D(x2, y2));
                         ((ReadOnlyObjectProperty) observable).get();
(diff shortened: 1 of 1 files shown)


ui/src/main/java/edu/wpi/grip/ui/pipeline/PipelineController.java
```diff
@@ -227,7 +222,7 @@ private void addConnectionView(Connection connection) {
 
     @Subscribe
     public void onSourceAdded(SourceAddedEvent event) {
-        platform.runAsSoonAsPossible(() -> {
+        PlatformImpl.runAndWait(() -> {
```
JLLeitschuh: Deadlock prone.

ui/src/main/java/edu/wpi/grip/ui/pipeline/PipelineController.java
```diff
@@ -235,7 +230,7 @@ public void onSourceAdded(SourceAddedEvent event) {
 
     @Subscribe
     public void onSourceRemoved(SourceRemovedEvent event) {
-        platform.runAsSoonAsPossible(() -> {
+        PlatformImpl.runAndWait(() -> {
```
JLLeitschuh: Deadlock prone

ui/src/main/java/edu/wpi/grip/ui/pipeline/PipelineController.java
```diff
@@ -245,7 +240,7 @@ public void onSourceRemoved(SourceRemovedEvent event) {
     @Subscribe
     public void onStepAdded(StepAddedEvent event) {
         // Add a new view to the pipeline for the step that was added
-        platform.runAsSoonAsPossible(() -> {
+        PlatformImpl.runAndWait(() -> {
```
JLLeitschuh: Deadlock prone

ui/src/main/java/edu/wpi/grip/ui/pipeline/PipelineController.java
```diff
@@ -259,7 +254,7 @@ public void onStepAdded(StepAddedEvent event) {
     @Subscribe
     public void onStepRemoved(StepRemovedEvent event) {
         // Remove the control that corresponds with the step that was removed
-        platform.runAsSoonAsPossible(() -> {
+        PlatformImpl.runAndWait(() -> {
```
JLLeitschuh: Deadlock prone.

ui/src/main/java/edu/wpi/grip/ui/pipeline/PipelineController.java
```diff
@@ -269,7 +264,7 @@ public void onStepRemoved(StepRemovedEvent event) {
 
     @Subscribe
     public void onStepMoved(StepMovedEvent event) {
-        platform.runAsSoonAsPossible(() -> {
+        PlatformImpl.runAndWait(() -> {
```
JLLeitschuh: Deadlock prone

ui/src/main/java/edu/wpi/grip/ui/pipeline/PipelineController.java
```diff
@@ -278,13 +273,13 @@ public void onStepMoved(StepMovedEvent event) {
...
         // Add the new connection view
-        platform.runAsSoonAsPossible(() -> addConnectionView(event.getConnection()));
+        addConnectionView(event.getConnection());
     }
 
     @Subscribe
     public void onConnectionRemoved(ConnectionRemovedEvent event) {
         // Remove the control that corresponds with the connection that was removed
-        platform.runAsSoonAsPossible(() -> {
+        PlatformImpl.runAndWait(() -> {
```
JLLeitschuh: Deadlock prone

ThomasJClark: I don't think this should deadlock, and believe I've read all the relevant literature on `runAndWait()`.  Any evidence to back this up?

JLLeitschuh: Yes, that's why I created `runAsSoonAsPossible`.
I had major deadlocking problems when writing dependency injection. The only 100% deadlock free way of doing this is, using `Platform.runLater`.

ThomasJClark: Ok, but I really don't believe this code deadlocks.  Do you have evidence that it does or not?

JLLeitschuh: I don't have evidence that it doesn't work in this case no.
What I do know is that `runAndWait` can deadlock the javafx thread on the event bus because another thread is waiting on the javafx thread to run and the javafx thread is trying to run an event.

Basically, I don't trust `runAndWait` at all. I think it is a nasty hack around a problem that is a simple drawing problem. Try invalidating the view and triggering a redraw instead.

ThomasJClark: We *want* it to wait on the JavaFX thread, otherwise there's flickering.

> Basically, I don't trust runAndWait at all. 

I think that's a bit unfair. `runAndWait` would only deadlock if we post an event inside of it. It's totally appropriate to use for cases where we're only manipulating JavaFX components (not grip objects) from an arbitrary thread and we want it to complete atomically.

There's nothing untrustworthy about it, it just doesn't interact well with the unusual behavior of `eventBus.post()`.

Trust me, I've tested this code in every possible situation, and I've spent some time reasoning about how this needs to work. Doing it this way removes some annoying flickering, and I think that will have a positive impact on users trying this for the first time.

JLLeitschuh: Okay, can you leave comments on the lines saying that this is intentional, and that it won't break anything. And also include that there should be no events posted inside of the blocks of code.

PR #326, opened 2016-01-09
Revert "Don't automatically start up NetworkTables"

JLLeitschuh:
Reverts WPIRoboticsProjects/GRIP#311
This is an invalid fix.
This should not be merged this way.
I don't care if it bothered you. You do not get to make the executive decision on how this is fixed..

### DIFF
--- a/core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
+++ b/core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
@@ -28,11 +28,8 @@ public class NTPublishOperation<T extends NTPublishable> implements Operation {
     private final List<Method> ntValueMethods = new ArrayList<>();
 
     public NTPublishOperation(Class<T> type) {
+        this.table = NetworkTable.getTable("GRIP");
         this.type = checkNotNull(type, "Type was null");
-
-        // NetworkTables automatically initializes itself when we get a table.  We don't want this behavior.
-        table = NetworkTable.getTable("GRIP");
-        NetworkTable.shutdown();
 
         // Any accessor method with an @NTValue annotation can be published to NetworkTables.
         for (Method method : type.getDeclaredMethods()) {


ThomasJClark:
Alright, as long as you also no longer have the privilege to make a merge ultimatum.

JLLeitschuh:
I do when we have a deadline and you don't review (my) code for over 2 days and then try to get (your) code merged 20 min before a deadline.

ThomasJClark:
Ok, if you think you should sometimes have permission to make merge ultimatums, we should come up some rules for when it's appropriate before just doing it.

JLLeitschuh:
Sure, that seems fair. Hopefully we won't have another deadline like the one last week.
But I agree we should have a plan in place for situations like this in the future.

ThomasJClark:
Again, we didn't have a deadline last week.  We just had a feature freeze.

JLLeitschuh:
> A (complete) feature freeze, in which all work on adding new features is suspended, shifting the effort towards fixing bugs and improving the user experience. The addition of new features may have a disruptive effect on other parts of the program, due both to the introduction of new, untested source code or resources and to interactions with other features; thus, a feature freeze helps improve the program's stability.
For example: "user interface feature freeze" means no more features will be permitted to the user interface portion of the code; bugs can still be fixed.
> \- [Wikipedia](https://en.wikipedia.org/wiki/Freeze_(software_engineering))

ThomasJClark:
Cool, so we're on the same page about feature freezes now?  It just means we don't come up with new features, but we can fix the ones we're currently working on.

The section after this quote brings up a "code freeze", which I think might be what you were confusing that with.

JLLeitschuh:
No, it means they aren't being merged in with master.
Tom, stop trying to be slippery and justify your not having your work done on time.
We will discuss this when we get back to WPI.

ThomasJClark:
> No, it means they aren't being merged in with master.

It doesn't say that in the Wikipedia article you quoted

PR #331, opened 2016-01-09
Exception alert enhancements

### LINKED ISSUE #330: "Open GitHub Issues" sometimes crashes web browser or GitHub servers
This is caused by the ridiculously long URLs we give it to auto-fill the issue text.

This issue alone is mostly a somewhat ironic minor inconvenience, but I marked it as HIGH PRIORITY because it prevents us from finding out about new bugs if users aren't eager enough to manually find our issues tracker.

ThomasJClark:
GitHub's servers (and many modern browsers) apparently don't like us redirecting users to thousand-character-long URLs to automatically fill in issue forms.  So, I got rid of that whole thing and just put the generated text in the exception alert.  Users can copy and paste it themselves.

The list of system properties was also updated to remove some redundant/useless ones.  We can always ask for more info if necessary, but the ones removed can all be derived from existing ones.  This should just clear up some visual clutter.

The lack of styling in the alert has also bothered me for a while, so that's fixed now.

Closes #330

### DIFF
--- a/ui/src/main/java/edu/wpi/grip/ui/ExceptionAlert.java
+++ b/ui/src/main/java/edu/wpi/grip/ui/ExceptionAlert.java
@@ -32,7 +32,7 @@ public final class ExceptionAlert extends Alert {
     };
 
     private static final ButtonType COPY = new ButtonType("Copy to Clipboard");
-    private static final ButtonType GITHUB = new ButtonType("Open GitHub Issues");
+    private static final ButtonType REPORT = new ButtonType("Report Issue");
 
     private final Node initialFocusElement;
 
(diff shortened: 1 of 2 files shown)


JLLeitschuh:
Sure, I really don't like how the monospace font looks, I also don't like how the line wraps, the stack trace error box should have both a horizontal and vertical scroll bar.

What it looks like on a mac

ThomasJClark:
A horizontal scroll bar is a good idea. I'll do that when I get home today.

JLLeitschuh:
Cool!

JLLeitschuh:
Also, can you keep the copy to clipboard button? I liked it. It was convenient.

ui/src/main/java/edu/wpi/grip/ui/ExceptionAlert.java
```diff
@@ -20,39 +18,19 @@
...
 
-    private final String exceptionMessage;
-    private final String systemInfoMessage;
-    private final String additionalInfoMessage;
-    private final String message;
-    private final Throwable initialCause;
+    private static final ButtonType GITHUB = new ButtonType("Open GitHub Issues");
 
-    private final ButtonType openGitHubIssuesBtnType = new ButtonType("Open GitHub Issues");
-    private final ButtonType copyToClipboardBtnType = new ButtonType("Copy To Clipboard");
```
JLLeitschuh: Don't delete this. I like this button. Don't remove it. 10/10

JLLeitschuh:
This is what it looks like on mac

PR #359, opened 2016-01-14
Adds save dirty checking event

JLLeitschuh:
This provides a cleaner UI experience.
You are now only requested to save when an event dirties the save file.
The deployer now also ensures that you have not dirtied the save.

### DIFF
--- a/core/src/main/java/edu/wpi/grip/core/events/ConnectionAddedEvent.java
+++ b/core/src/main/java/edu/wpi/grip/core/events/ConnectionAddedEvent.java
@@ -9,7 +9,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
  * An event that occurs when a new connection is added to the pipeline.  This is triggered by the user adding a
  * connection with the GUI.
  */
-public class ConnectionAddedEvent implements RunPipelineEvent {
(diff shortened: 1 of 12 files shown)


ThomasJClark:
IMO we should implement undo/redo instead.  With undo/redo implemented, determining if the project has changed should be easy

JLLeitschuh:
> This is really hard to get right.
> @bradamiller 

I spent a few hours on it before. I agree that it's a feature that could be added. The events that are fired when an undo/redo event happens are the same events that this deals with so I'm not sure if the two are mutually exclusive.

JacisNonsense:
I think Dirty saves is a good feature to implement until we find a suitable way to go about undo/redo operations. I think both these features should be implemented in the near future.

JLLeitschuh:
I agree with what @JacisNonsense said

AustinShalit:
Before merging can we add a UI indicator that the pipeline is dirty? Many apps add an asterisk after the file name in the title bar to indicate the file is dirty.

JLLeitschuh:
@AustinShalit How's this coming?

AustinShalit:
#647 completes this feature.

PR #369, opened 2016-01-15
Add a "Publish Video" operation

### LINKED ISSUE #355: Http jpeg stream to smart dashboard step
We should have a step to stream a mat to smartdashboard.
This is a request from @bradamiller
@ThomasJClark you probably understand how to do this better than I do.

ThomasJClark:
This allows you to view the video feed with any compatible FRC dashboard
(tested with SmartDashboard, but the FRC dashboard uses the same
protocol when it's in "hardware" mode).  Intermediate steps can also be
published, allowing teams to visualize parts of the pipeline even when
it's running in headless mode.

screenshot

Closes #355


core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
```diff
@@ -0,0 +1,158 @@
...
+            ServerSocket serverSocket = null;
+            Socket socket = null;
+
+            try {
+                serverSocket = new ServerSocket(PORT);
+//                serverSocket.setReuseAddress(true);
+//                serverSocket.bind(new InetSocketAddress(PORT));
+                logger.info("Starting camera server");
+
+                while (!serverSocket.isClosed()) {
```
JLLeitschuh: Check that the thread isn't interrupted.

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
```diff
@@ -0,0 +1,158 @@
...
+                    socket = serverSocket.accept();
+                    DataOutputStream socketStream = new DataOutputStream(socket.getOutputStream());
+                    logger.info("Got connection from " + socket.getInetAddress());
+
+                    while (!socket.isClosed()) {
```
JLLeitschuh: Check that the thread isn't interrupted.

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
ThomasJClark: k

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
ThomasJClark: k

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
```diff
@@ -0,0 +1,157 @@
...
+//                serverSocket.setReuseAddress(true);
+//                serverSocket.bind(new InetSocketAddress(PORT));
+                logger.info("Starting camera server");
+
+                while (!serverSocket.isClosed() && !Thread.interrupted()) {
+                    socket = serverSocket.accept();
+                    DataOutputStream socketStream = new DataOutputStream(socket.getOutputStream());
+                    logger.info("Got connection from " + socket.getInetAddress());
+
+                    while (!socket.isClosed() && !Thread.interrupted()) {
```
JLLeitschuh: `Thread.interrupted()` resets the flag to false.
You need to use `!Thread.currentThread().isInterrupted()` so the flag isn't reset.
Otherwise the outer while loop will not be exited.

JLLeitschuh:
What happens if you have two of these operations in a pipeline? Will they both work?

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
ThomasJClark: k

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
```diff
@@ -0,0 +1,185 @@
...
+    private transient volatile float compressionQuality = 0.5f;
+    private transient volatile boolean connected = false;
+
+    /**
+     * Listens for incoming connections on port 1180 and writes JPEG data whenever there's a new frame.
+     */
+    private final Runnable runServer = () -> {
+        // Loop forever (or at least until the thread is interrupted).  This lets us recover from the dashboard
+        // disconnecting or the network connection going away temporarily.
+        while (true) {
```
JLLeitschuh: `!Thread.currentThread().isInterrupted()`

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
```diff
@@ -0,0 +1,189 @@
...
+                        // The FRC dashboard image protocol consists of a magic number, the size of the image data,
+                        // and the image data itself.
+                        socketStream.write(MAGIC_NUMBER);
+                        socketStream.writeInt(baos.size());
+                        baos.writeTo(socketStream);
+                    }
+                }
+            } catch (IOException e) {
+                logger.log(Level.WARNING, e.getMessage(), e);
+            } catch (InterruptedException e) {
```
JLLeitschuh: You must set the thread's interrupted flag here. The default thread handler depends on it being set.

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
```diff
@@ -0,0 +1,204 @@
...
+            this.compressionQuality = quality.floatValue();
+            imageLock.notify();
+        }
+    }
+
+    @Subscribe
+    public void onStepRemoved(StepRemovedEvent event) {
+        if (event.getStep().getOperation() == this) {
+            if (--numSteps == 0) {
+                serverThread.ifPresent(Thread::interrupt);
```
JLLeitschuh: Make sure that this thread joins correctly. You can do this in a separate thread if you want to keep this from hanging.

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
```diff
@@ -0,0 +1,196 @@
...
+    public OutputSocket<?>[] createOutputSockets(EventBus eventBus) {
+        return new OutputSocket<?>[0];
+    }
+
+    @Override
+    public Optional<?> createData() {
+        numSteps++;
+        if (!serverThread.isPresent()) {
+            serverThread = Optional.of(new Thread(runServer, "Camera Server"));
+            serverThread.get().start();
```
JLLeitschuh: This needs to be a daemon thread.

PR #372, opened 2016-01-16
Add a resize operation

ThomasJClark:


core/src/main/java/edu/wpi/grip/core/operations/Operations.java
```diff
@@ -15,6 +15,7 @@ private Operations() { /* no op */}
 
     public static void addOperations(EventBus eventBus) {
         // Add the default built-in operations to the palette
+        eventBus.post(new OperationAddedEvent(new ResizeOperation()));
```
AustinShalit: What is the order of the operations?

AustinShalit:
How is this operation different from the Open CV Resize operation?  Do we want to document the differences?

ThomasJClark:
The generated OpenCV one:

- Requires you to use a "new size" operation to specify the width and hight
- Has two different ways of resizing (using a size, or with an x and y scale). It's really unintuitive how it ignores one if the other is specified.
- Has unfriendly socket names
- Has unfriendly interpolation mode names ("INTER_LINEAR" instead of "Linear")
- Has a bunch of "interpolation modes" that don't work and actually just make the operation crash (like "FILL_OUTLIERS", which isn't actually an interpolation mode in OpenCV)

AustinShalit:
@ThomasJClark Should the OpenCV operation be removed if this PR is merged?

ThomasJClark:
Probably eventually, but removing it would break compatibility with existing projects that use it

AustinShalit:
A solution to that would be removing it from the operation palate but leaving it as a valid operation.

core/src/main/java/edu/wpi/grip/core/operations/Operations.java
JLLeitschuh: What do you mean?

AustinShalit: What is the sort order of the operations? It should have some set order so people can expect where the operations are in the list.

JLLeitschuh: We don't have a sort order the order is defined by the order that we list them in this file and the generated operations file.

core/src/main/java/edu/wpi/grip/core/operations/composite/ResizeOperation.java
```diff
@@ -0,0 +1,92 @@
...
+
+    private enum Interpolation {
+        NEAREST("None", INTER_NEAREST),
+        LINEAR("Linear", INTER_LINEAR),
+        CUBIC("Cubic", INTER_CUBIC),
+        LANCZOS("Lanczos", INTER_LANCZOS4),
+        AREA("Area", INTER_AREA);
+
+        String label;
+        int value;
```
JLLeitschuh: These should both be final variables.

ThomasJClark: k

core/src/main/java/edu/wpi/grip/core/operations/Operations.java
JLLeitschuh: To rephrase @AustinShalit's question.
We should have some sort of policy for ordering and defining how we want the palette ordered.
Why we are putting specific operations in a specified order.

Do we have a policy like this? Can we come up with one?

Issue #316, opened 2016-01-08
Create complex pipeline file to test against compatibility breaking


JLLeitschuh:
We need to create a grip save file with almost everything inside of it (except for perhaps camera sources) that we can add to the test system. This file will ensure that we don't break compatibility during the season.

PaulaRudy:
I can take a stab at this.

PaulaRudy:
Did we need all the operations chained together? Or will just giving them enough appropriate input to execute be enough?

JLLeitschuh:
Uhhh. I would connect a few of them, sure. You don't need to connect all of them though no.

AustinShalit:
I am going to hop on this

PaulaRudy:
So! Here is the file:

```
<grip:Pipeline>
  <sources>
    <grip:ImageFile>
      <property name="path" value="REPLACEME"/>
    </grip:ImageFile>
    <grip:Camera>
      <property name="deviceNumber" value="0"/>
    </grip:Camera>
  </sources>
  <steps>
    <grip:Step name="Resize Image">
      <grip:Input step="0" socket="0"/>
      <grip:Input step="0" socket="1">
        <value>200.0</value>
      </grip:Input>
      <grip:Input step="0" socket="2">
        <value>100.0</value>
      </grip:Input>
      <grip:Input step="0" socket="3">
        <value>CUBIC</value>
      </grip:Input>
      <grip:Output step="0" socket="0" previewed="true"/>
    </grip:Step>
    <grip:Step name="Blur">
      <grip:Input step="1" socket="0"/>
      <grip:Input step="1" socket="1">
        <value>BOX</value>
      </grip:Input>
      <grip:Input step="1" socket="2">
        <value>46.846846846846844</value>
      </grip:Input>
      <grip:Output step="1" socket="0" previewed="true"/>
    </grip:Step>
    <grip:Step name="Desaturate">
      <grip:Input step="2" socket="0"/>
      <grip:Output step="2" socket="0" previewed="true"/>
    </grip:Step>
    <grip:Step name="RGB Threshold">
      <grip:Input step="3" socket="0"/>
      <grip:Input step="3" socket="1">
        <value>
          <double>82.55395683453237</double>
          <double>255.0</double>
        </value>
      </grip:Input>
      <grip:Input step="3" socket="2">
        <value>
          <double>41.0402645622285</double>
          <double>157.3718868704939</double>
        </value>
      </grip:Input>
[...]

PR #378, opened 2016-01-16
Add solidity filter

ThomasJClark:
This lets you filter contours by their "solidity", which is the ratio of their area to the area of their convex hull. 

You can technically filter out false positives by just having a minimum area, but that means that actual targets will be filtered out if you're too far away from them.  In addition, larger false positives will still show up.

The FIRST Stronghold targets have a solidity of about 1/3 regardless of how far away they are, so using a solidity filter gets them very reliably.

demo

### DIFF
--- a/core/src/main/java/edu/wpi/grip/core/operations/composite/ContoursReport.java
+++ b/core/src/main/java/edu/wpi/grip/core/operations/composite/ContoursReport.java
@@ -6,10 +6,8 @@ import edu.wpi.grip.core.operations.networktables.NTValue;
 
 import java.util.Optional;
 
-import static org.bytedeco.javacpp.opencv_core.MatVector;
-import static org.bytedeco.javacpp.opencv_core.Rect;
-import static org.bytedeco.javacpp.opencv_imgproc.boundingRect;
-import static org.bytedeco.javacpp.opencv_imgproc.contourArea;
+import static org.bytedeco.javacpp.opencv_core.*;
+import static org.bytedeco.javacpp.opencv_imgproc.*;
 
 /**
  * The output of {@link FindContoursOperation}.  This stores a list of contours (which is basically a list of points) in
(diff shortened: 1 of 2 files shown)


AustinShalit:
Is there a way we can make the text below the slider display a double instead of an int?

untitled

ThomasJClark:
Good question.  I might just make it a percentage from 0 to 100, just so we don't have to deal with figuring how how many digits to round the label to

AustinShalit:
That sounds like a better idea.

AustinShalit:
Do we want this code to handle NaN or should teams handle that?

ThomasJClark:
I actually didn't realize that convex hulls could have 0 area, but I guess that's possible if it's a single pixel. I think that teams will usually have a minimum area filter of at least one, so it should be fine.

JLLeitschuh:
Won't this break save files?
This socket won't exist in older save files which will cause an exception to be thrown.

ThomasJClark:
Nope, it will just use the default value

JLLeitschuh:
You tested this?
I'd really like to see #316 finished and merged before we go messing with existing operations.
@PaulaRudy What is your timeline on getting that finished?

ThomasJClark:
> You tested this?

Yes

> I'd really like to see #316 finished and merged before we go messing with existing operations.

Ideally I guess we would have something like that, but I don't think we need to stop adding new features until then

JLLeitschuh:
I aggree. Feel free to add features, we shouldn't merge them until #316 is finished.
#316 Should be done before we do a new release anyways so this can wait for that.

ThomasJClark:
> I aggree. Feel free to add features, we shouldn't merge them until #316 is finished.
> #316 Should be done before we do a new release anyways so this can wait for that.

I disagree with waiting on #316 to do a release.

JLLeitschuh:
Okay, sure. You can disagree. I'm not merging this yet though.
We can bring this up until the next meeting. Until then, I don't want to risk breaking save files.
Sorry.

ThomasJClark:
> Okay, sure. You can disagree. I'm not merging this yet though.
> We can bring this up until the next meeting. Until then, I don't want to risk breaking save files.
> Sorry.

Jonathan, we can decide as a group what to prioritize for the next release.  It's not your decision alone.

JLLeitschuh:
I agree. Let's wait and chat about this at the next meeting.

PR #383, opened 2016-01-17
Use OpenCV instead of javax.imageio to encode M-JPEG stream

ThomasJClark:
`javax.imageio` is not supported on the embedded JRE on the roboRIO. Also, since we can convert Mats directly instead of converting them into AWT images, this is probably faster and uses less memory.


JLLeitschuh:
Sure

PR #386, opened 2016-01-18
Fix the deploy UI

### LINKED ISSUE #352: Wiki topic: How to deploy to RoboRio
I'm struggling on finding how to deploy to the RoboRio or capture the stream from the RoboRio into Grip (on the PC) for processing.

### LINKED ISSUE #367: When trying to deploy GRIP to roboRio it fails everytime.
Whenever we try to deploy GRIP to our roboRio it always fails. We have tried deploying wirelessly and have also tried do to it through the mDNS address and the IP address and have also tried it while plugged directly into the roboRio with the USB type B cable and its IP address and we still cannot successfully deploy to the roboRio. Running the program from the laptop however works fine as it is able to successfully publish the contours report to the network table but we have tried seemingly everything when it comes to attempting to deploy to the roboRio so that we can have it run in headless mode.

### LINKED ISSUE #375: Grip deploy failed with error output attached
In lower right hand corner it says success Waiting for start. However in the error output it includes the following:
[GripError.txt](https://github.com/WPIRoboticsProjects/GRIP/files/92962/GripError.txt)

Java 8 JRE is installed on the roborio with V19 firmware, computer doing the deploying is directly ethernetted to the roborio(deploy target was determined via cmd ipconfig /all) and the deploy target is the same in both settings and advanced deploy.

### LINKED ISSUE #376: Failed to Stop
Froze when I tried to end my FRC deploy.

### LINKED ISSUE #377: Fix deploying
There's definitely something up with the deploy feature.  A lot of these reports have little useful information, but hopefully between them there should be enough to figure out what's going on.  We can use this issue as a single place to collect any information about problems with deploying.

- http://www.chiefdelphi.com/forums/showpost.php?p=1523851&postcount=49
- http://www.chiefdelphi.com/forums/showpost.php?p=1524008&postcount=54
- http://www.chiefdelphi.com/forums/showpost.php?p=1524508&postcount=62
- #352
- #367
- #375
- #376

### LINKED ISSUE #388: I don't know what happened...
Thread[JavaFX Application Thread,5,main] threw an exception
## Stack Trace:
```java
java.lang.NullPointerException
	at edu.wpi.grip.ui.DeployerController.lambda$null$14(DeployerController.java:133)
	at com.sun.javafx.application.PlatformImpl.lambda$null$173(PlatformImpl.java:295)
	at java.security.AccessController.doPrivileged(Native Method)
	at com.sun.javafx.application.PlatformImpl.lambda$runLater$174(PlatformImpl.java:294)
	at com.sun.glass.ui.InvokeLaterDispatcher$Future.run(InvokeLaterDispatcher.java:95)

```
## System Info:

Property Name | Property 
 ----- | -----
GRIP Version | 1.0.0
java.version | 1.8.0_66
javafx.version | 8.0.66
os.name | Mac OS X
os.version | 10.11.3
os.arch | x86_64

ThomasJClark:
A lot of users seem to have problems using the deploy UI.  This PR should hopefully address those problems, and help us get information about future ones.

- People seem to think the deploy is frozen a lot.   I don't think we actually have a bug, I think this is either because the progress bar is always indeterminate, because you actually need to hit another button that appears after you deploy for the program to run, or because the UI actually does briefly freeze during deployment.
- I think the "FRC" vs "FRC Advanced" thing is tripping people up.  Users don't seem to completely understand the relationship between these two settings and the team number/deploy address settings, and that's probably because the two features were developed independently around the same time and kind of hacked together.   Now that both are merged in, we can make the deployment GUI offload that logic to the settings and just provide a manual override. 
- When an error does happen, people usually have very little information to provide us.  I think most cases would be solved without any effort on our part if we just had a label that plainly stated the error, and if not then at least we would have something to go on when helping users debug.

Note: all of the text fields below are automatically filled in, so users don't have to know what to use, but they can change it if they want.

### No connection to the roboRIO
unknown host

### Wrong username/password (if manually overridden)
username

### Java not installed (or in the wrong place if manually overridden)
java

### Indeterminate (ex: waiting for connections)
status

### Determinate (ex: uploading files)
uploading

## Running

#383 needs to get merged before master will run on a roboRIO again, but this is what it looks like when it starts to run.

running

## Misc. Improvements
- Everything is configurable by the user, so this should be able to be used by teams employing coprocessors (as long as they deploy with SSH)
- The project can be deployed even if it's not saved.  Also, if there have been changed made to it since the last save, this will upload the changed version.
- The GRIP jar doesn't have to be copied to a second, temporary local file.  This may improve speed and disk usage.

## TODO
- ~~Stop the pipeline when GRIP is running headlessly~~ ✓
- ~~Add a button to stop GRIP~~ ✓
- ~~Possible save username, directories, etc... to project settings?~~ ✓
- @JLLeitschuh let me know if there's anything else you can think of, or any ideas on how the current changes can be improved.

Closes #352
Closes #367
Closes #375
Closes #376
Closes #377
Closes #388

See also:
- http://www.chiefdelphi.com/forums/showpost.php?p=1523851&postcount=49
- http://www.chiefdelphi.com/forums/showpost.php?p=1524008&postcount=54
- http://www.chiefdelphi.com/forums/showpost.php?p=1524508&postcount=62
- http://www.chiefdelphi.com/forums/showpost.php?p=1525560&postcount=4


cpapplefamily:
When do you plan to release the latest version?

ThomasJClark:
I'm thinking sometime this week. If depends on when all of these PRs get finished and merged

AustinShalit:
If this is a more general deploy menu, should it be labeled with the FIRST logo?

AustinShalit:
Maybe add a button that allows users to default to FRC defaults

ThomasJClark:
> If this is a more general deploy menu, should it be labeled with the FIRST logo?

Good point, maybe a generic "play" icon would be better 

> Maybe add a button that allows users to default to FRC defaults

It actually fills in FRC defaults automatically

JLLeitschuh:
I agree with @AustinShalit on this one.
I still haven't reviewed this yet.

ThomasJClark:
> I agree with @AustinShalit on this one.

Which one?  The icon?

JLLeitschuh:
Yea, the genericizing the icon and the menu.
Also, I think that we should retain the ability to stop and then restart the deployed instance of GRIP.

JLLeitschuh:
I don't know if you care, but currently your commit message won't automatically close those issues when this is merged.

ThomasJClark:
They're already closed except #377.

Do I need to add it to a commit message to close it?

ui/src/main/java/edu/wpi/grip/ui/MainWindowController.java
```diff
@@ -185,27 +191,20 @@ public void quit() {
...
-
-            // Workaround for JavaFX's default Alert styling causing text to get cut off
-            Label alertLabel = (Label) alert.getDialogPane().getChildren().filtered(node -> node instanceof Label).get(0);
-            alertLabel.setMinHeight(Region.USE_PREF_SIZE);
-
-            alert.showAndWait();
-        }
-
+    public void deploy() {
+        ImageView graphic = new ImageView(new Image("/edu/wpi/grip/ui/icons/first.png"));
```
JLLeitschuh: Let's make this more generic since we don't just want to upload for FRC.

ui/src/main/java/edu/wpi/grip/ui/DeployController.java
```diff
@@ -0,0 +1,189 @@
...
+    public final static String DEFAULT_USER = "lvuser";
+    private final static String DEFAULT_PASSWORD = "";
+    private final static String DEFAULT_JAVA_HOME = "/usr/local/frc/JRE/";
+    private final static String DEFAULT_DIR = "/home/" + DEFAULT_USER;
+    private final static String GRIP_JAR = "grip.jar";
+    private final static String PROJECT_FILE = "project.grip";
+
+    @FXML private TextField address;
+    @FXML private TextField user;
+    @FXML private TextField password;
```
JLLeitschuh: Use a password field for this.
https://docs.oracle.com/javase/8/javafx/api/javafx/scene/control/PasswordField.html

ui/src/main/java/edu/wpi/grip/ui/MainWindowController.java
ThomasJClark: K

ui/src/main/java/edu/wpi/grip/ui/DeployController.java
ThomasJClark: I am using a password field...

ui/src/main/java/edu/wpi/grip/ui/DeployController.java
```diff
@@ -0,0 +1,216 @@
...
+    @FXML private TextField deployDir;
+    @FXML private ProgressIndicator progress;
+    @FXML private Button deployButton;
+    @FXML private Label status;
+    @FXML private TextArea console;
+
+    @Inject private EventBus eventBus;
+    @Inject private Project project;
+    @Inject private Pipeline pipeline;
+    @Inject private Logger logger;
```
JLLeitschuh: Formatting

core/src/main/java/edu/wpi/grip/core/settings/ProjectSettings.java
```diff
@@ -12,21 +12,48 @@
...
+    private String deployAddress = computeFRCAddress(teamNumber);
 
+    @Setting(label = "Default deploy directory", description = "The directory on the remote host to deploy GRIP to.")
+    private String deployDir = "/home/lvuser";
+
+    @Setting(label = "Default deploy user", description = "The username to log in with when deploying over SSH.")
+    private String deployUser = "lvuser";
+
+    @Setting(label = "Default deploy java home", description = "Where Java is installed on the robot.")
+    private String deployJavaHome = "/usr/local/frc/JRE/";
```
JLLeitschuh: I don't like that this is specific to FRC by default.
I think we should do this differently, perhaps a button to assign the FRC defaults.

ui/src/main/java/edu/wpi/grip/ui/DeployController.java
ThomasJClark: Both ways are used in the project, and we have no officially documented style guide.  I've mostly been  going by the Google Java Style document for now:
> A single parameterless annotation **may** instead appear together with the first line of the signature

This is consistent with the project so far, and I think is much more readable than this:

```java
@FXML
private TextField address;
@FXML
private TextField user;
@FXML
private TextField password;
@FXML
private TextField javaHome;
@FXML
private TextField deployDir;
@FXML
private ProgressIndicator progress;
@FXML
private Button deployButton;
@FXML
private Label status;
@FXML
private TextArea console;
@Inject
private EventBus eventBus;
@Inject
private Project project;
@Inject
private Pipeline pipeline;
@Inject
private Logger logger;
```

core/src/main/java/edu/wpi/grip/core/settings/ProjectSettings.java
ThomasJClark: FRC teams are really our primary audience, especially for the deploy tool.  I think we'll have lots of issues/angry CD posts if we don't just make it work when you hit "Deploy".

IMO researchers/hobbyists/college students can probably figure out pretty quickly how to configure the deploy tool, if they use it at all.

ui/src/main/java/edu/wpi/grip/ui/DeployController.java
JLLeitschuh: We haven't really made an official decision on styling. We decided originally to use the default for Intellij. If you want to change it then we need to talk about it as a team.

ThomasJClark: > We haven't really made an official decision on styling

Like I said, this is true.

> We decided originally to use the default for Intellij.

Like both of us said, this isn't true.  Although I have recommended using IntelliJ's autoformatter, we never said our official style guide was the default settings in IntelliJ, and in practice our codebase more closely matches Google Java Style.

>If you want to change it then we need to talk about it as a team.

I'm not changing anything.  This code does not violate any official style guidelines we have, or any written but unofficial ones, and it's consistent with our current code.  If we want to establish an official style guide, I would think *that* would involve meeting as a team (although really I don't see why anyone would object to officially adopting Google's style)

ui/src/main/resources/edu/wpi/grip/ui/Deploy.fxml
```diff
@@ -0,0 +1,69 @@
...
+<?import javafx.scene.layout.*?>
+<VBox styleClass="deploy-pane" maxWidth="Infinity" xmlns:fx="http://javafx.com/fxml/1"
+      xmlns="http://javafx.com/javafx/null" fx:controller="edu.wpi.grip.ui.DeployController">
+    <GridPane maxHeight="Infinity">
+        <columnConstraints>
+            <ColumnConstraints hgrow="NEVER"/>
+            <ColumnConstraints hgrow="ALWAYS"/>
+        </columnConstraints>
+
+        <Label disable="${deployButton.disabled}" text="Address" GridPane.columnIndex="0" GridPane.rowIndex="0"/>
```
JLLeitschuh: For all labels make sure you set what the label is a label for.
There should be a tag that is called `for` or similar.
https://docs.oracle.com/javafx/2/api/javafx/scene/control/Label.html#labelForProperty()

ThomasJClark: `labelFor` is for keyboard shortcuts, which I'm not using here

ui/src/main/java/edu/wpi/grip/ui/DeployController.java
```diff
@@ -0,0 +1,216 @@
...
+                        if (isNotCanceled())
+                            Platform.runLater(() -> progress.setProgress((double) transferred / size));
+                    };
+                }
+            });
+
+            // The project may or may not be saved to a file (and even if it was saved, it might be modified), so we
+            // serialize it to a string before deploying.
+            StringWriter projectWriter = new StringWriter();
+            project.save(projectWriter);
```
JLLeitschuh: Shouldn't we just ask if they want to save before they open the deploy UI?

ThomasJClark: Why?  What if you've made some changes that you want to test before saving?

JLLeitschuh: Sure. Thats fine. But shouldn't you offer them the option of saving before deploying.

ThomasJClark: Why?

JLLeitschuh:
Closing the deploy UI doesn't stop the upload attempt.

ThomasJClark:
> Closing the deploy UI doesn't stop the upload attempt.

That's correct.  I think people want to be able to work on their pipeline while they're running a test on the roboRIO to adjust things if it needs improvement.  That's why I added separate "Stop" and "Close" buttons.

stephenstockman:
The new deploy interface looks awesome and am especially happy about the new method of seeing errors great work. As for the whole prompt for saving as a user I just want to be able to quickly alter and test different values without overwriting a previous working file for this getting promoted to save all the time will eventually get annoying. When something works better I'll either save it as a seperate file or overwrite the existing one manually.

JLLeitschuh:
Moving chat to here so it doesn't get deleted by updating the files:

> I don't like that this is specific to FRC by default.
I think we should do this differently, perhaps a button to assign the FRC defaults.
FRC teams are not our exclusive audience and if we do this I feel that it will make people feel that this application won't work for them if they aren't using a FRC control system
> \- [Jonathan](https://github.com/WPIRoboticsProjects/GRIP/pull/386#discussion_r50277450)

.

> FRC teams are really our primary audience, especially for the deploy tool. I think we'll have lots of issues/angry CD posts if we don't just make it work when you hit "Deploy".

> IMO researchers/hobbyists/college students can probably figure out pretty quickly how to configure the deploy tool, if they use it at all.

> Maybe eventually we can have a project wizard, and one of the templates you can select would be FRC
> \- [Tom](https://github.com/WPIRoboticsProjects/GRIP/pull/386#discussion_r50280267)


I think we should have different deploy configurations. For example, deploying to a PI may require more configuration than to an FRC device.
We may need to mess with jars more for deploying to a PI. There's a whole problem with hard vs softabi and needing to upload different versions of network tables for that to work. (I was talking to Patrick and Fred about this)

core/src/main/java/edu/wpi/grip/core/settings/ProjectSettings.java
nightpool: I'm with @ThomasJClark on this one—right now non-FRC systems are the exception.

maybe the right thing to do would be to provide a FRC-specific build that makes a lot of these things the default?

nightpool:
@JLLeitschuh 

I'm with @ThomasJClark on this one—right now non-FRC systems are the exception, especially when it comes to deployment. I don't even really see it as a very useful option in many other contexts tbqh.

maybe the right thing to do would be to provide a FRC-specific build that makes a lot of these things the default?

PR #392, opened 2016-01-20
Pipeline run in own thread

### LINKED ISSUE #190: New Pipeline Structure
Pipeline should be run in a separate thread

The pipeline should have a timer and will run at a specific rate.
Setting the source value will no longer runs the pipeline; instead it flags a dirty bit.

```java
for(;;) {
    if(sem.wait(time) || sources.dirty) {
        run();
    }
}
```

This will make everything sequential and makes mutable data safe.

JLLeitschuh:
Refactors the pipeline to run in its own thread.
This also means that steps will always run sequentially.
Sources will have their sockets updated by the pipeline instead
of doing so manually.

Closes #190

# Todo
- [x] Make it possible to stop/start the pipeline from within the UI.
- [ ] Make it possible to dynamically set the the running rate of the pipeline.


core/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
```diff
@@ -165,6 +169,25 @@ public String getName() {
     }
 
     @Override
+    protected boolean updateOutputSockets() {
```
ThomasJClark: JavaCV's `FrameGrabber` API works by polling.  If we're changes sources to poll instead of using events, we should just call the JavaCV functions here instead of running it in a separate thread

JLLeitschuh: 1. If you have multiple camera sources then that will throttle the pipeline to the slowest one.
The way that I implemented we only run the pipeline when any of the sources/steps have new values.

2. Imagine in an extreme case where you have an IPCamera that provides images at 1 frame per 5 seconds. You want to be able to change pipeline values and see those changes without having the pipeline be blocked on one source.

3. Also, what happens if a source crashes? This can cause the `grab` method to hang for some indeterminate length of time. That would halt the entire pipeline until `grab` returns.

ThomasJClark: > If you have multiple camera sources then that will throttle the pipeline to the slowest one.

I think this is what most people would want.  If you have two cameras at 24fps and 23fps, you wouldn't want to run at 47fps.

core/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
```diff
@@ -74,7 +82,7 @@
 
         public FrameGrabber create(int deviceNumber) {
             // On Windows, videoInput is much more reliable for webcam capture.  On other platforms, OpenCV's frame
-            // grabber class works fine.
+            // grabberSupplier class works fine.
```
JLLeitschuh: Fix this

core/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
```diff
@@ -165,151 +251,60 @@ public String getName() {
...
+    public Service.State getState() {
+        return cameraService.state();
     }
 
     @Subscribe
     public void onSourceRemovedEvent(SourceRemovedEvent event) throws InterruptedException, TimeoutException, IOException {
         if (event.getSource() == this) {
             try {
-                if (this.isStarted()) this.stop();
+                if (this.isRunning()) this.stopAsync();
```
JLLeitschuh: I no longer need to check if its running.

core/src/main/java/edu/wpi/grip/core/util/service/RestartableService.java
```diff
@@ -0,0 +1,165 @@
...
+
+        @Override
+        public void failed(final State from, final Throwable failure) {
+            if (policy.shouldRestart()) {
+                startAsync(false);
+            }
+        }
+    }
+
+    private synchronized Service createNewDelegate() {
```
JLLeitschuh: Turn this into its own consumer

ui/src/main/java/edu/wpi/grip/ui/components/StartStoppableButton.java
```diff
@@ -19,15 +20,13 @@
...
-import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
-import java.util.concurrent.TimeoutException;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
 public final class StartStoppableButton extends ToggleButton {
-    protected static final String BASE_STYLE_CLASS = "start-stoppable-button";
+    protected static final String BASE_STYLE_CLASS = "startAsync-stoppable-button";
```
JLLeitschuh: fix this

core/src/main/java/edu/wpi/grip/core/Pipeline.java
```diff
@@ -44,27 +57,108 @@
...
 
+    private final <R> R readSourcesSafely(Function<List<Source>, R> sourceListFunction) {
+        return accessSafely(sourceLock.readLock(), Collections.unmodifiableList(sources), sourceListFunction);
+    }
+
+    /**
+     * Returns a snapshot of all of the sources in the pipeline.
+     *
+     * @return an Immutable copy of the sources at the current point in the pipeline.
+     * @see <a href="https://youtu.be/ZeO_J2OcHYM?t=16m35s">Why we use ImmutableList return type</a>
```
ThomasJClark: Based on that talk, we should just use immutable collections for sources and steps.  When we modify the pipeline, we can replace the existing list with a modified copy, making all of this stuff simpler and have a more standard interface.

So, we should ask ourselves why we need something to be mutable instead of asking ourselves why it should be immutable.

Thoughts?

core/src/main/java/edu/wpi/grip/core/Pipeline.java
```diff
@@ -151,20 +249,24 @@ public void onSourceRemoved(SourceRemovedEvent event) {
         }
     }
 
-    public synchronized void addStep(int index, Step step) {
+    public void addStep(int index, Step step) {
```
ThomasJClark: So, this method is a wrapper around a wrapper around a wrapper around a wrapper around `lock.lock()` and `lock.unlock()`.  I personally found this difficult to follow, and it seems to have significantly increased the line count in this file.  I wonder if maybe it might be a good idea to use less layers of abstraction here, and just call `lock()` and `unlock()` directly.

core/src/main/java/edu/wpi/grip/core/PipelineRunner.java
```diff
@@ -0,0 +1,159 @@
...
+    @Override
+    public void stopAsync() {
+        pipelineService.stopAsync();
+        // Ensure that we unblock the pipeline so it can actually stop
+        pipelineFlag.countDown();
+    }
+
+    @Override
+    public Service.State getState() {
+        return pipelineService.state();
```
ThomasJClark: If you're gonna put all of the methods of `Service` in here, why not make this implement `Service`?  This is a classic is-a relationship.

core/src/main/java/edu/wpi/grip/core/events/RenderEvent.java
```diff
@@ -0,0 +1,4 @@
+package edu.wpi.grip.core.events;
+
+public class RenderEvent {
```
ThomasJClark: Document this

core/src/main/java/edu/wpi/grip/core/events/StopPipelineEvent.java
```diff
@@ -1,7 +1,7 @@
 package edu.wpi.grip.core.events;
 
 /**
- * This event is posted to tell the pipeline to stop executing.
+ * This event is posted to tell the pipeline to stopAsync executing.
```
ThomasJClark: Fix this

core/src/main/java/edu/wpi/grip/core/util/service/SingleActionListener.java
```diff
@@ -0,0 +1,40 @@
+package edu.wpi.grip.core.util.service;
+
+import com.google.common.util.concurrent.Service;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Will run the same runnable regardless of what listener callback is being called.
+ */
+public class SingleActionListener extends Service.Listener {
```
ThomasJClark: What even is this?

ui/src/main/java/edu/wpi/grip/ui/preview/PreviewsController.java
```diff
@@ -26,20 +29,20 @@
 @Singleton
 public class PreviewsController {
 
-    @FXML private HBox previewBox;
+    @FXML
```
ThomasJClark: Why did you change this class?  What does this have to do with the pipeline stuff?

ThomasJClark:
Why can I have a camera source running while the pipeline is stopped?  What does that even mean?  The camera is capturing images and throwing them away?

ThomasJClark:
It seems like `CameraSource` will keep retrieving frames while the pipeline is running, and just throw them away in the mean time.  Since cameras may run up to around 30 FPS and the pipeline will probably run at 10-15 or less, this seems like a bad idea.

After looking at the JavaCV API more closely, you can actually get the cvGrab/cvRetrieve functionality without modifying `OpenCVFrameGrabber` by calling `setTriggerMode(true)`.  I'm not sure but I think that might be helpful.

ThomasJClark:
I'm not a big fan of the Running/Terminated labels.  Partially the terminology, but I also think it's confusing how that relates to the play button on the bottom right of the window despite being so far from it.

core/src/main/java/edu/wpi/grip/core/util/service/SingleActionListener.java
JLLeitschuh: It runs the same function for every single listener callback method

ThomasJClark: Do you even use this?  Like, more than once.

core/src/main/java/edu/wpi/grip/core/PipelineRunner.java
JLLeitschuh: I'm not pulling all of the methods out of service.
I'm only using 4/10 of them.
I don't need the other 6.

ThomasJClark: I know, but you're using a lot of them, and you're using a service internally anyways

core/src/main/java/edu/wpi/grip/core/Pipeline.java
JLLeitschuh: It ensures you lock an unlock correctly. Event if there is an exception.
If there is an exception and you forget to unlock in a `finally` block you could deadlock the entire application.

core/src/main/java/edu/wpi/grip/core/Pipeline.java
JLLeitschuh: Doesn't XStream deserialization only support creating ArrayLists?

core/src/main/java/edu/wpi/grip/core/Pipeline.java
ThomasJClark: > If there is an exception and you forget to unlock in a finally block you could deadlock the entire application.

```java
lock.lock();
try {
    eventBus.register(step);
    eventBus.post(new StepAddedEvent(step, index));
} finally {
    lock.unlock();
}
```

> It ensures you lock an unlock correctly

I'm not sure I understand this.  If you do the above, the lock will definitely get unlocked no matter what.  Since it's private, you don't have to worry about any external class messing with it, and someone reading this method can clearly see exactly how it works.

I believe this is pretty standard usage of `Lock`, and I'm not sure why we would abstract that into something else unless it were for public methods.

JLLeitschuh:
> Why can I have a camera source running while the pipeline is stopped? What does that even mean? The camera is capturing images and throwing them away?

So should stopping the pipeline stop all of the sources as well?

> It seems like CameraSource will keep retrieving frames while the pipeline is running, and just throw them away in the mean time. Since cameras may run up to around 30 FPS and the pipeline will probably run at 10-15 or less, this seems like a bad idea.

> After looking at the JavaCV API more closely, you can actually get the cvGrab/cvRetrieve functionality without modifying OpenCVFrameGrabber by calling setTriggerMode(true). I'm not sure but I think that might be helpful.

Can you link me a document that describes how this works and an example implementation?

> I'm not a big fan of the Running/Terminated labels. Partially the terminology, but I also think it's confusing how that relates to the play button on the bottom right of the window despite being so far from it.

So move the text and the button closer together? Also, map the names to something more friendly? Something like `Pipeline Stopped` `Pipeline Running`

core/src/main/java/edu/wpi/grip/core/Pipeline.java
ThomasJClark: XStream doesn't deserialize these lists, we populate them with `StepAddedEvent`s, `SourceAddedEvent`s, etc...

ThomasJClark:
> So should stopping the pipeline stop all of the sources as well?

I don't think stopping the pipeline and stopping sources should both be things.  One or the other.

>Can you link me a document that describes how this works and an example implementation?

http://docs.opencv.org/2.4/modules/highgui/doc/reading_and_writing_images_and_video.html#videocapture-grab

In the JavaCV FrameGrabber wrapper with `setTriggerMode(true)`:
- `trigger()` basically means `cv.GrabFrame()`
- `grab()` is basically `cv.RetrieveFrame()`

Kind of confusing choice of words, but it's actually a little simpler than I originally thought.

> So move the text and the button closer together? Also, map the names to something more friendly? Something like Pipeline Stopped Pipeline Running

If we decide the keep the whole "starting" and "stopping" the pipeline thing, I would either get rid of the text, or make a button with text that changes between "Start Pipeline" and "Stop Pipeline".  I like the second one, since there's currently no indication of what that play button controls.

ui/src/main/java/edu/wpi/grip/ui/preview/PreviewsController.java
JLLeitschuh: Did you even read the rest of the changes to this class?

core/src/main/java/edu/wpi/grip/core/Pipeline.java
JLLeitschuh: So you want to copy on write instead of copy on read?

core/src/main/java/edu/wpi/grip/core/Pipeline.java
```diff
@@ -35,7 +41,14 @@
     @XStreamOmitField
     private NTManager ntManager;
 
+    /*
+     * We have separate locks for sources and steps because we don't want to
+     * block access to both resources when only one is in use.
+     */
+
+    private final ReadWriteLock sourceLock = new ReentrantReadWriteLock();
```
JLLeitschuh: Make this `transient`

core/src/main/java/edu/wpi/grip/core/Pipeline.java
JLLeitschuh: Spot the deadlock?
```
lock.lock();
try {
    eventBus.register(step);
    eventBus.post(new StepAddedEvent(step, index));
} finally {
    lock.unlock();
}
```
If you wrote that code, congratulations, you just deadlocked the application.
You should never post events from within a lock.

What I have done is write convenient helper methods for locking and unlocking the pipeline safely.
If we ever find that that core lock/unlock procedure is faulty we don't need to modify the same code in multiple places. This removes duplicate code.

core/src/main/java/edu/wpi/grip/core/events/RenderEvent.java
JLLeitschuh: You know that we could avoid using `runAndWait` or anything similar by providing permits to read socket values from the pipeline. You could acquire a permit for the socket value, read the value, copy it into your own local variable and then let it go.
This would even work if we ever used the async event bus.

I don't know if this idea is completely convoluted.

core/src/main/java/edu/wpi/grip/core/util/service/SingleActionListener.java
JLLeitschuh: > Do you even use this?
Yes

> Like, more than once.
No?

core/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
```diff
@@ -40,7 +50,8 @@
 
     private final static String DEVICE_NUMBER_PROPERTY = "deviceNumber";
     private final static String ADDRESS_PROPERTY = "address";
-    private static Logger logger = Logger.getLogger(CameraSource.class.getName());
+    private static final int MAX_CAMERA_FAILURE_COUNT = 100;
```
JLLeitschuh: This shouldn't be here.

core/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
```diff
@@ -144,14 +153,95 @@ public FrameGrabber create(String addressProperty) throws MalformedURLException
...
-            this.grabber = grabberFactory.create(deviceNumber);
+            this.grabberSupplier = () -> grabberFactory.create(deviceNumber);
         } else if (addressProperty != null) {
             this.name = "IP Camera " + new URL(addressProperty).getHost();
-            this.grabber = grabberFactory.create(addressProperty);
+            this.grabberSupplier = () -> {
+                try {
+                    return grabberFactory.create(addressProperty);
+                } catch (MalformedURLException e) {
+                    throw new IllegalStateException(e.getMessage(), e);
```
JLLeitschuh: Make this an illegal argument.

ui/src/test/java/edu/wpi/grip/ui/MainWindowTest.java
```diff
@@ -20,10 +24,16 @@ public void start(Stage stage) throws Exception {
...
 
         final PaletteController palette = main.injector.getInstance(PaletteController.class);
+        pipelineRunner = main.injector.getInstance(PipelineRunner.class);
         palette.clearOperations();
         palette.onOperationAdded(new OperationAddedEvent(new AdditionOperation()));
     }
 
+    @After
+    public void tearDown() {
+        pipelineRunner.stopAsync();
```
JLLeitschuh: Add the await.

ThomasJClark:
Do "removed", "cleanup", and "performDisabled" mean the same thing?  If so, name them similarly so that correlation is obvious.

`Pipeline#sources` and `Pipeline#steps` should be immutable.  I see no reason why they must be mutable.

core/src/main/java/edu/wpi/grip/core/Step.java
```diff
@@ -29,6 +28,8 @@
     private final InputSocket<?>[] inputSockets;
     private final OutputSocket<?>[] outputSockets;
     private final Optional<?> data;
+    private final AtomicBoolean performDisabled = new AtomicBoolean(false);
```
ThomasJClark: performDisabled does not need to be an AtomicBoolean

core/src/main/java/edu/wpi/grip/core/events/ConnectionAddedEvent.java
```diff
@@ -3,18 +3,20 @@
...
-public class ConnectionAddedEvent {
+public class ConnectionAddedEvent implements RunPipelineEvent {
     private final Connection connection;
 
     /**
      * @param connection The connection being added
      */
     public ConnectionAddedEvent(Connection connection) {
-        this.connection = connection;
+        this.connection = checkNotNull(connection, "Connection can not be null");
```
ThomasJClark: Nitpicking, but *cannot. "can not" usually indicates that "not" is modifying the word after it.

core/src/main/java/edu/wpi/grip/core/events/ConnectionRemovedEvent.java
```diff
@@ -14,7 +16,7 @@
      * @param connection The connection being deleted
      */
     public ConnectionRemovedEvent(Connection connection) {
-        this.connection = connection;
+        this.connection = checkNotNull(connection, "Connection can not be null");
```
ThomasJClark: same

core/src/main/java/edu/wpi/grip/core/events/ExceptionEvent.java
```diff
@@ -22,7 +23,7 @@
      * @param message   The message associated with this event.
      *                  If <tt>null</tt> will use {@link Exception#getMessage()}
      */
-    public ExceptionEvent(Object origin, Exception exception, String message) {
+    public ExceptionEvent(Object origin, @Nullable Exception exception, @Nullable String message) {
```
ThomasJClark: What?  Since when did we use a framework that provides `@Nullable`?  And why are you even modifying this class in this PR?

core/src/main/java/edu/wpi/grip/core/events/OperationAddedEvent.java
```diff
@@ -14,9 +16,11 @@
      * @param operation The operation being added
      */
     public OperationAddedEvent(Operation operation) {
```
ThomasJClark: Why are you modifying all of these classes that have nothing to do with this PR?

core/src/main/java/edu/wpi/grip/core/events/RenderEvent.java
```diff
@@ -0,0 +1,8 @@
+package edu.wpi.grip.core.events;
+
+/**
+ * An event that occurs when the pipeline has completed a run.
+ * Anything that needs to render an image from the pipeline should do so by subscribing to this event.
```
ThomasJClark: {＠link} to whatever actually uses this

core/src/main/java/edu/wpi/grip/core/events/RunPipelineEvent.java
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Any event that indicates to the {@link edu.wpi.grip.core.PipelineRunner} that it should run.
+ */
+public interface RunPipelineEvent {
+
+    /**
+     * Indicates to the {@link edu.wpi.grip.core.PipelineRunner} that there is an update to one of the values and it should run again.
+     * @return true if the pipeline should be released to run because of the update
+     */
+    default boolean pipelineShouldRun() {
```
ThomasJClark: This makes no sense.  Things can send `RunPipelineEvent`s, but sometimes it doesn't mean run?

core/src/main/java/edu/wpi/grip/core/events/SocketConnectedChangedEvent.java
```diff
@@ -17,7 +19,7 @@
      * @param socket The socket that was either connected or disconneted.
      */
     public SocketConnectedChangedEvent(Socket socket) {
```
ThomasJClark: Why did you modify this class?  Please don't modify files unrelated to your PR.

core/src/main/java/edu/wpi/grip/core/events/SocketPreviewChangedEvent.java
```diff
@@ -14,7 +16,7 @@
      * @param socket The socket being previewed
      */
     public SocketPreviewChangedEvent(OutputSocket socket) {
-        this.socket = socket;
+        this.socket = checkNotNull(socket, "Socket can not be null");
```
ThomasJClark: Same

core/src/main/java/edu/wpi/grip/core/events/SourceAddedEvent.java
```diff
@@ -17,9 +19,11 @@
      * @param source The source being added
      */
     public SourceAddedEvent(Source source) {
-        this.source = source;
+        this.source = checkNotNull(source, "Source can not be null");
```
ThomasJClark: Again

core/src/main/java/edu/wpi/grip/core/events/SourceHasPendingUpdateEvent.java
```diff
@@ -0,0 +1,24 @@
...
+import edu.wpi.grip.core.Source;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * An event that occurs when a source has an update pending for the pipeline to handle.
+ * This event should be triggered when a Source a value that is ready to be moved into
+ * one or more of its sockets.
+ */
+public final class SourceHasPendingUpdateEvent implements RunPipelineEvent {
```
ThomasJClark: What's the difference between this and SocketChangedEvent?

core/src/main/java/edu/wpi/grip/core/events/SourceRemovedEvent.java
```diff
@@ -17,7 +19,7 @@
      * @param source The source being removed
      */
     public SourceRemovedEvent(Source source) {
-        this.source = source;
+        this.source = checkNotNull(source, "Source can not be null");
```
ThomasJClark: Again...

core/src/main/java/edu/wpi/grip/core/events/StepAddedEvent.java
```diff
@@ -11,23 +15,23 @@
...
+    private final OptionalInt index;
 
     /**
      * @param step  The step being added
      * @param index The index that the step should be added at.
      */
-    public StepAddedEvent(Step step, int index) {
-        this.step = step;
-        this.index = Optional.of(index);
+    public StepAddedEvent(Step step, @Nonnegative int index) {
```
ThomasJClark: Again...

core/src/main/java/edu/wpi/grip/core/events/StepMovedEvent.java
```diff
@@ -15,7 +17,7 @@
      * @param distance The number of indices (positive or negative) to move the step by
      */
     public StepMovedEvent(Step step, int distance) {
-        this.step = step;
+        this.step = checkNotNull(step, "Step must not be null");
```
ThomasJClark: This one too...

ui/src/main/java/edu/wpi/grip/ui/preview/PreviewsController.java
ThomasJClark: Yes, almost all of them seem to be random unnecessary refactoring to your preferred level of abstraction and coding style.

core/src/main/java/edu/wpi/grip/core/events/RenderEvent.java
JLLeitschuh: You can see that by using your editor.

core/src/main/java/edu/wpi/grip/core/events/SourceHasPendingUpdateEvent.java
JLLeitschuh: The socket hasn't changed.
I'll update the documentation to reflect this.

core/src/main/java/edu/wpi/grip/core/events/StepAddedEvent.java
```diff
@@ -1,9 +1,13 @@
 package edu.wpi.grip.core.events;
 
 import com.google.common.base.MoreObjects;
-import com.google.common.base.Optional;
```
JLLeitschuh: This class was using the google common Optional, we are using the util one.

core/src/test/java/edu/wpi/grip/core/PipelineRunnerTest.java
```diff
@@ -0,0 +1,363 @@
...
+
+        public synchronized void throwIfProblemPresent() throws Throwable {
+            try {
+                if (failedFrom != null || failure != null) {
+                    System.err.println("Failed from state " + failedFrom);
+                    throw failure;
+                }
+            } finally {
+                failedFrom = null;
+                failure = null;
```
JLLeitschuh: :+1:

core/src/main/java/edu/wpi/grip/core/PipelineRunner.java
```diff
@@ -0,0 +1,212 @@
...
+                        // Acquire the first this one should permit if there is at least one permit
+                        pipelineFlag.acquire();
+                        // Acquire the rest of the permits from the flag
+                        // Every time release is called another permit is added.
+                        // We need to clean up any old permits that we may have been given.
+                        pipelineFlag.acquire(
+                                Math.max(0, pipelineFlag.availablePermits()));
+
+                        if (!super.isRunning()) return;
+                        runPipeline(this::isRunning);
```
JLLeitschuh: This should be `super::isRunning`

core/src/main/java/edu/wpi/grip/core/Step.java
JLLeitschuh: You are correct. I will change that.

core/src/main/java/edu/wpi/grip/core/events/RunPipelineEvent.java
JLLeitschuh: Yes, some events have conditional logic regarding whether or not it should run. https://github.com/WPIRoboticsProjects/GRIP/pull/392/files#diff-18bf65d3e77dad8266ea91101e0284feR32

JLLeitschuh:
> Do "removed", "cleanup", and "performDisabled" mean the same thing? If so, name them similarly so that correlation is obvious.

Sure

> Pipeline#sources and Pipeline#steps should be immutable. I see no reason why they must be mutable

They are mutable though. You can add steps, you can remove steps.
The pipeline is a mutable object, there's no way around that. By making them immutable you would have to copy them every time you wanted to modify them. This leads to race conditions if you call `add` from two separate threads.

core/src/main/java/edu/wpi/grip/core/events/ExceptionEvent.java
JLLeitschuh: Because I added a google sanity test to the package. It was either fix the code or disable it in the tests. I decided to fix it in the code.

ThomasJClark:
> The pipeline is a mutable object, there's no way around that. By making them immutable you would have to copy them every time you wanted to modify them. This leads to race conditions if you call add from two separate threads.

It's kind of the opposite, actually.  Make a synchronized `add` method that makes a new, modified copy of the list.  When you return the list, just return the field.  Since the object the field references will never change, you don't need to do nearly as much complicated synchronization around consumers.

core/src/main/java/edu/wpi/grip/core/events/SourceHasPendingUpdateEvent.java
ThomasJClark: What's a "pending update" then?  It can happen without changing a socket?

ui/src/main/java/edu/wpi/grip/ui/preview/PreviewsController.java
JLLeitschuh: Then put a comment on those lines and I'll leave responses explaining the change. Or better yet, ask me to leave a comment explaining why I changed something.

core/src/main/java/edu/wpi/grip/core/events/ExceptionEvent.java
ThomasJClark: Ok.  I would wait and do that change in a different PR so everyone has a chance to weigh in on whether or not to use it.

ui/src/main/java/edu/wpi/grip/ui/preview/PreviewsController.java
ThomasJClark: Alright. 

- Why did you change the formatting of lines 31-37? You didn't need to change anything anywhere near them.
- Why did you make `compareSockets` static and add more parameters than `Comparable.compareTo()` actually has?  Just leave it the way it is and use member variables to get the lists.
- Why did you add `final` to a bunch of variables that you otherwise didn't have to modify?
- Why did you add `getPreviews()`?

Keep in mind that I'm not necessarily arguing that all of these are bad ideas (although I believe most of them are).  I just don't think you should be doing these random unrelated refactors in this PR.

core/src/main/java/edu/wpi/grip/core/events/OperationAddedEvent.java
JLLeitschuh: https://github.com/WPIRoboticsProjects/GRIP/pull/392#discussion_r51906483

core/src/main/java/edu/wpi/grip/core/events/SocketConnectedChangedEvent.java
JLLeitschuh: https://github.com/WPIRoboticsProjects/GRIP/pull/392#discussion_r51906483

core/src/main/java/edu/wpi/grip/core/events/SocketPreviewChangedEvent.java
JLLeitschuh: https://github.com/WPIRoboticsProjects/GRIP/pull/392#discussion_r51906483

core/src/main/java/edu/wpi/grip/core/events/SourceAddedEvent.java
JLLeitschuh: https://github.com/WPIRoboticsProjects/GRIP/pull/392#discussion_r51906483

core/src/main/java/edu/wpi/grip/core/events/SourceRemovedEvent.java
JLLeitschuh: https://github.com/WPIRoboticsProjects/GRIP/pull/392#discussion_r51906483

core/src/main/java/edu/wpi/grip/core/events/StepAddedEvent.java
JLLeitschuh: https://github.com/WPIRoboticsProjects/GRIP/pull/392#discussion_r51906483

Also, this class was using the wrong optional type.

core/src/main/java/edu/wpi/grip/core/events/StepMovedEvent.java
JLLeitschuh: https://github.com/WPIRoboticsProjects/GRIP/pull/392#discussion_r51906483

JLLeitschuh:
So by doing this you are hoping to gain readability, efficiency, thread safety?

What advantage does this get you? 
Also, if you synchronize on the whole add function you have the same deadlock I detailed here:
https://github.com/WPIRoboticsProjects/GRIP/pull/392#discussion_r50998450

ui/src/main/java/edu/wpi/grip/ui/preview/PreviewsController.java
```diff
@@ -69,10 +69,40 @@ public synchronized void onSocketPreviewChanged(SocketPreviewChangedEvent event)
...
+     */
+    private void sortPreviews(ObservableList<SocketPreviewView<?>> previews) {
+        assert Platform.isFxApplicationThread() : "Must be run in JavaFX thread";
+        // Take a snapshot of the sources and the steps in the current state so the comparison won't break.
+        // If we don't take a copy of the lists before trying to sort them then any concurrent modification to the pipeline
+        // could cause the sort result to be wrong or.
+        final ImmutableList<Source> sources = pipeline.getSources();
+        final ImmutableList<Step> steps = pipeline.getSteps();
+
+        final Comparator<SocketPreviewView<?>> comparePreviews =
```
ThomasJClark:
```java
FXCollections.sort(previews, Comparator.comparing(SocketPreviewView::getSocket,
                (a, b) -> compareSockets(a, b, pipeline.getSteps(), pipeline.getSources())));
```

PR #402, opened 2016-01-21
Fix std::thread thing on linux by vendoring NetworkTables

ThomasJClark:
For some reason, manually building ntcore.so fixes it


JLLeitschuh:
Fails on travis anyways. Should we wait for them to update NetworkTables and just use that??

ThadHouse:
What's the bug that's happening with ntcore and std::thread?

ThomasJClark:
ntcore was crashing on linux, but only in a packaged binary (running from gradle works fine).  I think this is because it's built without `-pthread`.  Running from gradle is ok because java must already link with pthread, but in the packaged JVM that JavaFX puts together, it doesn't work.

Rebuilding networktables on my machine and manually using it as the dependency for linux builds also works.  I think this is because newer versions of GCC don't require `-pthread`.

JLLeitschuh:
Can you resolve this before you leave @ThomasJClark
I want to be able to build releases myself and I don't have an Ubuntu OS to do so currently.

AustinShalit:
This was solved by the ntcore team. Closing the PR because it is no longer needed.

PR #409, opened 2016-01-22
Fixes Versioning breaking with leading v

JLLeitschuh:


ThomasJClark:
What about if the tag ends with "-rc1" or something?

What about non-tagged commits, where it appends part of the hash?

JLLeitschuh:
Those will work fine. Remember when we did the pre-release builds. Those build with no problem on linux.

JLLeitschuh:
Also, for developers having the commit hash will give us the right commit to pull to put any stack traces in context.

ThomasJClark:
They won't work, I literally just tested this.

PR #437, opened 2016-01-26
Fix lack of URL decoding when deploying

### LINKED ISSUE #435: Problem deploying GRIP
grip deploy error

After having updated GRIP, we are unable to deploy GRIP to the roboRIO, despite it having worked before. Following the path, we clearly have the .jar file, but GRIP says that it is "not a regular file or directory." We have no idea why it isn't working.

Any help would be appreciated. Thank you.

ThomasJClark:
This was causing problems when the full path of the local GRIP jar has spaces in it.

Closes #435

### DIFF
--- a/ui/src/main/java/edu/wpi/grip/ui/DeployController.java
+++ b/ui/src/main/java/edu/wpi/grip/ui/DeployController.java
@@ -27,6 +27,7 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.InterruptedIOException;
 import java.io.StringWriter;
+import java.net.URLDecoder;
 import java.net.UnknownHostException;
 import java.util.Optional;
 import java.util.logging.Level;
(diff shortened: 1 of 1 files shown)


ui/src/main/java/edu/wpi/grip/ui/DeployController.java
```diff
@@ -160,7 +162,8 @@ private void deploy(String address, String user, String password, String javaHom
             setStatusAsync("Running GRIP", false);
             Session session = ssh.startSession();
             session.allocateDefaultPTY();
-            Session.Command cmd = session.exec(javaHome + "/bin/java -jar " + deployDir + "/" + GRIP_JAR + " " + deployDir + "/" + PROJECT_FILE);
+            Session.Command cmd = session.exec(javaHome + "/bin/java -jar " + deployDir + "/" + GRIP_JAR + "' '"
+                    + deployDir + "/" + PROJECT_FILE + "'");
```
JLLeitschuh: Shouldn't we support uploading multiple project files and allowing the user to pick which one that they want to use?

PR #449, opened 2016-01-28
Issue316: make a test for ALL the steps to ensure backwards compatibility.

PaulaRudy:
This closes issue #316.

### DIFF
--- a/core/src/test/java/edu/wpi/grip/core/serialization/CompatibilityTest.java
+++ b/core/src/test/java/edu/wpi/grip/core/serialization/CompatibilityTest.java
@@ -3,12 +3,10 @@ package edu.wpi.grip.core.serialization;
 import com.google.common.eventbus.EventBus;
 import com.google.inject.Guice;
 import com.google.inject.Injector;
-import com.google.inject.Key;
-import com.google.inject.TypeLiteral;
-import edu.wpi.grip.core.*;
+import edu.wpi.grip.core.GRIPCoreModule;
+import edu.wpi.grip.core.Pipeline;
 import edu.wpi.grip.core.operations.Operations;
 import edu.wpi.grip.core.settings.ProjectSettings;
-import edu.wpi.grip.core.sources.ImageFileSource;
 import edu.wpi.grip.generated.CVOperations;
 import edu.wpi.grip.util.Files;
 import org.junit.Before;
@@ -16,29 +14,22 @@ import org.junit.Test;
 
 import java.io.*;
 import java.net.URI;
-import java.util.List;
-import java.util.Optional;
 
 import static junit.framework.TestCase.assertEquals;
 
 /**
  * This tests for backwards compatibility by opening a project file containing ALL the steps of GRIP
- * at the time of this file's creation.
+ * at the time of this file's creation. Please note that this test requires a webcam at position 0.
  */
 public class CompatibilityTest {
 
     private static final URI testphotoURI = Files.testphotoURI; //The location of the photo source for the test
     private static final URI testprojectURI = Files.testprojectURI; //The location of the save file for the test
 
-    private Connection.Factory<Object> connectionFactory;
-    private ImageFileSource.Factory imageSourceFactory;
-    private Step.Factory stepFactory;
     private Pipeline pipeline;
     private Project project;
     private ProjectSettings settings;
 
-    private List<Operation> operationList;
-    private Optional<Throwable> throwableOptional;
     private EventBus eventBus;
 
     @Before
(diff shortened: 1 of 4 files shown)


ThomasJClark:
We probably shouldn't include another ~10 MB image in the repository.  You can just use one of the much smaller sample images we have.

Also, can you give this PR an actual name so we can keep track of it?

core/src/test/java/edu/wpi/grip/core/serialization/CompatibilityTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+        Reader temp = new FileReader(file);
+        BufferedReader reader = new BufferedReader(temp);
+        String line = "", oldtext = "";
+        while ((line = reader.readLine()) != null) {
+            oldtext += line + "\r\n";
+        }
+        reader.close();
+        String newtext = oldtext.replaceAll("REPLACEME", photoFileName);//This gives the correct location of the test photo needed to the project file
+        //The following alters the project settings in the project file to the defaults given in the "settings" variable
```
ThomasJClark: Why?

JLLeitschuh: Because our save files are relative and need to be resolved differently depending upon the OS that we are working with.
This wouldn't be a problem if we were creating save file bundles.

core/src/test/java/edu/wpi/grip/core/serialization/CompatibilityTest.java
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * This tests for backwards compatibility by opening a project file containing ALL the steps of GRIP
+ * at the time of this file's creation.
+ */
+public class CompatibilityTest {
+
+    private static final URI testphotoURI = Files.testphotoURI; //The location of the photo source for the test
+    private static final URI testprojectURI = Files.testprojectURI; //The location of the save file for the test
+
+    private Connection.Factory<Object> connectionFactory;
```
JLLeitschuh: You don't need this

core/src/test/java/edu/wpi/grip/core/serialization/CompatibilityTest.java
```diff
@@ -0,0 +1,102 @@
...
+    private EventBus eventBus;
+
+    @Before
+    public void setUp() throws Exception {
+
+        //Set up the stuff we need for the core functionality for GRIP
+        final Injector injector = Guice.createInjector(new GRIPCoreModule());
+        connectionFactory = injector
+                .getInstance(Key.get(new TypeLiteral<Connection.Factory<Object>>() {
+                }));
```
JLLeitschuh: You don't need this

core/src/test/java/edu/wpi/grip/core/serialization/CompatibilityTest.java
```diff
@@ -0,0 +1,102 @@
...
+        imageSourceFactory = injector
+                .getInstance(ImageFileSource.Factory.class);
+        eventBus = injector.getInstance(EventBus.class);
+        pipeline = injector.getInstance(Pipeline.class);
+        project = injector.getInstance(Project.class);
+        settings = injector.getInstance(ProjectSettings.class);
+        stepFactory = injector.getInstance(Step.Factory.class);
```
JLLeitschuh: You don't need this

core/src/test/java/edu/wpi/grip/core/serialization/CompatibilityTest.java
```diff
@@ -0,0 +1,102 @@
...
+ * at the time of this file's creation.
+ */
+public class CompatibilityTest {
+
+    private static final URI testphotoURI = Files.testphotoURI; //The location of the photo source for the test
+    private static final URI testprojectURI = Files.testprojectURI; //The location of the save file for the test
+
+    private Connection.Factory<Object> connectionFactory;
+    private ImageFileSource.Factory imageSourceFactory;
+    private Step.Factory stepFactory;
```
JLLeitschuh: You don't need this

core/src/test/java/edu/wpi/grip/core/serialization/CompatibilityTest.java
PaulaRudy: I just wanted to make sure the path names and such weren't going to be an issue by using the defaults at the time of its run. It's a "just in case" thing. I can get rid of this if you're sure it won't be needed...?

PaulaRudy:
I tuned the project steps to this specific file to give good information-at-a-glance when loaded (IE, thresholds tuned to isolate specific areas and such- it makes a good demonstration of all the steps as well as a good test file). I could use a different file, but I'd rather not loose the tuning - would it be alright if I used a smaller version instead?  I could always just shrink it to a more manageable size.

core/src/test/java/edu/wpi/grip/core/serialization/CompatibilityTest.java
JLLeitschuh: We want to make sure that deserializing project settings won't break across versions. If an old projects settings can't be deserialized then we have a problem and we want to know about it.

PR #451, opened 2016-01-28
Add ZenHub.io Badge to readme

JLLeitschuh:
This tool is awesome. More people should know about it.
@ThomasJClark I know you're not completely sold on this plugin yet but I'm just going to leave this here until you are convinced of its awesomeness.


ThomasJClark:
It seems pretty cool, but I just want to give it a trial run before we add the badge. At this point, we literally just found out about it, so it seems like that would be jumping the gun a little

JLLeitschuh:
Thats fine. We can leave this PR open.

PR #454, opened 2016-01-29
Add JVM args and more options to the deploy dialog

### LINKED ISSUE #425: Roborio running out of memory
Our roboRIO is throwing this error: 

```
Loading Dependency Injection Framework
Feb 27, 2015 2:09:45 AM java.util.logging.LogManager$RootLogger log
CONFIG: Configuration done.
Feb 27, 2015 2:09:45 AM java.util.logging.LogManager$RootLogger log
CONFIG: GRIP Version: 1.1.1
platform: /Linux/arm/
Feb 27, 2015 2:09:50 AM edu.wpi.grip.core.Main start
INFO: Loading file /home/lvuser/project.grip
Java HotSpot(TM) Embedded Client VM warning: INFO: os::commit_memory(0xb48c2000, 32768, 1) failed; error='Cannot allocate memory' (errno=12)

 There is insufficient memory for the Java Runtime Environment to continue.
 Native memory allocation (mmap) failed to map 32768 bytes for committing reserved memory.
 An error report file with more information is saved as:
 /home/lvuser/hs_err_pid1963.log

 Compiler replay data is saved as:
 /home/lvuser/replay_pid1963.log
```

Anyone else run into this problem? Or any ideas for fixes?

ThomasJClark:
JVM arguments can be specified, and some reasonable defaults are included for new projects (and old projects created before this change).  These settings should help reduce OOM errors, and help us debug new ones.

Running from a robot program is now also easier, since a new wrapper script is included to facilitate the JVM options.

```c++
void RobotInit() override {
    system("/home/lvuser/grip &");
}
```

In addition, the resulting file name of the deployed project can be changed from the default of "project.grip".  This lets you upload multiple versions to the roboRIO, and then the robot code can select one of them at the start of a match, although the robot code to run the project might be a little more complicated in this case.

We also now check the MD5 sum of the JAR before deploying it.  If it matches the JAR we're about to upload, we just skip it.  This makes deploys after the first one a LOT faster.

And other mostly cosmetic deploying changes, including:

* A two-column layout
* Pressing "enter" anywhere in the dialog makes the deploy start
* The command that will run on the roboRIO is shown in the dialog

Closes #425

### DIFF
--- a/core/src/main/java/edu/wpi/grip/core/settings/ProjectSettings.java
+++ b/core/src/main/java/edu/wpi/grip/core/settings/ProjectSettings.java
@@ -31,6 +31,9 @@ public class ProjectSettings implements Cloneable {
     @Setting(label = "Deploy Java home", description = "Where Java is installed on the robot.")
     private String deployJavaHome = "/usr/local/frc/JRE/";
 
+    @Setting(label = "Deploy JVM options", description = "Command line options passed to the roboRIO JVM")
+    private String deployJvmOptions = "-Xmx50m";
+
     /**
      * Set the FRC team number.  If the deploy address and NetworkTables server address haven't been manually
      * overridden, this also changes them to the mDNS hostname of the team's roboRIO.
@@ -98,6 +101,14 @@ public class ProjectSettings implements Cloneable {
         if (deployJavaHome != null) this.deployJavaHome = deployJavaHome;
     }
 
+    public String getDeployJvmOptions() {
+        return deployJvmOptions;
+    }
+
+    public void setDeployJvmOptions(String deployJvmOptions) {
+        this.deployJvmOptions = deployJvmOptions;
+    }
+
     private String computeFRCAddress(int teamNumber) {
         return "roborio-" + teamNumber + "-frc.local";
     }
@@ -109,6 +120,7 @@ public class ProjectSettings implements Cloneable {
                 .add("deployDir", deployDir)
                 .add("deployUser", deployUser)
                 .add("deployJavaHome", deployJavaHome)
+                .add("deployJvmOptions", deployJvmOptions)
                 .add("publishAddress", publishAddress)
                 .add("teamNumber", teamNumber)
                 .toString();
--- a/core/src/main/java/edu/wpi/grip/core/settings/ProjectSettings.java
+++ b/core/src/main/java/edu/wpi/grip/core/settings/ProjectSettings.java
@@ -106,7 +106,7 @@ public class ProjectSettings implements Cloneable {
     }
 
     public void setDeployJvmOptions(String deployJvmOptions) {
-        this.deployJvmOptions = deployJvmOptions;
+        if (deployJvmOptions != null) this.deployJvmOptions = deployJvmOptions;
     }
 
     private String computeFRCAddress(int teamNumber) {
--- a/core/src/main/java/edu/wpi/grip/core/settings/ProjectSettings.java
+++ b/core/src/main/java/edu/wpi/grip/core/settings/ProjectSettings.java
@@ -32,7 +32,7 @@ public class ProjectSettings implements Cloneable {
     private String deployJavaHome = "/usr/local/frc/JRE/";
 
     @Setting(label = "Deploy JVM options", description = "Command line options passed to the roboRIO JVM")
-    private String deployJvmOptions = "-Xmx50m";
+    private String deployJvmOptions = "-Xmx50m -XX:-OmitStackTraceInFastThrow -XX:+HeapDumpOnOutOfMemoryError";
 
     /**
      * Set the FRC team number.  If the deploy address and NetworkTables server address haven't been manually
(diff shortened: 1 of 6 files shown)


core/src/main/java/edu/wpi/grip/core/settings/ProjectSettings.java
```diff
@@ -98,6 +101,14 @@ public void setDeployJavaHome(String deployJavaHome) {
         if (deployJavaHome != null) this.deployJavaHome = deployJavaHome;
     }
 
+    public String getDeployJvmOptions() {
+        return deployJvmOptions;
+    }
+
+    public void setDeployJvmOptions(String deployJvmOptions) {
+        this.deployJvmOptions = deployJvmOptions;
```
JLLeitschuh: This needs to check that the `deployJvmOptions` aren't null.
See the above examples.

ThomasJClark: fixed

JLLeitschuh:
Do you want to add the ability to copy and paste your command manually? Currently that text box doesn't allow you to do so.

Also, we should add the JVM arguments present at runtime to the logger so that we know what they are when trying to reproduce bugs.
http://stackoverflow.com/a/1518250/3708426

ThomasJClark:
>  Also, we should add the JVM arguments present at runtime to the logger so that we know what they are when trying to reproduce bugs.

We already log them with this PR. Doing what that SO post says would actually just log them on the RoboRIO, which means if you close the dialog or whatever, you have to use ftp to get them

JLLeitschuh:
But if someone has a problem with GRIP running on a robot then we will want the log on the robot to also have the JVM args.

Also, shouldn't we be SCPing a bash script instead of just creating a larger and larger run command? Otherwise the JVM args won't work for people trying to start GRIP with their user programs.

https://github.com/WPIRoboticsProjects/GRIP/wiki/Tutorial:-Run-GRIP-from-a-CPP-or-Java-FRC-program

ThomasJClark:
> Also, shouldn't we be SCPing a bash script instead of just creating a larger and larger run command? Otherwise the JVM args won't work for people trying to start GRIP with their user programs.

True.  Maybe the bash script can also have an optional argument for the project file.  That way, people can upload multiple projects without changing their robot program.  If they want to, they could still choose a project dynamically based on a preference or something.

ui/src/main/java/edu/wpi/grip/ui/DeployController.java
```diff
@@ -150,9 +157,12 @@ private void deploy(String address, String user, String password, String javaHom
             StringWriter projectWriter = new StringWriter();
             project.save(projectWriter);
 
-            // Upload the GRIP core JAR and the serialized project to the robot
-            scp.upload(new StringInMemoryFile(PROJECT_FILE, projectWriter.toString()), deployDir + "/");
-            scp.upload(new FileSystemFile(LOCAL_GRIP_JAR), deployDir + "/" + GRIP_JAR);
+            // Upload the GRIP core JAR, a wrapper script to run it, and the serialized project to the robot
+            final String commandStr = command.get();
```
JLLeitschuh: Don't you need to escape this?

JLLeitschuh:
Add the JVM arguments to the log file too.

ui/src/main/java/edu/wpi/grip/ui/DeployController.java
ThomasJClark: Nope (I tested it with spaces in the path)

ThomasJClark:
> Add the JVM arguments to the log file too.

I did.  Look at your GRIP.log.

ui/src/main/java/edu/wpi/grip/ui/DeployController.java
JLLeitschuh: No, what about `"` in the arguments.

ThomasJClark: Matched quotes are fine, since it's executed as a bash command.  We need to allow quotes so you can provide an argument with spaces in it.

JLLeitschuh:
I meant here:
https://github.com/WPIRoboticsProjects/GRIP/blob/master/core/src/main/java/edu/wpi/grip/core/GRIPCoreModule.java#L71

ThomasJClark:
That would only show it on the robot.  The JVM args are already on the robot, (since they're deployed in a shell script), and they're also already shown in the deploy console.  The only place they wouldn't already be saved is the local GRIP.log, but they will now after this change.

JLLeitschuh:
I think knowing what arguments grip is running under on a specific device is important.
This is a simple change that will give us more information. I don't want to have to ask users to send us both log files in order to figure out what problems they have experienced with deploying.

ThomasJClark:
> I don't want to have to ask users to send us both log files in order to figure out what problems they have experienced with deploying.

This makes no sense.  We'll have to look at people's project files anyways, which contain the JVM arguments.  And in most situations, they would give us the log file from their computer anyway, since the roboRIO one is harder to get and they already have the log output from the deploy dialog.

Also, this only matters for people who change the default deploy options.  If someone does that, they can tell us what they changed it to.

PR #468, opened 2016-02-03
Fix NTPublishOperation problems

### LINKED ISSUE #432: ContoursReport publishing incorrect NetworkTable keys
ContoursReport publishing user selected values "area" and "height" creates NetworkTables keys "height" and "centerY" (and once I saw "solidity" and "height".

Fields are passed "positionally" and have different names in the User Interface and the NetworkTables.  I didn't check for any other instances of this type of error.

Portion of .grip file deployed to the roboRIO:
```xml
   < grip:Step name="Publish ContoursReport">
   < /grip:Input>
      < grip:Input step="3" socket="5">
        < value>true</value>
      < /grip:Input>
      < grip:Input step="3" socket="6">
        < value>true</value>
```

ThomasJClark:
This fixes two problems:
- Publishable fields now much have a "weight".  This determines the
  order that the checkboxes are in, preventing issues with different JVM
  versions returning methods in different orders.
- If a publishable value has no (sub) key, it's put directly in the key
  for the entire step.  This was sort of the behavior before, but
  previously NetworkTables automatically put a trailing slash at the
  end.  OutlineViewer still shows it the same, but robot programs did
  not work unless they included this trailing slash.

Closes #432

### DIFF
--- a/core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
+++ b/core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
@@ -1,6 +1,8 @@
 package edu.wpi.grip.core.operations.networktables;
 
 import com.google.common.base.Throwables;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Ordering;
 import com.google.common.eventbus.EventBus;
 import edu.wpi.first.wpilibj.networktables.NetworkTable;
 import edu.wpi.first.wpilibj.tables.ITable;
@@ -9,8 +11,8 @@ import edu.wpi.grip.core.*;
 import java.io.InputStream;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
-import java.util.ArrayList;
-import java.util.List;
+import java.util.Arrays;
+import java.util.Comparator;
 import java.util.Optional;
 import java.util.function.Function;
 
@@ -26,7 +28,7 @@ public class NTPublishOperation<S, T extends NTPublishable> implements Operation
 
     private final Class<S> type;
     private final Function<S, T> converter;
-    private final List<Method> ntValueMethods = new ArrayList<>();
+    private final ImmutableList<Method> ntValueMethods;
 
     /**
      * Create a new publish operation for a socket type that implements {@link NTPublishable} directly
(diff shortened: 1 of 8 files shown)


core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
```diff
@@ -59,6 +60,10 @@ public NTPublishOperation(Class<S> socketType, Class<T> reportType, Function<S,
                 ntValueMethods.add(method);
             }
         }
+
+        // Sort the accessors by their weight. As long as they all have unique weights, this lets us avoid the issue
+        // of different JVMs returning methods in different orders.
+        ntValueMethods.sort(Comparator.comparing(method -> method.getAnnotation(NTValue.class).weight()));
```
JLLeitschuh: Throw an exception if there is duplicate weights and write a test that ensures that this happens.

JLLeitschuh:
Can you add a test with a mock NTPublishable value that this works as expected?
Ensure that if you have duplicate weights then you get an `IllegalArgumentException`. Also, can you test that the right booleans cause the right methods to be published.

core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
ThomasJClark: > Also, can you store the `ntValueMethods` in an immutable list? You get some performance boosts out of immutable lists.

ok, fixed

core/src/main/java/edu/wpi/grip/core/operations/networktables/NTPublishOperation.java
```diff
@@ -49,16 +50,20 @@ public NTPublishOperation(Class<S> socketType, Class<T> reportType, Function<S,
...
-        // Any accessor method with an @NTValue annotation can be published to NetworkTables.
-        for (Method method : reportType.getDeclaredMethods()) {
-            if (method.getAnnotation(NTValue.class) != null) {
-                if (method.getParameters().length > 0) {
+        // Any accessor method with an @NTValue annotation can be published to NetworkTables.  We sort them by their
+        // "weights" in order to avoid the issue of different JVM versions returning methods in a different order.
+        this.ntValueMethods = ImmutableList.copyOf(Arrays.asList(reportType.getDeclaredMethods()).stream()
+                .filter(method -> method.getAnnotation(NTValue.class) != null)
+                .sorted(Comparator.comparing(method -> method.getAnnotation(NTValue.class).weight()))
+                .iterator());
```
JLLeitschuh: If there are two methods with the same weight throw an `IllegalArgumentException`

core/src/test/java/edu/wpi/grip/core/operations/networktables/NTPublishOperationTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testNonDistinctWeights() {
+        new NTPublishOperation<>(ReportWithNonDistinctWeights.class);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testInvalidParameters() {
+        new NTPublishOperation<>(ReportWithParameters.class);
+    }
```
JLLeitschuh: Add a test that `getInputSockets` returned the expected number of sockets in the expected order.

JLLeitschuh:
Add an estimate to #432 and I'll merge this.

ThomasJClark:
k

PR #472, opened 2016-02-05
Cleans up most of FindBugs Warnings

JLLeitschuh:
- There were several performant issues.
 - Now we use UTF-8 file encoding for all save writing/reading/uploading


core/src/main/java/edu/wpi/grip/core/serialization/Project.java
```diff
@@ -17,13 +19,10 @@
  */
 @Singleton
 public class Project {
+    public static final Charset PROJECT_CHARSET = StandardCharsets.UTF_8;
```
JLLeitschuh: Don't make this a constant

PR #475, opened 2016-02-06
Adds Sanity Tests to Most Core Packages

JLLeitschuh:


core/src/main/java/edu/wpi/grip/core/Palette.java
```diff
@@ -42,7 +43,7 @@ public void onOperationAdded(OperationAddedEvent event) {
     /**
      * @return The operation with the specified unique name
      */
-    public Optional<Operation> getOperationByName(String name) {
+    public Optional<Operation> getOperationByName(@Nullable String name) {
```
JLLeitschuh: Convert to check not null

PR #498, opened 2016-02-12
Add command line parsing in GUI mode and headless Windows support

### LINKED ISSUE #492: GRIP should take command line options in GUI mode
We're attempting to run grip on our driver's station. We have the basics working, but we have to open the .grip file from the menus every time we start grip. I cannot find documentation on how to automatically load the file and our attempts to guess at command-line parameters that might do so have not worked.

### LINKED ISSUE #493: GRIP shouldn't use respond to HUP signal
GRIP now kills existing versions of GRIP before it runs, so there's not need to close when we get a hangup.  This prevents headless mode from working on non-unix systems, since SIGHUP is a posix thing.

ThomasJClark:
In GUI mode, a file name can be specified to open, and --headless can be
added to the command to make it not really open the GUI.

In headless mode, we don't really need to listen for hangup signals
(since the wrapper script will kill any old instances of GRIP), so we
can remove that to make it work on non-unix systems.

Closes #493
Closes #492

### DIFF
--- a/core/src/main/java/edu/wpi/grip/core/Main.java
+++ b/core/src/main/java/edu/wpi/grip/core/Main.java
@@ -8,9 +8,7 @@ import edu.wpi.grip.core.events.ExceptionClearedEvent;
 import edu.wpi.grip.core.events.ExceptionEvent;
 import edu.wpi.grip.core.operations.Operations;
 import edu.wpi.grip.core.serialization.Project;
-import edu.wpi.grip.core.util.SafeShutdown;
 import edu.wpi.grip.generated.CVOperations;
-import sun.misc.Signal;
 
 import javax.inject.Inject;
 import java.io.File;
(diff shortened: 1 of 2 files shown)


core/src/main/java/edu/wpi/grip/core/Main.java
```diff
@@ -23,22 +21,13 @@
...
+    @Inject private EventBus eventBus;
+    @Inject private Logger logger;
 
     @SuppressWarnings("PMD.SystemPrintln")
     public static void main(String[] args) throws IOException, InterruptedException {
-        // Close GRIP when we get SIGHUP.  This signal is sent, for example, when GRIP is run in an SSH session
-        // and the session is closed.
-        Signal.handle(new Signal("HUP"), signal -> SafeShutdown.exit(0));
-
-        System.out.println("Loading Dependency Injection Framework");
```
JLLeitschuh: If you're going to remove this put something in its place instead.

core/src/main/java/edu/wpi/grip/core/Main.java
```diff
@@ -23,22 +21,13 @@
...
+    @Inject private Project project;
+    @Inject private PipelineRunner pipelineRunner;
+    @Inject private EventBus eventBus;
+    @Inject private Logger logger;
 
     @SuppressWarnings("PMD.SystemPrintln")
     public static void main(String[] args) throws IOException, InterruptedException {
-        // Close GRIP when we get SIGHUP.  This signal is sent, for example, when GRIP is run in an SSH session
-        // and the session is closed.
-        Signal.handle(new Signal("HUP"), signal -> SafeShutdown.exit(0));
```
JLLeitschuh: Now GRIP on a remote device will continue to publish after you disconnect.
The deployed script should respond to the hangup instead.

ThomasJClark: Why is that bad? If you're testing robot code and deploying new GRIP stuff, I don't understand why you would want it to stop when you're disconnected. The reason we used to do that was to avoid letting multiple GRIPs run at the same time, but that's not a problem anymore with the wrapper script

PR #504, opened 2016-02-15
When dragging, visually disable sockets that can't be dragged onto

ThomasJClark:
This should make it a little easier to figure stuff out by trial-and-error.  Basically, we used to check to make sure you didn't connect to sockets that weren't compatible (for example, they're different types, or one of them was already connected to something), but there was very little visual feedback to show what connections would be possible.

To do this, I replaced the `Optional<Socket>` that stores the socket being dragged with an `ObjectProperty<Socket>`.  This works nicely with JavaFX's event system and lets us have all sockets react when one of them is dragged.  We could also use a `ObjectProperty<Optional<Socket<?>>>`, but that would make any practical usage really verbose and hard to read.


JLLeitschuh:
It's still kind of subtle. Do you want to make the ones that can connect "glow" or something.

ThomasJClark:
I like it like this, and I'm not sure how to make something glow with JavaFX

PR #505, opened 2016-02-15
Start and stop headless GRIP using NetworkTables

### LINKED ISSUE #501: Feature Request: Be able to start and stop pipeline from within program
Being able to start and stop the pipeline from inside an FRC program would help the program save resources and run GRIP only when actually needed.

ThomasJClark:
This lets robot programs start and stop GRIP to save CPU without having
to wait for the process to start up again.

For example,

```java
NetworkTable table = NetworkTable.getTable("GRIP");

table.putBoolean("run", false); // Stop GRIP without killing the process

doSomeStuffThatDoesntNeedVision();

table.putBoolean("run", true); // Start GRIP again
```

This should be useful for teams that want to do auto alignment in teleop, since they could start and stop vision processing whenever the driver presses the button to auto aim.

Closes #501

### DIFF
--- a/core/src/main/java/edu/wpi/grip/core/GRIPCoreModule.java
+++ b/core/src/main/java/edu/wpi/grip/core/GRIPCoreModule.java
@@ -10,11 +10,13 @@ import com.google.inject.spi.InjectionListener;
 import com.google.inject.spi.TypeEncounter;
 import com.google.inject.spi.TypeListener;
 import edu.wpi.grip.core.events.UnexpectedThrowableEvent;
+import edu.wpi.grip.core.operations.networktables.NTManager;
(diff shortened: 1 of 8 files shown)


JLLeitschuh:
If I have grip on a roborio and running via deploy then this will start both pipelines which will create unexpected (and really hard to diagnose) issues.

ThomasJClark:
It'll only start both of them if they're both connected to the same NetworkTables server, which is probably a bad idea anyways 

We maybe  make this only work on headless mode. I doubt anyone would need to use it in GUI mode

JLLeitschuh:
This needs to work with the new `--headless` parameter.

PR #527, opened 2016-02-25
GRIP publish as a ROS Node

JLLeitschuh:
Makes it possible to publish GRIP data to ROS topics


core/src/main/java/edu/wpi/grip/core/Operation.java
```diff
@@ -24,6 +25,10 @@
      */
     String getName();
 
+    default ImmutableSet<String> getAliases() {
```
ThomasJClark: Document this.  It should probably also just be an annotation

core/src/main/java/edu/wpi/grip/core/Palette.java
```diff
@@ -31,7 +32,21 @@
...
+        }
+    }
+
+    /**
+     * Maps the key to the given operation
+     * @param key The key the operation should be mapped to
+     * @param operation The operation to map the key to
+     * @throws IllegalArgumentException if the key is already in the {@link #operations} map.
+     */
+    private void map(String key, Operation operation) {
```
ThomasJClark: Get rid of this method

core/src/main/java/edu/wpi/grip/core/operations/Operations.java
```diff
@@ -1,47 +1,78 @@
...
+    private final EventBus eventBus;
+    private final ImmutableList<Supplier<Operation>> operations;
 
-    public static void addOperations(EventBus eventBus) {
+    @Inject
+    Operations(EventBus eventBus, NTManager ntManager, ROSManager rosManager) {
+        this.eventBus = checkNotNull(eventBus, "EventBus cannot be null");
+        checkNotNull(ntManager, "ntManager cannot be null");
+        checkNotNull(rosManager, "rosManager cannot be null");
+        this.operations = ImmutableList.of(
```
ThomasJClark: Why

core/src/main/java/edu/wpi/grip/core/operations/network/Manager.java
```diff
@@ -0,0 +1,11 @@
+package edu.wpi.grip.core.operations.network;
+
+
+import java.util.Set;
+
+/**
+ *
+ */
+public interface Manager {
```
ThomasJClark: What?

core/src/main/java/edu/wpi/grip/core/operations/network/NetworkPublisher.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.Set;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+/**
+ * Manages the interface between the {@link PublishOperation} and the actual network
+ * protocol implemented by a specific {@link Manager}.
+ * <p>
+ * This class is designed to be in one of two states, either the set of keys
+ * will be empty. In which case the name will be used as the publish key.
```
ThomasJClark: Grammar

core/src/main/java/edu/wpi/grip/core/operations/network/networktables/RealNTManager.java
```diff
@@ -24,7 +29,7 @@
  * This class encapsulates the way we map various settings to the global NetworkTables state.
  */
 @Singleton
-public class NTManager {
+public class RealNTManager implements NTManager {
```
ThomasJClark: No.  Just call this `NTManager`.

core/src/main/java/edu/wpi/grip/core/operations/network/ros/RealROSManager.java
```diff
@@ -0,0 +1,184 @@
...
+
+import java.util.*;
+import java.util.concurrent.ConcurrentMap;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ *
+ */
+public class RealROSManager implements ROSManager {
```
ThomasJClark: Just make this `ROSManager`.  Don't pollute the code with ridiculous classes for the sake of working around weird testing methods.

core/src/main/java/edu/wpi/grip/core/Operation.java
JLLeitschuh: I like this better because it's more consistent.
I'll document it though.

core/src/main/java/edu/wpi/grip/core/Palette.java
JLLeitschuh: Why and how? Put the same code twice?
If you want it copy pasted twice I'm not doing that.

core/src/main/java/edu/wpi/grip/core/operations/Operations.java
JLLeitschuh: Because `eventBus.post(new OperationAddedEvent(.....` a million times was just getting ridiculous.
Its duplicate code and was due for a refactor.

core/src/main/java/edu/wpi/grip/core/operations/network/Manager.java
JLLeitschuh: Forgot to document. Good call.

core/src/main/java/edu/wpi/grip/core/operations/network/Publishable.java
```diff
@@ -0,0 +1,9 @@
+package edu.wpi.grip.core.operations.network;
+
+/**
+ * Interface for classes that can be published with a NetworkOperation
```
JLLeitschuh: with a NetworkPublisher

core/src/main/java/edu/wpi/grip/core/operations/network/ros/ROSPublishOperation.java
```diff
@@ -0,0 +1,33 @@
+package edu.wpi.grip.core.operations.network.ros;
+
+
+import edu.wpi.grip.core.operations.network.PublishOperation;
+import edu.wpi.grip.core.operations.network.Publishable;
+
+import java.util.function.Function;
+
+public class ROSPublishOperation<S, T extends Publishable> extends PublishOperation<S, T> {
```
JLLeitschuh: Document this

core/src/test/java/edu/wpi/grip/core/operations/OperationsTest.java
```diff
@@ -57,7 +59,7 @@ public void testCreateAllCVSteps() {
 
     @Test
     public void testCreateAllCoreSteps() {
-        Operations.addOperations(eventBus);
+        new Operations(eventBus, new RealNTManager(), new RealROSManager()).addOperations();
```
JLLeitschuh: This should be using the mock ones.

core/src/main/java/edu/wpi/grip/core/Palette.java
ThomasJClark: > Why and how? Put the same code twice?

Yes.  It's literally just two lines of code.  You're actually using *more* lines to separate it into a separate method than it would take to just call `operations.put` twice.

core/src/main/java/edu/wpi/grip/core/operations/Operations.java
ThomasJClark: Using method references and streams for something so simple is even more ridiculous.

Image if someone wants to read this code and find out about how operations work. `eventBus.post(new OperationAddedEvent(new ResizeOperation()));` makes it perfectly clear.  This just seems like you're trying to show off your knowledge of Java 8 features.  Like, why would you even use suppliers instead of instances of the class?

core/src/main/java/edu/wpi/grip/core/operations/network/ros/RealROSManager.java
JLLeitschuh: So then `ROSPublishOperation` should just take a Manager interface.
I did this so I could do class binding in the Guice module. I guess I can use named bindings instead.
In my mind, that screamed "not type safe" but if you'd rather me use named injection bindings I'll do that instead.

core/src/main/java/edu/wpi/grip/core/operations/Operations.java
JLLeitschuh: Because every time you call addOperations you need to create new instances of the class.
That's the design we've been using all along.
If you look at the old version thats what we have been doing already I assumed that we always wanted to create new instances each time `addOperation` was called.

And, I'm not "showing off" I'm taking advantage of the language features. This is perfectly clear java code.

core/src/main/java/edu/wpi/grip/core/Palette.java
JLLeitschuh: :-1: Wontfix

core/src/main/java/edu/wpi/grip/core/Operation.java
JLLeitschuh: Actually, it's impossible to do this with an annotation for the NTPublish operation.
This is why: https://github.com/WPIRoboticsProjects/GRIP/pull/527/files#diff-108198640a1078f1b20a004632a786dbR45

PR #530, opened 2016-03-01
mjpeg reading format works for both Trendnet Cameras and Axis-cameras

BBBBlarry:
mjpeg reading format works for both Trendnet Cameras and Axis-cameras


JLLeitschuh:
Can you merge/rebase your changes with master? We can't merge this currently.

Issue #441, opened 2016-01-26
Add Advanced Filter Contours capabilities.


ThadHouse:
This could easily be a 2nd Filter Contours operation command, but it would be nice to have the following filters.

* Vertex Count (Allows easy filtering of rectangles)
* Force Convex (Allows easy filtering of odd shapes)
* Ratio filter (Check height to width ratio, and filter if out of range)
* Parallelism filter (Check how many lines are vertical and horizontal)

With these 4 added, it would make it much easier to force only the target to be seen and filter in GRIP, when we have more information.

JLLeitschuh:
We welcome the addition of more operations.
Perhaps we should create a wiki topic about how to write/add operation PR's

AustinShalit:
This issue was solved with #531.  If you think this is not the case, please reopen this issue.

PR #531, opened 2016-03-02
Add AdvancedFilterCountoursOperation

drbobbob:
Fixes issue #441


drbobbob:
I would like to write tests for this feature but I am not quite sure where to put them.

ThomasJClark:
Looks good.  Do you wanna just add these parameters to the regular FilterContoursOperation?

drbobbob:
I certainly can.  Just let me know what you would like me to do.

ThomasJClark:
Nice!  Just a few notes:

- "Min Vertex Count" and "Max Vertex Count" get cut off because they're too long.  They could be renamed to "Min Vertices" and "Max Verticies"
- Max vertices should have a much higher initial value (like 1000000), so it shouldn't actually filter anything by default.
- "Force convex" can be removed, since solidity does the same thing (100% solidity === convex)

PR #548, opened 2016-03-20
Fixed Max Vertices defaulting to zero and displaying 1,000,000

### LINKED ISSUE #538: 0 contours found: Max Vertices Problem
I am going through the tutorial for the 2016 game, but when I get to the step where I have to filter contours, I get 0 results each time, no matter what the values I put in there. I even put in 1 as the minimum area and still got nothing

Sam-Carlberg:
Fixes #538. The default value and minimum value were switched.


JLLeitschuh:
Awesome. Thanks!

PR #551, opened 2016-03-23
Add HTTP publishing and image source operations

SamCarlberg:
## How to use the publishing operations
Use them in the UI just like a NetworkTable publish operation

#### Getting the data
All data published will be at `http://localhost:8080/GRIP/data` (or whatever port you set in the settings) in json format. To get specific data, put their names as parameters in the URL -- e.g. if you want `myContoursReport` and `myBlobsReport`, go to `http://localhost:8080/GRIP/data?myContoursReport&myBlobsReport`

#### GRIP-side
Just use it like a normal image source. The URL is configurable under `/GRIP/upload/image/` -- e.g. `/GRIP/upload/image/my-image`. Note that this path _must_ be under `/GRIP/upload/image/`

#### Supplier-side
Perform an HTTP `POST` operation to `/GRIP/upload/image/your-image` with the raw image bytes as binary data. It will be reconstructed server-side and trigger the pipeline to run if the image is different form the last one POSTed to that path.


buildSrc/build.gradle
```diff
@@ -12,7 +12,7 @@ configurations {
 dependencies {
     compile gradleApi()
     compile localGroovy()
-    compile group: 'com.google.guava', name: 'guava', version: '19.0'
+    compile group: 'com.google.guava', name: 'guava', version: '18.0'
```
JLLeitschuh: Why?

build.gradle
```diff
@@ -189,6 +189,10 @@ project(":core") {
         compile group: 'org.apache.commons', name: 'commons-lang3', version: '3.4'
         compile group: 'com.google.guava', name: 'guava', version: '19.0'
         compile group: 'com.google.auto.value', name: 'auto-value', version: '1.1'
+        compile group: 'com.google.code.gson', name:'gson', version:'2.6.2'
+        testCompile 'org.apache.httpcomponents:httpclient:4.5.2'
+        testCompile 'org.apache.httpcomponents:httpcore:4.4.4'
+        testCompile 'org.apache.httpcomponents:httpmime:4.5.2'
```
JLLeitschuh: Can you use the format with group: name: version: ect...

core/src/main/java/edu/wpi/grip/core/operations/network/http/HttpManager.java
```diff
@@ -0,0 +1,66 @@
...
+import edu.wpi.grip.core.operations.network.Manager;
+import edu.wpi.grip.core.operations.network.MapNetworkPublisher;
+import edu.wpi.grip.core.operations.network.MapNetworkPublisherFactory;
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ */
```
JLLeitschuh: Document this a bit better.

buildSrc/build.gradle
SamCarlberg: I'm not sure why that's there; I had pulled from WPIRoboticsProjects/master so this may have gotten changed

core/src/main/java/edu/wpi/grip/core/operations/network/http/HttpManager.java
SamCarlberg: Documented in 438cf2f

build.gradle
SamCarlberg: Fixed in 438cf2f

SamCarlberg:
A few test files got deleted for some reason, I'll revert that

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
```diff
@@ -0,0 +1,399 @@
...
+            } catch (IOException ex) {
+                throw new GripException(ex);
+            }
+        }
+    }
+
+    private final HttpServerFactory serverFactory;
+    private HttpServer server;
+    private final Map<String, GetHandler> getHandlers;
+    private final Map<String, List<PostHandler>> postHandlers;
```
JLLeitschuh: You could have used a  multi-map for this: https://github.com/google/guava/wiki/NewCollectionTypesExplained#multimap

SamCarlberg: Not much point, it would be used exactly the same.

core/src/main/java/edu/wpi/grip/core/util/Holder.java
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Sets the value of this holder. This is a thread-safe operation.
+     *
+     * @param value the new value of this holder.
+     */
+    public synchronized void set(T value) {
+        if (value == null) {
+            throw new NullPointerException("Held value cannot be null");
+        }
```
JLLeitschuh: This can be replaced with the google preconditions check.
https://github.com/google/guava/wiki/PreconditionsExplained

This whole method could become:
```java
public synchronized void set(T value) {
    this.value = checkNotNull(value, "Held value cannot be null");
}
```

core/src/test/java/edu/wpi/grip/core/http/GripServerTest.java
```diff
@@ -0,0 +1,180 @@
...
+        client = new DefaultHttpClient();
+        client.getParams().setParameter(CoreProtocolPNames.PROTOCOL_VERSION, HttpVersion.HTTP_1_1);
+    }
+
+    /**
+     * Test of GetHandler methods, of class GripServer.
+     */
+    @Test
+    public void testGetHandlers() {
+        System.out.println("testGetHandlers");
```
JLLeitschuh: Remove this.

core/src/test/java/edu/wpi/grip/core/http/GripServerTest.java
```diff
@@ -0,0 +1,180 @@
...
+            instance.removeGetHandler(path); // cleanup
+        }
+    }
+
+    /**
+     * Test of addPostHandler method, of class GripServer.
+     */
+    @Test
+    public void testAddPostHandler() {
+        System.out.println("addPostHandler");
```
JLLeitschuh: Remove this

core/src/test/java/edu/wpi/grip/core/http/GripServerTest.java
```diff
@@ -0,0 +1,180 @@
...
+            instance.removePostHandler(handler); // cleanup
+        }
+    }
+
+    /**
+     * Test of data supplier methods
+     */
+    @Test
+    public void testDataSuppliers() {
+        System.out.println("testDataSuppliers");
```
JLLeitschuh: Remove this

core/src/test/java/edu/wpi/grip/core/http/GripServerTest.java
```diff
@@ -0,0 +1,180 @@
...
+        try {
+            System.out.println("Starting server");
+            instance.start(); // should do nothing since the server's already running
+            System.out.println("Stopping server (1/2)");
+            instance.stop();  // stop the server so we know we can start it
+            System.out.println("Stopping server (2/2)");
+            instance.stop();  // second call should do nothing
+            System.out.println("Restarting server (1/2)");
+            instance.start(); // restart the server
+            System.out.println("Restarting server (2/2)");
```
JLLeitschuh: Instead of using println do some sort of assertions.

ui/src/main/java/edu/wpi/grip/ui/pipeline/source/HttpSourceController.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    @Inject
+    HttpSourceController(
+            final EventBus eventBus,
+            final OutputSocketController.Factory outputSocketControllerFactory,
+            final ExceptionWitnessResponderButton.Factory exceptionWitnessResponderButtonFactory,
+            @Assisted final HttpSource source) {
+        super(eventBus, outputSocketControllerFactory, exceptionWitnessResponderButtonFactory, source);
+        System.out.println("Creating new HttpSourceController");
```
JLLeitschuh: Remove this.

ui/src/main/java/edu/wpi/grip/ui/preview/PointSizeSocketPreviewView.java
```diff
@@ -16,15 +16,15 @@
 /**
  * A {@link SocketPreviewView} for OpenCV points and sizes
  */
-public class PointSizeSocketPreviewView<T extends IntPointer> extends SocketPreviewView<T> {
+public class PointSizeSocketPreviewView extends SocketPreviewView<IntPointer> {
```
JLLeitschuh: Why was this file changed?

ui/src/main/java/edu/wpi/grip/ui/preview/SocketPreviewViewFactory.java
```diff
@@ -38,7 +39,7 @@
         if (socket.getSocketHint().getType() == Mat.class) {
             previewView = (SocketPreviewView) new ImageSocketPreviewView(platform, (OutputSocket<Mat>) socket);
         } else if (socket.getSocketHint().getType() == Point.class || socket.getSocketHint().getType() == Size.class) {
-            previewView = (SocketPreviewView) new PointSizeSocketPreviewView(platform, socket);
+            previewView = (SocketPreviewView) new PointSizeSocketPreviewView(platform, (OutputSocket<IntPointer>) socket);
```
JLLeitschuh: Why was this file changed?

core/src/main/java/edu/wpi/grip/core/util/Holder.java
```diff
@@ -0,0 +1,42 @@
...
+
+    private T value;
+
+    /**
+     * Creates a new holder with the given default value.
+     */
+    public Holder(T value) {
+        if (value == null) {
+            throw new NullPointerException("Held value cannot be null");
+        }
```
JLLeitschuh: Use the preconditions stuff.

core/src/main/java/edu/wpi/grip/core/util/Holder.java
```diff
@@ -0,0 +1,42 @@
+
+package edu.wpi.grip.core.util;
+
+/**
+ * Simple thread-safe class to allow for setting values from lambdas.
+ */
+public final class Holder<T> {
```
JLLeitschuh: Why?
This class doesn't seem necessary.

SamCarlberg: It's better than using arrays.

JLLeitschuh: It seems like the only usage is in the source.
Why can't you just store the data in a variable.
What advantage does this abstraction give you?

SamCarlberg: Ah, it used to be before I refactored out the lambdas. I can just remove it, then.

core/src/main/java/edu/wpi/grip/core/sources/HttpSource.java
```diff
@@ -0,0 +1,103 @@
...
+    @Inject
+    HttpSource(
+            ExceptionWitness.Factory exceptionWitnessFactory,
+            EventBus eventBus,
+            GripServer server) {
+        super(exceptionWitnessFactory);
+        this.eventBus = eventBus;
+        this.server = server;
+        this.imageOutput = new OutputSocket<>(eventBus, outputHint);
+        System.out.println("Creating new HttpSource");
```
JLLeitschuh: Remove this println

core/src/test/java/edu/wpi/grip/core/http/GripServerTest.java
```diff
@@ -0,0 +1,173 @@
...
+        try {
+            instance.start(); // should do nothing since the server's already running
+            instance.stop();  // stop the server so we know we can start it
+            instance.stop();  // second call should do nothing
+            instance.start(); // restart the server
+            instance.start(); // second call should do nothing
+            instance.restart(); // should stop and then start again
+        } catch (Exception e) {
+            // Starting or stopping when in an invalid state will throw an exception
+            fail(e.getMessage());
```
JLLeitschuh: Don't do this. Let Junit catch the exception.
If you do this you loose the stack trace.

SamCarlberg:
I feel like this is ready to be merged, unless there is some glaring bug/defect/missing functionality that needs to be addressed.

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
```diff
@@ -0,0 +1,404 @@
...
+        HttpServer create(int port);
+    }
+
+    public static class HttpServerFactoryImpl implements HttpServerFactory {
+        @Override
+        public HttpServer create(int port) {
+            try {
+                return HttpServer.create(new InetSocketAddress(HOSTNAME, port), BACKLOG);
+            } catch (IOException ex) {
+                throw new GripException(ex);
```
JLLeitschuh: I would turn this into an IllegalStateException with the message "Failed to create HTTPServer"

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
```diff
@@ -0,0 +1,404 @@
...
+     * @return true if the handler was removed from at least one path
+     */
+    public boolean removePostHandler(PostHandler handler) {
+        return postHandlers.values()
+                .stream()
+                .map(list -> list.remove(handler))
+                .anyMatch(b -> b);
+    }
+
+    private final Set<String> contextPaths = new HashSet<>();
```
JLLeitschuh: Put this at the top of the file.

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
```diff
@@ -0,0 +1,404 @@
...
+        dataSuppliers.put(name, supplier);
+    }
+
+    /**
+     * Checks if the given name is associated with a supplier.
+     *
+     * @param name the name to check
+     * @return true if the name is associated with a supplier, false if not
+     */
+    public boolean hasDataSupplier(String name) {
```
JLLeitschuh: If this is only used in tests then make it package private and mark it `@VisibleForTesting`
but ideally you should only really test things you need to exist as a public method.

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
```diff
@@ -0,0 +1,404 @@
...
+        try {
+            return gson.toJson(data);
+        } catch (Throwable t) {
+            logger.log(Level.SEVERE, "Error generating json response", t);
+            throw new GripException(t);
+        }
+    }
+
+    private boolean didStart = false;
+    private boolean stopped = false;
```
JLLeitschuh: Put these at the top of the file.

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
```diff
@@ -0,0 +1,404 @@
...
+        } catch (IOException | IllegalStateException ex) {
+            Logger.getLogger(GripServer.class.getName()).log(Level.SEVERE, null, ex);
+            throw new GripException("Could not restart GripServer", ex);
+        }
+    }
+    
+    /**
+     * Gets the port this server is running on.
+     */
+    public int getPort() {
```
JLLeitschuh: Does this need to be public?

core/src/test/java/edu/wpi/grip/core/http/GripServerTest.java
```diff
@@ -0,0 +1,166 @@
...
+            for (int offset = 0;; offset++) {
+                try {
+                    server = HttpServer.create(new InetSocketAddress("localhost", port + offset), 1);
+                    break;
+                } catch (IOException e) {
+                    // That port is taken -- keep trying different ports
+                }
+            }
+            return server;
+        }
```
JLLeitschuh: This method could probably be written better. Something like this:
```java
@Override
public HttpServer create(int port) {
    final int MAX_ATTEMPTS = 200;
    IOException anException = null;
    for(int offset = 0; offset < MAX_ATTEMPTS; offset++) {
        try {
            // Will either return or throw an exception
            return HttpServer.create(new InetSocketAddress("localhost", port + offset), 1);
        } catch (IOException e) {
            anException = e;
            continue;
        }
    }
    throw new AssertionError("Could not create HTTP Server after trying "
            + MAX_ATTEMPTS + " different ports. Last exception as cause:", anException);
}
```

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
```diff
@@ -0,0 +1,404 @@
...
+    }
+    
+    
+    /**
+     * Restarts the server on the current port.
+     */
+    public void restart() {
+        try {
+            stop();
+            server = HttpServer.create(new InetSocketAddress("localhost", getPort()), BACKLOG);
```
JLLeitschuh: Shouldn't this use the factory??

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
```diff
@@ -0,0 +1,404 @@
...
+        }
+        return requestParameters;
+    }
+
+    private static String decodeUrlComponent(final String urlComponent) {
+        try {
+            return URLDecoder.decode(urlComponent, CHARSET.name());
+        } catch (final UnsupportedEncodingException ex) {
+            // Will never happen, UTF-8 is supported by all JVMs
+            throw new GripException(ex);
```
JLLeitschuh: I'd turn this into an `IllegalStateException` with some sort of message about expecting UTF8.

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
```diff
@@ -0,0 +1,404 @@
...
+    /**
+     * Restarts the server on the current port.
+     */
+    public void restart() {
+        try {
+            stop();
+            server = HttpServer.create(new InetSocketAddress("localhost", getPort()), BACKLOG);
+            start();
+        } catch (IOException | IllegalStateException ex) {
+            Logger.getLogger(GripServer.class.getName()).log(Level.SEVERE, null, ex);
```
JLLeitschuh: Why are you getting the logger again? Isn't the logger declared at the top of the file?

core/src/main/java/edu/wpi/grip/core/http/PostHandler.java
```diff
@@ -0,0 +1,18 @@
+
+package edu.wpi.grip.core.http;
+
+/**
+ * Handler for converting incoming data over an HTTP {@code POST} operation.
+ */
+public interface PostHandler {
```
JLLeitschuh: You could make this an `@FunctionalInterface` and use it like a lambda instead of having classes implement it.

core/src/main/java/edu/wpi/grip/core/http/PostHandler.java
```diff
@@ -0,0 +1,18 @@
...
+
+    /**
+     * Converts the given bytes to some object. This operation should have side
+     * effects (i.e. it is <i>stateful</i>}.
+     *
+     * @param bytes the bytes to convert
+     * @return true if the conversion was a success; false if not.
+     */
+    public boolean convert(byte[] bytes);
```
JLLeitschuh: Public keyword is unnecessary. This is an interface.

core/src/main/java/edu/wpi/grip/core/http/GetHandler.java
```diff
@@ -0,0 +1,13 @@
+
+package edu.wpi.grip.core.http;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Handler for creating responses to incoming HTTP {@code GET} requests.
+ */
+public interface GetHandler {
```
JLLeitschuh: Functional interface?

core/src/main/java/edu/wpi/grip/core/http/GetHandler.java
```diff
@@ -0,0 +1,13 @@
...
+
+    public String createResponse(Map<String, List<String>> requestParameters);
```
JLLeitschuh: Doesn't need the public keyword.

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
```diff
@@ -0,0 +1,404 @@
...
+        return b.toByteArray();
+    }
+
+    @Subscribe
+    public void settingsChanged(ProjectSettingsChangedEvent event) {
+        int port = event.getProjectSettings().getServerPort();
+        if (port != getPort()) {
+            stop();
+            server = serverFactory.create(port);
+            start();
```
JLLeitschuh: What if the server was stopped when the project settings changed? Now you've just restarted the server because the settings have changed.

core/src/main/java/edu/wpi/grip/core/sources/HttpSource.java
```diff
@@ -0,0 +1,102 @@
...
+    protected boolean updateOutputSockets() {
+        if (!gotImage) {
+            return false;
+        }
+        if (data.length == 0) {
+            // Got data, but it's empty
+            return false;
+        }
+        gotImage = false;
+        imageOutput.setValue(opencv_imgcodecs.imdecode(new Mat(data), opencv_imgcodecs.CV_LOAD_IMAGE_COLOR));
```
JLLeitschuh: This could cause a potential memory leak.
You are allocating a new mat for every image uploaded.
What is the fastest rate you could see images moving through this source?

core/src/test/java/edu/wpi/grip/core/http/GripServerTest.java
```diff
@@ -0,0 +1,166 @@
...
+    public void testGetHandlers() {
+        String path = "/testGetHandlers";
+        GetHandler handler = params -> path;
+        instance.addGetHandler(path, handler);
+        instance.addDataSupplier(path, () -> path);
+        try {
+            String data = doGet(path);
+            assertEquals(data, path);
+        } catch (IOException ex) {
+            fail(ex.getMessage());
```
JLLeitschuh: Don't do this. Just let the exception propigate to JUnit.
Add `throws IOException` to the method signature and JUnit will give us more information if there is a problem.

core/src/test/java/edu/wpi/grip/core/sources/HttpSourceTest.java
```diff
@@ -0,0 +1,112 @@
...
+    @Test
+    public void testPostImage() throws IOException {
+        System.out.println("testPost()");
+        server.addPostHandler(GripServer.IMAGE_UPLOAD_PATH, source);
+        OutputSocket<Mat> imageSource = source.getOutputSockets()[0];
+        assertTrue("The value should not be present if the source hasn't been initialized and no image POSTed", imageSource.getValue().get().empty());
+        source.initialize();
+        assertTrue("The value should not be present since the source has been initialized but no image POSTed", imageSource.getValue().get().empty());
+        doPost(GripServer.IMAGE_UPLOAD_PATH, logoFile);
+        assertTrue("The value should now be present after POSTing the image", imageSource.getValue().get().empty());
```
JLLeitschuh: Wait, this doesn't make sense. The image source value should not be empty now.....

core/src/test/java/edu/wpi/grip/core/sources/HttpSourceTest.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    @Test
+    public void testNulls() {
+        // TODO
+        System.out.println("testNulls()");
```
JLLeitschuh: Remove this.

core/src/test/java/edu/wpi/grip/core/http/GripServerTest.java
```diff
@@ -0,0 +1,166 @@
...
+        } finally {
+            instance.removeGetHandler(path); // cleanup
```
JLLeitschuh: You shouldn't need to do this? The test constructor gets called again each time for each test.

core/src/test/java/edu/wpi/grip/util/GRIPCoreTestModule.java
```diff
@@ -69,6 +74,12 @@ public void tearDown() {
     protected void configure() {
         assert setUp : "The GRIPCoreTestModule handler was not set up. Call 'setUp' before passing the injector";
         super.configure();
+        // HTTP server injection bindings
+        bind(GripServer.HttpServerFactory.class).to(GripServerTest.TestServerFactory.class);
+        bind(GripServer.class).asEagerSingleton();
+        bind(MapNetworkPublisherFactory.class)
+                .annotatedWith(Names.named("httpManager"))
+                .to(HttpManager.class);
```
JLLeitschuh: Do you really need the real http server in the tests?

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
```diff
@@ -0,0 +1,404 @@
...
+                .serializeSpecialFloatingPointValues()
+                .create();
+        final Map<String, Object> data = new HashMap<>();
+        dataSuppliers.entrySet()
+                .stream()
+                .filter(e -> params.isEmpty() || params.containsKey(e.getKey())) // send everything if nothing is specified
+                .forEach(e -> data.put(e.getKey(), e.getValue().get()));
+        try {
+            return gson.toJson(data);
+        } catch (Throwable t) {
```
JLLeitschuh: Throwable includes everything from IOExceptions, to OutOfMemoryErrors to InterruptedExceptions.
You should probably be more specific about what exception you want to catch here.

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
```diff
@@ -0,0 +1,404 @@
...
+    private String createJson(Map<String, List<String>> params) {
+        final Gson gson = new GsonBuilder()
+                .setPrettyPrinting()
+                .serializeSpecialFloatingPointValues()
+                .create();
+        final Map<String, Object> data = new HashMap<>();
+        dataSuppliers.entrySet()
+                .stream()
+                .filter(e -> params.isEmpty() || params.containsKey(e.getKey())) // send everything if nothing is specified
+                .forEach(e -> data.put(e.getKey(), e.getValue().get()));
```
JLLeitschuh: This can become:
```java
        final Map<String, Object> data = 
                dataSuppliers
                .entrySet()
                .stream()
                .filter(e -> params.isEmpty() || params.containsKey(e.getKey())) // send everything if nothing is specified
                .collect(Collectors.toMap(v -> v.getKey(), v -> v.getValue().get()));
```

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
```diff
@@ -0,0 +1,404 @@
...
+    public int getPort() {
+        return server.getAddress().getPort();
+    }
+
+    /**
+     * Parses request parameters from a URI. For example, the URI
+     * "http://roborio-190-frc.local:8080/GRIP/data?foo=1;bar=2" will be parsed to
+     * {"foo"="1", "bar"="2"}.
+     */
+    private static Map<String, List<String>> getRequestParameters(final URI requestUri) {
```
JLLeitschuh: This method could do with having higher test coverage.
Right now it doesn't run anything inside the `if (requestQuery != null) {` statement

JLLeitschuh:
Can you also remove the println's from the code. Both in the tests and other places too.

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
SamCarlberg: Good idea. I keep forgetting that there are built in collectors.

core/src/test/java/edu/wpi/grip/core/sources/HttpSourceTest.java
SamCarlberg: `testNulls()` is autogenerated by some automatic test suite if I don't declare it here. The autogenerated one fails because it can't construct a `GripServer` instance, so this is just here to shut it up.

core/src/test/java/edu/wpi/grip/core/sources/HttpSourceTest.java
SamCarlberg: Oh whoops

core/src/main/java/edu/wpi/grip/core/sources/HttpSource.java
SamCarlberg: Apparently `Mats` keep track of how many references they have, so shouldn't it be unnecessary to `release()` it after it goes out of scope?

Are you asking about maximum throughput for this operation or just how often a new image would be uploaded to the server?

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
SamCarlberg: I don't follow. The server should only be stopped at shutdown or when the port gets changed. In the first case, the settings won't change; in the second case, the server's getting restarted anyway. I should probably make `stop()` a class-private method to enforce this.

core/src/main/java/edu/wpi/grip/core/http/GetHandler.java
SamCarlberg: ~~Yes, functional interface.~~

 `@FunctionalInterface` is just a marker annotation, it doesn't affect it being used as a lambda

core/src/main/java/edu/wpi/grip/core/http/PostHandler.java
SamCarlberg: `HttpSource` didn't originally implement this interface, but I needed a good way of removing the handler when the source gets removed from the pipeline. I could have a variable for this, but the current way is cleaner in my opinion

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
SamCarlberg: Whoops, that was autogenerated by Netbeans and I didn't remember to change it

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
SamCarlberg: [That's not the correct exception to use](https://docs.oracle.com/javase/7/docs/api/java/lang/IllegalStateException.html).

Normally I'd stick a `@SneakyThrows(UnsupportedEncodingException.class)` on this method, but adding lombok as a dependency just for this would be overkill.

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
SamCarlberg: Yes. Yes it should.

core/src/test/java/edu/wpi/grip/core/http/GripServerTest.java
SamCarlberg: Yeah, that's a good idea

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
SamCarlberg: It's nice to be able to know what port the server's running on. It doesn't change state so it may as well be public

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
SamCarlberg: `IllegalStateException` "Signals that a method has been invoked at an illegal or inappropriate time. In other words, the Java environment or Java application is not in an appropriate state for the requested operation."

There doesn't seem to be an applicable `RuntimeException` subclass in the java libraries

core/src/test/java/edu/wpi/grip/core/http/GripServerTest.java
SamCarlberg: Gotcha. I'm not very familiar with JUnit, so that's good to know

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
SamCarlberg: What? You said earlier that we'd really only need to test `public` methods. And everything runs inside that `if` statement....

core/src/test/java/edu/wpi/grip/core/sources/HttpSourceTest.java
SamCarlberg: Fixed in 8e9c160

core/src/test/java/edu/wpi/grip/core/http/GripServerTest.java
SamCarlberg: Fixed in 8e9c160

core/src/test/java/edu/wpi/grip/util/GRIPCoreTestModule.java
SamCarlberg: I need to be able to respond to HTTP messages. So yes.

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
JLLeitschuh: Yes, you only need to test public methods. But your test coverage should hit more branches of code.
I'm looking at the code coverage results and the code inside that if statement never runs according to the test coverage tool I'm using.

core/src/test/java/edu/wpi/grip/core/sources/HttpSourceTest.java
```diff
@@ -0,0 +1,115 @@
...
+        assertTrue("The value should not be present since the source has been initialized but no image POSTed", imageSource.getValue().get().empty());
+
+        doPost(GripServer.IMAGE_UPLOAD_PATH, logoFile);
+        Thread.sleep(100); // wait for the image to be uploaded
+        source.updateOutputSockets();
+        assertFalse("The value should now be present after POSTing the image", imageSource.getValue().get().empty());
+    }
+
+    @Test
+    public void testNulls() {
```
JLLeitschuh: Can you actually fix the automated test instead of just doing this.

core/src/test/java/edu/wpi/grip/util/GRIPCoreTestModule.java
JLLeitschuh: Right, but do you need to use dependency injection in your tests?
If you want to test HTTP messages in a test create a concrete instance.
You are writing unit tests not integration tests.

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
SamCarlberg: Oh, you're saying the tests don't run over that part. I'll add something for that.

core/src/test/java/edu/wpi/grip/util/GRIPCoreTestModule.java
SamCarlberg: Every test class that uses injection will fail because the injector can't be created due to no implementation bound for HttpServerFactory when it's trying to create an HttpSource. So this has to be in here because the source factories are bound in `GRIPCoreModule.java`

[Stacktrace](https://gist.github.com/SamCarlberg/bd9ae83998b1d283df69)

core/src/test/java/edu/wpi/grip/core/sources/HttpSourceTest.java
SamCarlberg: Fixed in 87b635f

core/src/main/java/edu/wpi/grip/core/exception/GripException.java
```diff
@@ -0,0 +1,26 @@
+
+package edu.wpi.grip.core.exception;
+
+/**
+ * An exception thrown when something goes wrong with an internal GRIP
+ * operation.
+ */
+public class GripException extends RuntimeException {
+
+    public GripException() {
```
JLLeitschuh: I'm not a huge fan of having this exception in the first place. But I'm having a hard time justifying telling you to remove it as well.

Regardless. Please remove this constructor. A message should be required for this exception.

core/src/main/java/edu/wpi/grip/core/exception/GripException.java
```diff
@@ -0,0 +1,26 @@
...
+        super();
+    }
+
+    public GripException(String message) {
+        super(message);
+    }
+
+    public GripException(Throwable cause) {
```
JLLeitschuh: Remove this constructor.
If we're going to add a `GripException` we should require that a message is provided.

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
```diff
@@ -0,0 +1,410 @@
...
+                .entrySet()
+                .stream()
+                .filter(e -> params.isEmpty() || params.containsKey(e.getKey())) // send everything if nothing is specified
+                .collect(Collectors.toMap(Map.Entry::getKey, v -> v.getValue().get()));
+        try {
+            return gson.toJson(data);
+        } catch (RuntimeException ex) {
+            // If we don't catch this, it will be silently consumed by something up the call stack
+            logger.log(Level.SEVERE, "Error generating json response", ex);
+            return "{}"; // empty json object
```
JLLeitschuh: I'd provide a json object with an error message. Just so that it shows up in the http request too.

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
```diff
@@ -0,0 +1,410 @@
...
+            stopped = false;
+        }
+    }
+
+    /**
+     * Stops this server. Note that a shutdown hook has been registered to call
+     * this method, so it's unlikely that this should need to be called.
+     */
+    @VisibleForTesting
+    public void stop() {
```
JLLeitschuh: Methods annotated with `@VisibleForTesting` really should be package private.

core/src/main/java/edu/wpi/grip/core/operations/network/http/HttpPublishOperation.java
```diff
@@ -0,0 +1,43 @@
...
+import edu.wpi.grip.core.operations.network.Publishable;
+
+import java.io.InputStream;
+import java.util.Optional;
+import java.util.function.Function;
+
+/**
+ *
+ */
+public abstract class HttpPublishOperation<S, T extends Publishable, P> extends PublishAnnotatedOperation<S, T, P> {
```
JLLeitschuh: Add documentation. This is just an empty javadoc at the moment

core/src/main/java/edu/wpi/grip/core/sources/HttpSource.java
```diff
@@ -0,0 +1,111 @@
...
+            // Got data, but it's empty
+            return false;
+        }
+        gotImage = false;
+
+        Mat tmp = new Mat(data);
+        imageOutput.setValue(opencv_imgcodecs.imdecode(tmp, opencv_imgcodecs.CV_LOAD_IMAGE_COLOR));
+        tmp.release();
+        return true;
+    }
```
JLLeitschuh: A bit better way of writing this that would manage memory a bit better would be this way:

```java
@Override
protected boolean updateOutputSockets() {
    if (!gotImage) {
        return false;
    }
    if (data.length == 0) {
        // Got data, but it's empty
        return false;
    }
    gotImage = false;

    try (final Mat tmp = new Mat(data)) {
        opencv_imgcodecs.imdecode(
                tmp, opencv_imgcodecs.CV_LOAD_IMAGE_COLOR, imageOutput.getValue().get());
    } catch (Exception e) { // NO-PMD
        // This should never happen
        getExceptionWitness().flagException(e, "Failed to deallocate tmp mat");
    }
    return true;
}
```

core/src/test/java/edu/wpi/grip/core/sources/HttpSourceTest.java
```diff
@@ -0,0 +1,110 @@
...
+            @Subscribe
+            public void onUnexpectedThrowableEvent(UnexpectedThrowableEvent event) {
+                event.handleSafely((throwable, message, isFatal) -> {
+                    throwable.printStackTrace();
+                });
+            }
+        }
+        eventBus.register(new UnhandledExceptionWitness());
+
+        server = injector.getInstance(GripServer.class);
```
JLLeitschuh: Do you really need to use dependency injection to create the GRIPServer? Can't you just call the constructor?

core/src/test/java/edu/wpi/grip/core/sources/HttpSourceTest.java
```diff
@@ -0,0 +1,110 @@
...
+    private GripServer server;
+    private GRIPCoreTestModule coreTestModule;
+    private HttpSource source;
+    private CloseableHttpClient postClient;
+
+    @Before
+    public void setup() throws Exception {
+        coreTestModule = new GRIPCoreTestModule();
+        coreTestModule.setUp();
+        final Injector injector = Guice.createInjector(coreTestModule);
```
JLLeitschuh: If you really need to use dependency injection, you can do this to get the network dependencies.

```java
Guice.createInjector(coreTestModule, new GRIPNetworkModule());
```

core/src/test/java/edu/wpi/grip/util/GRIPCoreTestModule.java
JLLeitschuh: https://github.com/WPIRoboticsProjects/GRIP/pull/551/files#r57812557

Also, do you need the entire server?
It seems like the HTTPSource only needs a class that provides the `addPostHandler` and `removePostHandler` methods. You could bind that interface to the instance for the GRIPNetworkModule and in the test module just use a mock.

JLLeitschuh:
How long does the server take to start? Is it something that could cause GRIP's launch to fail if it can't bind to the port because it's not available?

core/src/main/java/edu/wpi/grip/core/exception/GripException.java
SamCarlberg: It might be a good idea to use this as as base class for internal exceptions and use it like `RuntimeException` - make subclasses for specific uses but never use it directly. So maybe have something like `GripServerException extends GripException`

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
SamCarlberg: Whoops, I forgot remove that annotation.

core/src/main/java/edu/wpi/grip/core/sources/HttpSource.java
SamCarlberg: It doesn't need to be in a try-with-resources statement. `tmp` will be GC'd anyway after it goes out of scope (which will deallocate it anyway). Manually calling `deallocate` would be fine to quickly deallocate memory. (Also, I should have used `deallocate()` instead of `release()` - `release()` only works if the `Mat` owns the data, which `tmp` does not.)

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
SamCarlberg: Is that necessary? Getting an empty response already implies that something went wrong. I could have it respond with an HTTP 500 error in the HTTP response, though.

SamCarlberg: Is that necessary? Getting an empty response already implies that something went wrong. I could have it respond with an HTTP 500 error in the HTTP response, though.

core/src/test/java/edu/wpi/grip/util/GRIPCoreTestModule.java
SamCarlberg: It also needs to test that the source is publishing the bytes received over the HTTP connection, and that wants the server to be running to make sure that it's getting what the server would be giving it.

JLLeitschuh: Right, but every other test doesn't need to have the HTTPServer bound.

What I'm trying to get at is that there are other tests that don't need the HTTPServer so when dependency injection is used in other tests I don't want another instance of the HTTPServer floating around. Thats why I moved NetworkTables and ROS into a seperate DI module.
I'm trying to avoid having network tables from being instantiated when I don't want it to be.
If there was a problem with network tables it was causing a ripple effect to other tests.

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
JLLeitschuh: Idk. Up to you I suppose.

Some JSON message responses have standard reply formats that include errors.
http://stackoverflow.com/a/14538774/3708426

core/src/main/java/edu/wpi/grip/core/GRIPCoreModule.java
```diff
@@ -9,9 +9,12 @@
 import com.google.inject.spi.InjectionListener;
 import com.google.inject.spi.TypeEncounter;
 import com.google.inject.spi.TypeListener;
+
 import edu.wpi.grip.core.events.UnexpectedThrowableEvent;
+import edu.wpi.grip.core.http.GripServer;
```
JLLeitschuh: :+1:

core/src/test/java/edu/wpi/grip/core/http/GripServerTest.java
```diff
@@ -0,0 +1,215 @@
...
+        assertEquals("Server should return an internal error (500)", response.getStatusLine().getStatusCode(), 500);
+        assertTrue("Handler should have run", didHandle[0]);
+    }
+
+    @Test
+    public void testAddRemoveDataSupplier() {
+        String name = "testDataSuppliers";
+        Supplier<?> supplier = () -> name;
+        instance.addDataSupplier(name, supplier);
+        assertTrue(instance.hasDataSupplier(name));
```
JLLeitschuh: :+1:

core/src/test/java/edu/wpi/grip/core/sources/HttpSourceTest.java
```diff
@@ -0,0 +1,98 @@
...
+        // We have to manually update the output sockets to get the image
+        source.updateOutputSockets();
+        assertTrue("The value should not be present if the source hasn't been initialized and no image POSTed", imageSource.getValue().get().empty());
+
+        source.initialize(); // adds the source as a PostHandler to the server
+        source.updateOutputSockets();
+        assertTrue("The value should not be present since the source has been initialized but no image POSTed", imageSource.getValue().get().empty());
+
+        doPost(GripServer.IMAGE_UPLOAD_PATH, logoFile);
+        Thread.sleep(100); // wait for the image to be uploaded
```
JLLeitschuh: Can you just add a responder that waits for the callback to happen from the HTTP server when the connection happens. Adding waits can slow down tests.
We have a deferred testing library that allows for waiting for callbacks to execute.
This is the library.

core/src/test/java/edu/wpi/grip/core/http/GripServerTest.java
```diff
@@ -0,0 +1,216 @@
...
+        final String dataName = "testWithSupplier";
+        final Map data = ImmutableMap.of("foo", "bar");
+        instance.addDataSupplier(dataName, () -> data);
+        assertEquals("Generated json was malformed", gson.toJson(ImmutableMap.of(dataName, data)), doGet(GripServer.DATA_PATH + "?" + dataName));
+        assertEquals("Generated json was malformed", gson.toJson(ImmutableMap.of(dataName, data)), doGet(GripServer.DATA_PATH));
+        assertEquals("There shouldn't be any data on this path", "{}", doGet(GripServer.DATA_PATH + "?no_data_here"));
+    }
+
+    @Test
+    public void testStartStop() throws GripException, IllegalStateException {
```
SamCarlberg: This will thrown an exception if something fails, no need to put in an `assert` or `fail` condition

core/src/test/java/edu/wpi/grip/core/MockPipeline.java
```diff
@@ -0,0 +1,25 @@
...
+import edu.wpi.grip.core.settings.ProjectSettings;
+
+/**
+ *
+ */
+public class MockPipeline extends Pipeline {
+
+    private ProjectSettings settings;
+
+    public MockPipeline(ProjectSettings settings) {
```
JLLeitschuh: You don't actually need this class. There is a "SettingsProvider" class that you can implement that will work.

SamCarlberg: Yeah, I just didn't change it after merging `master`

ui/src/main/java/edu/wpi/grip/ui/pipeline/AddSourceView.java
```diff
@@ -48,6 +49,7 @@
     private final MultiImageFileSource.Factory multiImageSourceFactory;
     private final ImageFileSource.Factory imageSourceFactory;
     private final CameraSource.Factory cameraSourceFactory;
+    private final HttpSource.Factory httpSourceFactory;
```
SamCarlberg: This is wrong; `httpSourceFactory` is used. Looks like codacy doesn't play nice with lambdas.

core/src/main/java/edu/wpi/grip/core/exception/ImpossibleExceptionError.java
```diff
@@ -0,0 +1,12 @@
+package edu.wpi.grip.core.exception;
+
+/**
+ * An error that occurs when a checked exception that should never be thrown is thrown. This is <i>not</i> an exception
+ * because, if it's thrown, something is wrong with the runtime environment or system that we cannot hope to fix.
+ */
+public class ImpossibleExceptionError extends Error {
```
SamCarlberg: This is intentional.

core/src/main/java/edu/wpi/grip/core/http/HttpPipelineSwitcher.java
```diff
@@ -0,0 +1,58 @@
...
+    protected void handleIfPassed(String target, Request baseRequest, HttpServletRequest request, HttpServletResponse response) throws IOException, ServletException {
+        if (!isPost(request)) {
+            response.setStatus(HttpServletResponse.SC_METHOD_NOT_ALLOWED);
+            baseRequest.setHandled(true);
+            return;
+        }
+        String projectXml = new String(IOUtils.toByteArray(request.getInputStream()), "UTF-8");
+        // Need to be careful - this will cause a deadlock if not called from the JavaFX application thread
+        // TODO change code in UI module to make this safe -- there shouldn't be any references to UI code in the core module
+        switch (mode) {
```
SamCarlberg: There's only one possible constructor, and that's the default one. Non-issue.

core/src/main/java/edu/wpi/grip/core/http/GenericHandler.java
```diff
@@ -0,0 +1,147 @@
...
+
+    /**
+     * HTTP content type for HTML.
+     */
+    public static final String CONTENT_TYPE_HTML = "text/html";
+
+    /**
+     * Creates a generic handler for all contexts on the server.
+     */
+    protected GenericHandler() {
```
SamCarlberg: There's only one possible constructor, and that's the default one. Non-issue.

core/src/main/java/edu/wpi/grip/core/http/GenericHandler.java
```diff
@@ -0,0 +1,147 @@
...
+     * Creates a generic handler that handles requests for the given context.
+     * <p>
+     * Note that the context <strong>is case sensitive</strong>.
+     * </p>
+     *
+     * @param context the context for this handler
+     * @param doClaim flag marking if the given context should be claimed
+     * @throws IllegalArgumentException if the given context has already been claimed
+     */
+    protected GenericHandler(String context, boolean doClaim) {
```
SamCarlberg: There's only one possible constructor, and that's the default one. Non-issue.

core/src/main/java/edu/wpi/grip/core/http/HttpPipelineSwitcher.java
SamCarlberg: There are only two possible modes and it will not change. Non-issue.

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
```diff
@@ -0,0 +1,215 @@
...
+
+    @Inject
+    GripServer(JettyServerFactory serverFactory, SettingsProvider settingsProvider) {
+        this.port = settingsProvider.getProjectSettings().getServerPort();
+        this.serverFactory = serverFactory;
+        this.server = serverFactory.create(port);
+        this.server.setHandler(handlers);
+        handlers.addHandler(new NoContextHandler());
+
+        Runtime.getRuntime().addShutdownHook(new Thread(this::stop));
```
JLLeitschuh: Why is this necessary? What happens if this isn't here?

core/src/main/java/edu/wpi/grip/core/operations/Operations.java
```diff
@@ -34,14 +59,15 @@
...
-               @Named("ntManager") MapNetworkPublisherFactory ntPublisherFactory,
-               @Named("rosManager") ROSNetworkPublisherFactory rosPublishFactory,
-               InputSocket.Factory isf,
-               OutputSocket.Factory osf) {
+    @Inject Operations(EventBus eventBus,
+                       @Named("ntManager") MapNetworkPublisherFactory ntPublisherFactory,
+                       @Named("httpManager") MapNetworkPublisherFactory httpPublishFactory,
+                       @Named("rosManager") ROSNetworkPublisherFactory rosPublishFactory,
+                       InputSocket.Factory isf,
+                       OutputSocket.Factory osf) {
```
JLLeitschuh: Unnecessary newline changes.

core/src/main/java/edu/wpi/grip/core/serialization/Project.java
```diff
@@ -78,10 +85,24 @@ public void open(File file) throws IOException {
...
+    public void open(String projectXml) {
+        open(new StringReader(projectXml));
+    }
+
     @VisibleForTesting
     void open(Reader reader) {
+        eventBus.post(new ProjectUnloadedEvent());
         this.pipeline.clear();
         this.xstream.fromXML(reader);
+        eventBus.post(new ProjectLoadedEvent());
```
JLLeitschuh: Just make `ProjectLoadedEvent` implement `RunPipelineEvent`?

core/src/test/java/edu/wpi/grip/core/operations/network/http/HttpPublisherTest.java
```diff
@@ -0,0 +1,183 @@
...
+
+    private void unclaimDataHandler() {
+        try {
+            Field f = GenericHandler.class.getDeclaredField("claimedContexts");
+            f.setAccessible(true);
+            @SuppressWarnings("unchecked")
+            Set<String> claimedContexts = (Set<String>) f.get(null);
+            claimedContexts.remove(GripServer.DATA_PATH);
+        } catch (IllegalAccessException | NoSuchFieldException e) {
+            fail("Could not unclaim /GRIP/data");
```
JLLeitschuh: You're suppressing the stack trace here.
I'd throw an `AssertionError` instead and pass the exception into that.

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
SamCarlberg: Basically tells Jetty to stop handling new requests and finish up any currently running handlers. Otherwise it'll keep responding to stuff until the server thread gets killed by the JVM.

core/src/test/java/edu/wpi/grip/core/sources/HttpSourceTest.java
```diff
@@ -0,0 +1,96 @@
...
+    private GripServer server;
+    private HttpSource source;
+    private CloseableHttpClient postClient;
+
+    @Before
+    public void setUp() throws URIException, URISyntaxException {
+        GRIPCoreTestModule m = new GRIPCoreTestModule();
+        m.setUp();
+        Injector injector = Guice.createInjector(m);
+        OutputSocket.Factory osf = injector.getInstance(OutputSocket.Factory.class);
```
JLLeitschuh: There's a mock output socket factory you can use.

core/src/main/java/edu/wpi/grip/core/operations/Operations.java
SamCarlberg: Yeah, that was a weird autoformat thing Intellij did

ui/src/main/java/edu/wpi/grip/ui/pipeline/AddSourceView.java
```diff
@@ -97,17 +111,17 @@ private SourceDialog(final Parent root, Control inputField) {
...
-                    "*.tiff", "*.tif"           // TIFF files
+                    new ExtensionFilter("Image Files",
+                            "*.bmp", "*.dib",           // Windows bitmaps
+                            "*.jpeg", "*.jpg", "*.jpe", // JPEG files
+                            "*.jp2",                    // JPEG 2000 files
+                            "*.png",                    // Portable Network Graphics
+                            "*.webp",                   // WebP
+                            "*.pbm", "*.pgm", "*.ppm",  // Portable image format
+                            "*.sr", "*.ras",            // Sun rasters
+                            "*.tiff", "*.tif"           // TIFF files
```
JLLeitschuh: Random line changes?

core/src/main/java/edu/wpi/grip/core/serialization/Project.java
SamCarlberg: That makes a lot of sense

ui/src/main/java/edu/wpi/grip/ui/pipeline/AddSourceView.java
SamCarlberg: Random Intellij autoformatting

core/src/main/java/edu/wpi/grip/core/serialization/Project.java
JLLeitschuh: This seems like a level of indirection. Could you just call a method on an object instead??
What is the goal of these events? Are these events only used in one place?

core/src/test/java/edu/wpi/grip/core/operations/network/http/HttpPublisherTest.java
SamCarlberg: Good catch, thanks

core/src/main/java/edu/wpi/grip/core/serialization/Project.java
SamCarlberg: Project loaded/unloaded events are used to let the pipeline know that it shouldn't run while a new project is being loaded.

core/src/main/java/edu/wpi/grip/core/http/HttpPipelineSwitcher.java
```diff
@@ -0,0 +1,60 @@
...
+        }
+        switch (mode) {
+            case HEADLESS:
+                project.open(new String(IOUtils.toByteArray(request.getInputStream()), "UTF-8"));
+                response.setStatus(HttpServletResponse.SC_CREATED);
+                baseRequest.setHandled(true);
+                break;
+            case GUI:
+                // Don't run in GUI mode, it doesn't make much sense and can easily deadlock if pipelines are rapidly posted.
+                // Intentional fall-through to default
```
JLLeitschuh: I feel like this will leave people scratching their heads why it doesn't work.
Something better we can do here?

core/src/main/java/edu/wpi/grip/core/operations/network/http/DataHandler.java
```diff
@@ -0,0 +1,125 @@
...
+
+    /**
+     * Adds a supplier for data with the given name. The data will be published to {@code /GRIP/data} on the internal
+     * HTTP server.
+     *
+     * @param name     the name of the data
+     * @param supplier a supplier for the data
+     */
+    public void addDataSupplier(String name, Supplier<?> supplier) {
+        dataSuppliers.put(checkNotNull(name), checkNotNull(supplier));
```
JLLeitschuh: Can you put some sort of descriptive text in there so we know which one fails?

core/src/main/java/edu/wpi/grip/core/serialization/Project.java
JLLeitschuh: Why not just stop/start the pipeline here? Use `pipelineRunner.stopAndAwait()`
That would solve all the synchronization issues.

core/src/main/java/edu/wpi/grip/core/http/HttpPipelineSwitcher.java
SamCarlberg: How so? It doesn't make much sense to be setting the pipeline from a remote source during GUI mode, the whole point of it is to load and edit pipelines that *you* choose

core/src/main/java/edu/wpi/grip/core/operations/network/http/DataHandler.java
SamCarlberg: Be better to move the `checkNotNull` calls to their own lines before the `put`

JLLeitschuh: Either way. I think adding the string is easier because then its in the stack trace message.

core/src/main/java/edu/wpi/grip/core/serialization/Project.java
SamCarlberg: I guess that'd be alright. I didn't really like having the project know about or directly control the pipeline state. Just implemented your suggestion and works just the same so ¯\\\_(ツ)\_/¯

core/src/main/java/edu/wpi/grip/core/http/HttpPipelineSwitcher.java
```diff
@@ -0,0 +1,60 @@
...
+    @Override
+    protected void handleIfPassed(String target, Request baseRequest, HttpServletRequest request, HttpServletResponse response) throws IOException, ServletException {
+        if (!isPost(request)) {
+            response.setStatus(HttpServletResponse.SC_METHOD_NOT_ALLOWED);
+            baseRequest.setHandled(true);
+            return;
+        }
+        switch (mode) {
+            case HEADLESS:
+                project.open(new String(IOUtils.toByteArray(request.getInputStream()), "UTF-8"));
```
AustinShalit: Is it worth replacing "UTF-8" with `StandardCharsets.UTF_8.name()`?

[StandardCharsets (Java Platform SE 8)](https://docs.oracle.com/javase/8/docs/api/java/nio/charset/StandardCharsets.html)

SamCarlberg: I feel like it's an unnecessary level of indirection for *the* HTTP standard encoding. If it wasn't using UTF-8 I'd use StandardCharsets instead.

core/src/main/java/edu/wpi/grip/core/operations/network/http/DataHandler.java
```diff
@@ -0,0 +1,127 @@
...
+        super(GripServer.DATA_PATH, true);
+        checkNotNull(eventBus);
+        this.dataSuppliers = new HashMap<>();
+        this.gson = new GsonBuilder()
+                .setPrettyPrinting()
+                .serializeSpecialFloatingPointValues()
+                .create();
+        this.staleData = new AtomicBoolean(false);
+
+        eventBus.register(this);
```
JLLeitschuh: Don't do this. This is already done for classes that are injected.

core/src/test/java/edu/wpi/grip/core/http/GenericHandlerTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+    private GenericHandler gh;
+
+    @Test(expected = NullPointerException.class)
+    public void testNullContext() {
+        gh = new MockGenericHandler(null);
+    }
+
+    @Test
+    public void testNoClaim() {
```
JLLeitschuh: Or suppress the warning.

ui/src/main/java/edu/wpi/grip/ui/pipeline/source/HttpSourceController.java
```diff
@@ -0,0 +1,31 @@
...
+import com.google.inject.Inject;
+import com.google.inject.assistedinject.Assisted;
+
+import edu.wpi.grip.core.sources.HttpSource;
+import edu.wpi.grip.ui.components.ExceptionWitnessResponderButton;
+import edu.wpi.grip.ui.pipeline.OutputSocketController;
+
+/**
+ *
+ */
```
JLLeitschuh: Documentation

ui/src/main/java/edu/wpi/grip/ui/pipeline/AddSourceView.java
```diff
@@ -48,6 +59,7 @@
     private final MultiImageFileSource.Factory multiImageSourceFactory;
     private final ImageFileSource.Factory imageSourceFactory;
     private final CameraSource.Factory cameraSourceFactory;
+    private final HttpSource.Factory httpSourceFactory;
```
SamCarlberg: This is going to stay a private field to stay consistent with the rest of the source factories.

core/src/main/java/edu/wpi/grip/core/http/GripServer.java
```diff
@@ -0,0 +1,238 @@
...
+  GripServer(ContextStore contextStore,
+             JettyServerFactory serverFactory,
+             SettingsProvider settingsProvider) {
+    this.port = settingsProvider.getProjectSettings().getServerPort();
+    this.serverFactory = serverFactory;
+    this.server = serverFactory.create(port);
+    this.server.setHandler(handlers);
+    handlers.addHandler(new NoContextHandler(contextStore));
+
+    Runtime.getRuntime().addShutdownHook(new Thread(this::stop));
```
JLLeitschuh: This still needs to be made a daemon thread.

SamCarlberg: That's not how the shutdown hook API works. This is the proper way to do it

PR #557, opened 2016-03-30
Moves type resolution in Operations to SocketHint

JLLeitschuh:
SocketHints already know what type they are so they can be used
to safely perform the casting of the sockets passed to the perform method.


core/src/main/java/edu/wpi/grip/core/operations/composite/DesaturateOperation.java
```diff
@@ -54,9 +51,9 @@ public Category getCategory() {
     @Override
     @SuppressWarnings("unchecked")
```
nightpool: this can be removed, right?

JLLeitschuh: I think so, yea. Good catch. Thanks!

JLLeitschuh:
This is being done differently by @SamCarlberg so this PR is now irrelevant.

PR #565, opened 2016-04-06
Socket now holds Value instead of Optional

JLLeitschuh:
This will allow us to move publish infrastructure into the sockets value.


SamCarlberg:
I think `Value` should only be used as a return type, not a parameter type. It looks like right now you're just replacing `Optional` with `Value`, but `Optional` was being misused to begin with so you may as well fix that if you're going to change it out

JLLeitschuh:
I suppose you're right on that front.

AustinShalit:

PR #567, opened 2016-04-12
Tests now use MockFrameGrabberFactory so tests wont hang

JLLeitschuh:


core/src/main/java/edu/wpi/grip/core/sources/GRIPSourcesModule.java
```diff
@@ -0,0 +1,13 @@
...
+
+import com.google.inject.AbstractModule;
+
+/**
+ * Adds bindings for hardware that is required by {@link edu.wpi.grip.core.Source Sources}
+ */
+public class GRIPSourcesModule extends AbstractModule {
+    @Override
+    protected void configure() {
+        bind(CameraSource.FrameGrabberFactory.class).to(CameraSource.FrameGrabberFactoryImpl.class);
```
SamCarlberg: If you're going to call this `GRIPSourcesModule`, it should handle everything for every source -- not just the `FrameGrabberFactory`

PR #571, opened 2016-04-23
Pipeline: Use preferred size for min and max for sources window.

PeterJohnson:
Without this, the sources window shrunk when enough processing steps were
added, making it at first hard to read and then impossible to use.


JLLeitschuh:
Thanks, I've been meaning to fix this for a while now.

PR #572, opened 2016-04-23
Add NetworkTable Source.

PeterJohnson:
This reads a boolean, number, or string value from a NetworkTable key and
makes it available to GRIP.

### DIFF
--- a/core/src/main/java/edu/wpi/grip/core/sources/NetworkValueSource.java
+++ b/core/src/main/java/edu/wpi/grip/core/sources/NetworkValueSource.java
@@ -1,0 +1,199 @@
+package edu.wpi.grip.core.sources;
+
+
+import com.google.common.base.StandardSystemProperty;
+import com.google.common.eventbus.EventBus;
+import com.google.common.eventbus.Subscribe;
+import com.google.inject.assistedinject.Assisted;
+import com.google.inject.assistedinject.AssistedInject;
+import com.thoughtworks.xstream.annotations.XStreamAlias;
+import edu.wpi.first.wpilibj.networktables.NetworkTable;
+import edu.wpi.first.wpilibj.tables.ITable;
+import edu.wpi.grip.core.sockets.OutputSocket;
+import edu.wpi.grip.core.sockets.SocketHint;
+import edu.wpi.grip.core.sockets.SocketHints;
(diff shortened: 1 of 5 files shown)


PeterJohnson:
Note: I didn't create a new source icon for this so that should be changed before merging.

core/src/main/java/edu/wpi/grip/core/sources/NetworkValueSource.java
```diff
@@ -0,0 +1,199 @@
...
+            // We have a new value then we need to update the socket value
+            if (!newValue.equals(stringPrevValue)) {
+                stringPrevValue = newValue;
+                // Update the value
+                stringOutputSocket.setValue(newValue);
+                // We have updated output sockets
+                return true;
+            } else {
+                return false; // No output sockets were updated
+            }
```
JLLeitschuh: There is a lot of duplicate code here. Is there any way to generify it?

PeterJohnson: I was thinking of refactoring the type-specific handling to an internal class to make it a bit cleaner and get rid of the null assignments, but I don't know how much code can be deduplicated because fundamentally there are different types and thus different function calls required in this section in particular.

JLLeitschuh:
The thing that I dislike about this PR is that network tables creates static state which can cause cascading test failures that are really hard to debug/track down.

I've been working on trying to isolate network tables from our class hierarchy by exclusively having any reference to it in our `NTManager` class.

This means that I can provide a mock `NTManager` to any test that needs it. I don't know if you think that this is possible.
Please feel free to come chat with us/me in [Gitter](https://gitter.im/WPIRoboticsProjects/GRIP) about this.

PeterJohnson:
I'm certainly okay with changing this to use NTManager, although the amount of introspection used there means it may take me a bit to do it.  Will you be at champs this week?  We can discuss in person too.

JLLeitschuh:
I will not. I have to submit my MQP paper by Thursday and have several other projects I need to finish up before the end of the term. I'm open to a hangouts call or similar at any point. Send me a message on Gitter and we can chat.

AustinShalit:
Closing in favor of #646

PR #573, opened 2016-04-23
Add operation to save image snapshots to local disk.

PeterJohnson:
The period and path prefix and suffix can be set, and the operation enabled
and disabled at will (this is particularly useful in combination with PR #572).

### DIFF
--- a/core/src/main/java/edu/wpi/grip/core/operations/composite/SaveImageOperation.java
+++ b/core/src/main/java/edu/wpi/grip/core/operations/composite/SaveImageOperation.java
@@ -1,5 +1,6 @@
 package edu.wpi.grip.core.operations.composite;
 
+import com.google.common.base.Stopwatch;
 import com.google.common.eventbus.EventBus;
 import edu.wpi.grip.core.sockets.InputSocket;
 import edu.wpi.grip.core.Operation;
@@ -14,7 +15,7 @@ import java.io.InputStream;
 import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.Optional;
-import java.util.Properties;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -32,8 +33,7 @@ public class SaveImageOperation implements Operation {
     private final Object imageLock = new Object();
     private final BytePointer imagePointer = new BytePointer();
     private Optional<Thread> saveThread = Optional.empty();
-    private volatile boolean connected = false;
-    private long startTime = System.nanoTime();
+    private Stopwatch stopwatch = Stopwatch.createStarted();
     private int numSteps = 0;
     private SimpleDateFormat format = new SimpleDateFormat("yyyyMMdd_HHmmss.SSS");
     private String prefix = "./";
@@ -47,7 +47,8 @@ public class SaveImageOperation implements Operation {
         logger.info("Starting image saver");
         byte[] buffer = new byte[128 * 1024];
         int bufferSize;
-        String prefix, suffix;
+        String prefix;
+        String suffix;
         while (!Thread.currentThread().isInterrupted()) {
             try {
                 // Wait for the main thread to put a new image.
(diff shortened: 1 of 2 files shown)


core/src/main/java/edu/wpi/grip/core/operations/composite/SaveImageOperation.java
```diff
@@ -0,0 +1,171 @@
...
+
+                    // Copy the image data into a pre-allocated buffer, growing it if necessary
+                    bufferSize = imagePointer.limit();
+                    if (bufferSize > buffer.length) buffer = new byte[imagePointer.limit()];
+                    imagePointer.get(buffer, 0, bufferSize);
+                    prefix = this.prefix;
+                    suffix = this.suffix;
+                }
+            } catch (InterruptedException e) {
+                break;
```
JLLeitschuh: You need to reset the interrupt flag here.
`Thread.interrupt()`

core/src/main/java/edu/wpi/grip/core/operations/composite/SaveImageOperation.java
```diff
@@ -0,0 +1,171 @@
...
+            }
+
+            // Write the file
+            try {
+                FileOutputStream out = new FileOutputStream(prefix + format.format(new Date()) + suffix);
```
JLLeitschuh: Try with resource would be more appropriate here.
https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

core/src/main/java/edu/wpi/grip/core/operations/composite/SaveImageOperation.java
```diff
@@ -0,0 +1,171 @@
...
+            return;
+        }
+
+        if (input.empty()) {
+            throw new IllegalArgumentException("Input image must not be empty");
+        }
+
+        //if (period.doubleValue() < 0.03) {
+        //    throw new IllegalArgumentException("Delay must be more than 30 ms");
+        //}
```
JLLeitschuh: Why is this not a constraint?
Should this comment be deleted?

JLLeitschuh:
Thanks for this. This is not something I thought people would want. I like the idea though.

core/src/main/java/edu/wpi/grip/core/operations/composite/SaveImageOperation.java
```diff
@@ -0,0 +1,171 @@
...
+ */
+public class SaveImageOperation implements Operation {
+
+    private final Logger logger = Logger.getLogger(SaveImageOperation.class.getName());
+
+    private final Object imageLock = new Object();
+    private final BytePointer imagePointer = new BytePointer();
+    private Optional<Thread> saveThread = Optional.empty();
+    private volatile boolean connected = false;
+    private long startTime = System.nanoTime();
```
JLLeitschuh: Guava has the [stopwatch](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/base/Stopwatch.html) that may be more apropreate here. Up to you though.

SamCarlberg: I don't think this even needs to be timed; it'll only run when the pipeline gets fired anyway.

PeterJohnson:
This is very useful for testing and debugging at competition where you don't have realtime access because you can load up the saved images to play them back through the pipeline and adjust thresholds etc.  I'll fix up based on the comments so far.

core/src/main/java/edu/wpi/grip/core/operations/composite/SaveImageOperation.java
PeterJohnson: Well, it's nice to not necessarily save *every* frame all the time.  For example, my team saves images twice a second during a match but switches to saving every image when actually auto-aiming.

PeterJohnson:
I'm debating adding a directory chooser for the output directory instead of or in addition to the prefix.  However, I'm not sure if adding those sorts of controls is frowned upon in operations.

JLLeitschuh:
No, a directory chooser would be a good control to add.

PR #580, opened 2016-05-07
Explicitly add ui dependencies.

PeterJohnson:
Transitive dependencies through core don't seem to work in eclipse.


JLLeitschuh:
Is there any way that you can store this in a variable so that we don't have these dependencies listed twice?

AustinShalit:
@PeterJohnson Is this still an issue?

JLLeitschuh:
I have a lot more experience in gradle now, I don't think this is the right solution.

PR #584, opened 2016-05-12
Update Gradle to 2.13

JLLeitschuh:


333fred:
That doesn't update the gradle version until you run ./gradlew wrapper. That only updates it the next you run that task. Additionally, 2.13 is out, any reason you can't upgrade fully?

333fred:
Nevermind, another commit was pushed before I saw it again. Still not 2.13 though.

JLLeitschuh:
Good call

PR #591, opened 2016-05-29
Improve IntelliJ indexing of additional Maven repositories

pmalmsten:
IntelliJ has trouble discovering maven repositories nested within other types of dependency blocks in Gradle build scripts. This is fixed by moving them outside of the 'flatDir' block.


JLLeitschuh:
Woot! Thanks!

PR #596, opened 2016-06-02
Fixes issue with CV_Resize operation have fx socket hint twice

PatrickPenguinTurtle:
Before:
cv_resize_before
After: 
cv_resize_after
Note that the blue rectangles are added to draw attention to the correct location and were not present during the running of GRIP. Also the operation worked correctly and the hint text was the only thing that was wrong.


SamCarlberg:
Good catch.

@JLLeitschuh this is good to be merged

PR #598, opened 2016-06-02
Mac keyboard shortcuts now use command

AustinShalit:
Mac keyboard shortcuts used control before.  Now they use command.  Windows still uses control.

Tested on OS X 10.11.5 and Window 10.


JLLeitschuh:
Can you explain why this works this way? What up vs down does?

SamCarlberg:
Up means that button is up (not pressed). Down means it's pressed

AustinShalit:
Yup.  Just like @SamCarlberg said.  Control is the control key but the shortcut modifier is used to represent the modifier key which is used commonly in keyboard shortcuts on the host platform.

PR #529, opened 2016-02-29
Added PlaceRectangle, NumberThreshold, CountContours, and Crop Operations

MeRPG:
Place rectangle places a red rectangle between two points.
Number threshold returns a Boolean by running a number through a threshold.
CountContours returns a number of how many contours are in a contour report.
Crop can crop an input mat.

PR #600, opened 2016-06-03
Adds Number Threshold, Count Contours, and Crop Operations

JLLeitschuh:
Rewrite of @MeRPG's PR.

Closes #529

### DIFF
--- a/core/src/main/java/edu/wpi/grip/core/operations/Operations.java
+++ b/core/src/main/java/edu/wpi/grip/core/operations/Operations.java
@@ -6,6 +6,7 @@ import com.google.common.eventbus.EventBus;
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
 import com.google.inject.name.Named;
+import edu.wpi.grip.core.OperationDescription;
 import edu.wpi.grip.core.OperationMetaData;
 import edu.wpi.grip.core.events.OperationAddedEvent;
 import edu.wpi.grip.core.operations.composite.*;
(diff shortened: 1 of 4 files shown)


SamCarlberg:
Why do we need an operation to count contours? Why not just add a `size` field to the existing publishing operations?

AustinShalit:
Number Threshold and Count Contours would be good examples of python operations but should not be added as GRIP operations.

I am ok with the crop operation.

core/src/main/java/edu/wpi/grip/core/operations/Operations.java
```diff
@@ -67,6 +76,52 @@
...
+                        OperationDescription.builder()
+                                .name("Crop")
+                                .summary("Crop an image")
+                                .icon(Icon.iconStream("grip"))
+                                .build(),
+                        templateFactory.create(
+                                SocketHints.Inputs.createMatSocketHint("src", false),
+                                SocketHints.Inputs.createPointSocketHint("p1", false),
+                                SocketHints.Inputs.createPointSocketHint("p2", false),
+                                SocketHints.Inputs.createMatSocketHint("dst", true),
```
PatrickPenguinTurtle: Doesn't this make the dst an Input?

SamCarlberg: No. But the Crop operation should be in its own class file like every other custom operation.

SamCarlberg:
Rejected. Custom operations should be in their own class files.

core/src/main/java/edu/wpi/grip/core/operations/Operations.java
PatrickPenguinTurtle: Wait... SocketHints.Inputs.createMatSocketHint() can be used to create a socket hint used for an output? Isn't that a bit misleading?

SamCarlberg: Yep. `SocketHints` needs an overhaul (and Javadocs).

JLLeitschuh: How so? What are you thinking about this? I agree but I want to know what your thoughts are.
Open an issue and we can discuss that there.

AustinShalit:
I am going to close this PR because it does not look like this is going to happen in the near future. Please reopen if you have new information.

PR #603, opened 2016-06-04
Re-enables disabled UI CI tests

JLLeitschuh:


JLLeitschuh:
Conflicts. Probably won't ever get around to fixing these.

PR #605, opened 2016-06-08
Add checkstyle plugin to enforce consistency

SamCarlberg:
This also makes CI builds fail outright if a pull request does not meet the checkstyle standards.

I'd also like to have compilation tasks depend on checkstyle, but for some reason the checkstyle tasks depend on compilation...


SamCarlberg:
Wow, TIL git just doesn't notice when files change capitalization

checkstyle.xml
```diff
@@ -0,0 +1,217 @@
...
+        <module name="AbbreviationAsWordInName">
+            <property name="ignoreFinal" value="true"/>
+            <property name="allowedAbbreviationLength" value="4"/> <!-- Allow 'GRIP' in all caps -->
+        </module>
+        <module name="OverloadMethodsDeclarationOrder"/>
+        <module name="VariableDeclarationUsageDistance">
+            <property name="allowedDistance" value="5"/>
+        </module>
+        <module name="CustomImportOrder">
+            <property name="specialImportsRegExp" value="^.*$"/>
```
AustinShalit: Should this be changed to `edu.wpi.grip`?  This will separate project imports from out of project imports.

SamCarlberg: That may be a good idea. I'm not sure if checkstyle will like it, though -- this was the only way I could find to have _all_ imports ordered alphabetically

checkstyle.xml
```diff
@@ -0,0 +1,211 @@
...
+            <property name="format" value="\\u00(08|09|0(a|A)|0(c|C)|0(d|D)|22|27|5(C|c))|\\(0(10|11|12|14|15|42|47)|134)" />
+            <property name="message" value="Avoid using corresponding octal or Unicode escape." />
+        </module>
+        <module name="AvoidEscapedUnicodeCharacters">
+            <property name="allowEscapesForControlCharacters" value="true" />
+            <property name="allowByTailComment" value="true" />
+            <property name="allowNonPrintableEscapes" value="true" />
+        </module>
+        <module name="LineLength">
+            <property name="max" value="200" />
```
JLLeitschuh: I actually like the 100 max. It makes github reviewing easier to read.

checkstyle.xml
```diff
@@ -0,0 +1,211 @@
+<?xml version="1.0"?>
```
JLLeitschuh: Can you use the original styling of this file so that I can diff it easily.

checkstyle.xml
```diff
@@ -0,0 +1,211 @@
...
+            <message key="name.invalidPattern"
+                     value="Member name ''{0}'' must match pattern ''{1}''." />
+        </module>
+        <module name="ParameterName">
+            <property name="format" value="^[a-z][a-z0-9]*[a-zA-Z0-9]*$" />
+            <message key="name.invalidPattern"
+                     value="Parameter name ''{0}'' must match pattern ''{1}''." />
+        </module>
+        <module name="CatchParameterName">
+            <property name="format" value="^[a-z][a-z0-9]*[a-zA-Z0-9]*$" />
```
JLLeitschuh: Doesn't that star invalidate the rest of the regex.
Also, I like the original google format.

checkstyle.xml
JLLeitschuh: You can get intellij to format all of the files this way using a styleguide and a mass application of it to all of the files.

checkstyle.xml
SamCarlberg: This allows for catch blocks like `catch (SomeException e)` that the google guide doesn't allow.
And no, neither star invalidates anything. The first one could use a lazy qualifier, but it really won't make a difference.

checkstyle.xml
SamCarlberg: First point: that makes sense. I personally prefer longer lines because I have a big screen, but since GitHub limits the line length may as well stick to 100. That'll cause some funky formatting, though.

Second point: That's why this diff is at +4200 -3200 lines changed

AustinShalit:
:+1:

PR #607, opened 2016-06-17
Update WatershedOperation

SamCarlberg:
Update WatershedOperation to return contours for the features discovered by watershed, instead of (a rather unuseful) Mat.

### DIFF
--- a/core/src/main/java/edu/wpi/grip/core/operations/composite/WatershedOperation.java
+++ b/core/src/main/java/edu/wpi/grip/core/operations/composite/WatershedOperation.java
@@ -11,25 +11,28 @@ import edu.wpi.grip.core.util.Icon;
 import com.google.common.collect.ImmutableList;
 
 import java.nio.ByteBuffer;
-import java.util.LinkedHashMap;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Function;
 
 import static org.bytedeco.javacpp.opencv_core.CV_32SC1;
-import static org.bytedeco.javacpp.opencv_core.CV_8U;
 import static org.bytedeco.javacpp.opencv_core.CV_8UC1;
 import static org.bytedeco.javacpp.opencv_core.CV_8UC3;
 import static org.bytedeco.javacpp.opencv_core.LINE_8;
 import static org.bytedeco.javacpp.opencv_core.Mat;
 import static org.bytedeco.javacpp.opencv_core.MatVector;
 import static org.bytedeco.javacpp.opencv_core.Point;
+import static org.bytedeco.javacpp.opencv_core.Point2f;
 import static org.bytedeco.javacpp.opencv_core.Scalar;
+import static org.bytedeco.javacpp.opencv_core.bitwise_xor;
 import static org.bytedeco.javacpp.opencv_imgproc.CV_CHAIN_APPROX_TC89_KCOS;
 import static org.bytedeco.javacpp.opencv_imgproc.CV_FILLED;
 import static org.bytedeco.javacpp.opencv_imgproc.CV_RETR_EXTERNAL;
 import static org.bytedeco.javacpp.opencv_imgproc.circle;
 import static org.bytedeco.javacpp.opencv_imgproc.drawContours;
 import static org.bytedeco.javacpp.opencv_imgproc.findContours;
+import static org.bytedeco.javacpp.opencv_imgproc.pointPolygonTest;
 import static org.bytedeco.javacpp.opencv_imgproc.watershed;
 
 /**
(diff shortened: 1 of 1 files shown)


core/src/main/java/edu/wpi/grip/core/operations/composite/WatershedOperation.java
```diff
@@ -35,11 +53,15 @@
             .initialValueSupplier(ContoursReport::new)
             .build();
 
-    private final SocketHint<Mat> outputHint = SocketHints.Inputs.createMatSocketHint("Output", true);
+    @SuppressWarnings("unchecked")
+    private final SocketHint<ContoursReport> outputHint = new SocketHint.Builder<>(ContoursReport.class)
+            .identifier("Segmented Contours")
```
SamCarlberg: This is a bit too long; it gets cut off and rendered as "`Segmented C...`" on my machine. Maybe just "`Features`"?

JLLeitschuh:
You're going to need to rebase this so that it is just the one commit. Currently the diff is to large to correctly see what has been changed.

SamCarlberg:
Wait what, it was rebased. Weird.

core/src/main/java/edu/wpi/grip/core/operations/composite/WatershedOperation.java
```diff
@@ -91,19 +103,51 @@ public void perform() {
...
+      int bufIdx;
+      for (int y = 0; y < output.rows(); y++) {
+        for (int x = 0; x < output.cols(); x++) {
+          bufIdx = (stride * y) + (x * markers.channels());
+          byte val = buffer.get(bufIdx);
+          if (val == 0 || val == -1) {
+            // Background (0) or border (-1)
+            continue;
+          }
+          Mat m = segments.computeIfAbsent(val, k -> {
```
JLLeitschuh: When does this mat get used?

SamCarlberg: It's a temporary Mat used to find individual contours from the watershed.

SamCarlberg:
Crap, still getting segfaults. Only seems to happen when:

1. There are a ton of contours (> 100); and
2. The `FindContours` operation doesn't use `External only`

Tons of external-only contours either works normally or throws an exception (if `num_contours > 253`).  
Fewer non-`External only` contours also throws an exception. That gets caught [here](https://github.com/WPIRoboticsProjects/GRIP/pull/607/files#diff-452222b1cce028acf7b1292ab7bea63fR151), which _should_ also catch the conditions for the segfaults before they happen

core/src/main/java/edu/wpi/grip/core/operations/composite/WatershedOperation.java
```diff
@@ -85,29 +97,85 @@ public void perform() {
...
+      }
+    }
+    if (!found) {
+      // Should only happen if the image is clogged with contours
+      throw new IllegalStateException("Could not find a point for the background label");
+    }
+    return backgroundLabel;
+  }
+
+  private Point fromPoint2f(Point2f p) {
```
JLLeitschuh: Make this static?

core/src/main/java/edu/wpi/grip/core/operations/composite/WatershedOperation.java
```diff
@@ -85,29 +97,85 @@ public void perform() {
...
     } finally {
       // make sure that the working mat is freed to avoid a memory leak
       markers.release();
     }
   }
 
+  /**
+   * Finds the first available point to place a background marker for the watershed operation.
+   */
+  private Point2f findBackgroundMarker(Mat markers, MatVector contours) {
```
JLLeitschuh: Can this be static?

SamCarlberg: It could, but there's no real need.

JLLeitschuh: It doesn't use member variables so it should be static.

core/src/main/java/edu/wpi/grip/core/operations/composite/WatershedOperation.java
```diff
@@ -85,29 +97,85 @@ public void perform() {
...
   }
 
+  /**
+   * Finds the first available point to place a background marker for the watershed operation.
+   */
+  private static Point2f findBackgroundMarker(Mat markers, MatVector contours) {
+    final int cols = markers.cols();
+    final int rows = markers.rows();
+    final int minDist = 5;
+    Point2f backgroundLabel = new Point2f();
```
JLLeitschuh: Is this a memory resource that should be closed once done with it?

SamCarlberg: There's no need to manually free it, the GC will get it.

JLLeitschuh:
LGTM! We'll need to mention that this is a breaking change in the release notes.

PR #609, opened 2016-06-28
Update xstream, auto-value, guice, guice-assistedinject, sshj

AustinShalit:
This updates the versions of xstream, auto-value, guice, guice-assistedinject, and sshj.


JLLeitschuh:
Not sure why appveyor failed to build.

PR #611, opened 2016-07-01
Suppress Jetty logger in tests

SamCarlberg:


core/src/test/java/edu/wpi/grip/core/http/NoLogger.java
```diff
@@ -0,0 +1,84 @@
...
+
+  }
+
+  @Override
+  public void debug(String msg, long value) {
+
+  }
+
+  @Override
+  public void debug(Throwable thrown) {
```
JLLeitschuh: Can you suppress this warning on the entire class?

PR #613, opened 2016-07-07
Add support for timing steps and pipelines

SamCarlberg:
This lets users know which parts of their pipelines take the most time, and how long the pipeline takes as a whole.


ui/src/main/java/edu/wpi/grip/ui/MainWindowController.java
```diff
@@ -59,6 +66,7 @@
   private Pane aboutPane;
   @FXML
   private StatusBar statusBar;
+  private Label elapsedTimeLabel; // TODO improve the layout of this label
```
SamCarlberg: This label isn't centered vertically in the status bar, and it separates the start/stop button from the "`Pipeline STATE`" text.

ui/src/main/java/edu/wpi/grip/ui/pipeline/StepController.java
```diff
@@ -46,6 +53,8 @@
   @FXML
   private Labeled title;
   @FXML
+  private Label elapsedTime;
```
SamCarlberg: This label gets a little cramped with the operation icon directly above it. Not really sure what the best way would be to fix this.

SamCarlberg:
It may also be a good idea to collect the timing data and have a UI for displaying statistics about each operation, such as

- Mean runtime
- Median runtime
- Relative mean runtime (% of overall time the mean runtime takes)
- Relative median runtime (% of overall time the median runtime takes)
- GRIP overhead (% of time spend not running any operation)

JLLeitschuh:
Can you maybe put those stats in the right side of the status bar?

SamCarlberg:
I think it makes more sense on the left side, that stuff is all related and it makes sense to keep them together. It's also not great for UX

ui/src/main/java/edu/wpi/grip/ui/MainWindowController.java
AustinShalit: I would move the Pipeline status into its own label and then center both (i.e. do not use statusBar.setText() for the status)

core/src/main/java/edu/wpi/grip/core/metrics/Analysis.java
```diff
@@ -0,0 +1,90 @@
...
+ */
+public class Analysis {
+
+  // <0: error
+  // 0 = no change from original state
+  // 1 = returns most recent input
+  // >1: error
+  private static final double alpha = 0.35;
+
+  private final ExponentialMovingAverage ema = new ExponentialMovingAverage(alpha);
```
SamCarlberg: Maybe change this to a normal arithmetic moving average, the output of this still isn't very consistent.

AustinShalit:
When I open the analysis window I get a stack overflow [pastebin](http://pastebin.com/FXX6p5Rd)

SamCarlberg:
Does it only happen when you move the window?

AustinShalit:
Here is another log and a video: http://pastebin.com/JinS2Frq

https://youtu.be/BcppMd6XX-Q

SamCarlberg:
That looks like a JavaFX bug to me

core/src/main/java/edu/wpi/grip/core/events/TimerEvent.java
```diff
@@ -0,0 +1,59 @@
...
+   * @param source      the object that was timed
+   * @param elapsedTime the time elapsed in microseconds
+   * @param data        the analysis data of the event
+   */
+  public TimerEvent(T source, long elapsedTime, Analysis data) {
+    checkNotNull(source, "source");
+    checkNotNull(data, "data");
+    this.source = source;
+    this.elapsedTime = elapsedTime;
+    this.data = data;
```
JLLeitschuh: `checkNotNull(source, "source");` returns the source so you can just use it in the same line as the assignment.

core/src/main/java/edu/wpi/grip/core/metrics/Analysis.java
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * Analysis data of a timed action. Contains
+ * <ul>
+ * <li>The number of samples taken</li>
+ * <li>The average of the samples taken. This may not be an arithmetic mean, or even an average
+ * of <i>all</i> the samples</li>
+ * <li>Statistical analysis of the data.</li>
+ * </ul>
+ */
+public class Analysis {
```
JLLeitschuh: Can this be final??

core/src/main/java/edu/wpi/grip/core/metrics/ExponentialMovingAverage.java
```diff
@@ -0,0 +1,63 @@
...
+  private double lastValue = 0;
+  private final double alpha;
+
+  /**
+   * Creates a new exponential moving average filter with the given alpha value
+   *
+   * @param alpha the smoothing factor used. Must be in the range (0, 1).
+   */
+  ExponentialMovingAverage(double alpha) {
+    checkArgument(0 <= alpha && alpha <= 1, "0 ≤ alpha ≤ 1");
```
JLLeitschuh: Do you want to put the value of `alpha` somewhere in that string?

core/src/main/java/edu/wpi/grip/core/events/TimerEvent.java
SamCarlberg: This is cleaner IMO

core/src/main/java/edu/wpi/grip/core/metrics/Analysis.java
SamCarlberg: Yes

core/src/main/java/edu/wpi/grip/core/events/TimerEvent.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  /**
+   * Gets the source object that had an action being timed.
+   */
+  public T getSource() {
```
JLLeitschuh: This seems ambiguous with `Source` as the input to the pipeline.

core/src/main/java/edu/wpi/grip/core/metrics/BenchmarkRunner.java
```diff
@@ -0,0 +1,104 @@
...
+   * Benchmarks the pipeline.
+   *
+   * @param numRuns the number of runs of the pipeline to collect events from.
+   */
+  public void run(int numRuns) {
+    checkArgument(numRuns > 0, "Must run the pipeline at least once");
+    this.isBenchmarking.set(true);
+    this.runsRemaining.set(numRuns);
+    eventBus.post(BenchmarkEvent.started());
+    eventBus.post(new BenchmarkRunEvent());
```
JLLeitschuh: Why are these two separate events?

core/src/main/java/edu/wpi/grip/core/metrics/HotnessCalculator.java
```diff
@@ -0,0 +1,23 @@
+package edu.wpi.grip.core.metrics;
+
+/**
+ * Analyzes step and pipeline timings.
+ */
+public class HotnessCalculator {
```
JLLeitschuh: Ummmmm. Hahaha wut?

core/src/main/java/edu/wpi/grip/core/metrics/HotnessCalculator.java
```diff
@@ -0,0 +1,23 @@
...
+
+  /**
+   * Calculates the 'hotness' of the given value based on accrued statistics.
+   *
+   * @param statistics the accrued statistics to calculate the hotness from
+   * @param value      the value to calculate the hotness of
+   * @return the hotness of the given value.
+   */
+  public double hotness(Statistics statistics, double value) {
```
JLLeitschuh: This method doesn't use any member variables. This should really be static.

core/src/main/java/edu/wpi/grip/core/metrics/Statistics.java
```diff
@@ -0,0 +1,92 @@
...
+
+  private final int n;
+  private final double sum;
+  private final double mean;
+  private final double s;
+
+  /**
+   * "null" statistics with every value set to zero.
+   */
+  public static final Statistics nil = new Statistics(0, 0, 0, 0);
```
JLLeitschuh: Make `nill` all caps? Thats what we've been doing with other public final variables.

core/src/main/java/edu/wpi/grip/core/metrics/Statistics.java
```diff
@@ -0,0 +1,92 @@
...
+    for (double d : samples) {
+      sum += d;
+    }
+    final double mean = n == 0 ? 0 : sum / n;
+    double variance = 0;
+    for (double d : samples) {
+      variance += ((d - mean) * (d - mean)) / n;
+    }
+    final double s = Math.sqrt(variance);
+    return new Statistics(n, sum, mean, s);
```
JLLeitschuh: I like this!

core/src/main/java/edu/wpi/grip/core/metrics/Statistics.java
```diff
@@ -0,0 +1,92 @@
...
+   */
+  public static final Statistics nil = new Statistics(0, 0, 0, 0);
+
+  /**
+   * Calculates the statistics of the given samples.
+   *
+   * @param samples the samples to analyze
+   * @return a statistical analysis of the given samples
+   */
+  public static Statistics of(double... samples) {
```
JLLeitschuh: I like this method! It keeps the constructor as it should be, simple assignments.

core/src/main/java/edu/wpi/grip/core/metrics/Statistics.java
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Statistics analysis. Contains:
+ * <ul>
+ * <li>Number of samples ({@code n})</li>
+ * <li>Mean value ({@code mean})</li>
+ * <li>Standard deviation ({@code s})</li>
+ * </ul>
+ */
+public final class Statistics {
```
JLLeitschuh: You can annotate this class as `Immutable` if you want.

core/src/test/java/edu/wpi/grip/core/metrics/TimerTest.java
```diff
@@ -0,0 +1,87 @@
...
+    timer.stopped();
+    timer.stopped();
+    fail("Timer should have thrown an exception");
+  }
+
+  @Test
+  public void testTiming() throws InterruptedException {
+    Timer timer = new Timer(eventBus, this);
+    timer.started();
+    Thread.sleep(1000);
```
JLLeitschuh: Don't do this. This is why the `Stopwatch` API has a constructor that takes a `Ticker`. Explicitly so you don't have to do this.

core/src/test/java/edu/wpi/grip/core/metrics/TimerTest.java
```diff
@@ -0,0 +1,87 @@
...
+    Timer timer = new Timer(eventBus, this);
+    ThreadLocal<TimerEvent> timerEvent = new ThreadLocal<>();
+    eventBus.register(new Object() {
+      @Subscribe
+      private void onTimerEvent(TimerEvent event) {
+        timerEvent.set(event);
+      }
+    });
+    timer.started();
+    Thread.sleep(1000);
```
JLLeitschuh: Don't do this either.

ui/src/main/java/edu/wpi/grip/ui/analysis/AnalysisWindowController.java
```diff
@@ -0,0 +1,167 @@
...
+      stepData.add(Pair.of(source, event.getData()));
+      analysis.add(event.getData().getAverage());
+    }
+  }
+
+  @Subscribe
+  @SuppressWarnings({"PMD.UnusedPrivateMethod", "PMD.UnusedFormalParameter"})
+  private void onStepRemoved(StepRemovedEvent event) {
+    stepData.stream()
+        .filter(p -> p.getLeft() == event.getStep()) // Want reference equality
```
JLLeitschuh: Why?

core/src/main/java/edu/wpi/grip/core/events/TimerEvent.java
SamCarlberg: Maybe call it `target`?

core/src/main/java/edu/wpi/grip/core/metrics/BenchmarkRunner.java
SamCarlberg: Basic event flow:

1. BenchmarkEvent.started()
2. BenchmarkRunEvent
3. RunStartedEvent
4. RunStoppedEvent
5. Repeat 2-4 until the number of runs have completed (or it was cancelled)
6. BenchmarkEvent.stopped()

core/src/main/java/edu/wpi/grip/core/metrics/HotnessCalculator.java
SamCarlberg: ? I mean, this could just be refactored into a function at this point

core/src/test/java/edu/wpi/grip/core/metrics/TimerTest.java
SamCarlberg: Ah crap, I keep forgetting about that.

ui/src/main/java/edu/wpi/grip/ui/analysis/AnalysisWindowController.java
SamCarlberg: Doesn't matter, it's getting removed

core/src/test/java/edu/wpi/grip/core/metrics/TimerTest.java
JLLeitschuh: Why are you delaying?

SamCarlberg: Because otherwise it's impossible to know if it's actually measuring elapsed time? It's either use a delay or an ugly busy-wait.

core/src/test/java/edu/wpi/grip/core/metrics/BenchmarkRunnerTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Tests for {@link BenchmarkRunner}.
+ */
+public class BenchmarkRunnerTest {
+
+  private EventBus eventBus;
+  private BenchmarkRunner benchmarkRunner;
+
+  @Before
```
SamCarlberg: Why is this even getting reported??? This is the _only_ public test method getting this; another with _exactly the same signature_ isn't getting flagged for this. Nor is this breaking the build (as it should if checkstyle doesn't pass, so I'm guessing codacy is just broken)

core/src/test/java/edu/wpi/grip/core/metrics/TimerTest.java
JLLeitschuh: The stopwatch API allows you to "Mock out" time passing. So you don't need to do this.

core/src/test/java/edu/wpi/grip/core/metrics/BenchmarkRunnerTest.java
JLLeitschuh: Codacy will be removed when PMD is added.

SamCarlberg: Good, it's incredibly  annoying

AustinShalit:
The stackoverflow issue I was seeing has been solved!  Thank you @SamCarlberg!

ui/src/main/resources/edu/wpi/grip/ui/analysis/AnalysisWindow.fxml
```diff
@@ -0,0 +1,36 @@
...
+<?import javafx.scene.control.*?>
+<?import java.lang.*?>
+<?import javafx.scene.layout.*?>
+
+<AnchorPane maxHeight="-Infinity" maxWidth="-Infinity" minHeight="-Infinity" minWidth="-Infinity" prefHeight="400.0" prefWidth="600.0" xmlns="http://javafx.com/javafx/8" xmlns:fx="http://javafx.com/fxml/1" fx:controller="edu.wpi.grip.ui.analysis.AnalysisWindowController">
+   <children>
+      <TableView fx:id="table" prefHeight="400.0" prefWidth="600.0" AnchorPane.bottomAnchor="50.0" AnchorPane.leftAnchor="0.0" AnchorPane.rightAnchor="0.0" AnchorPane.topAnchor="0.0">
+        <columns>
+          <TableColumn fx:id="operationColumn" editable="false" minWidth="200.0" prefWidth="300.0" resizable="false" sortable="false" text="Operation" />
+          <TableColumn fx:id="timeColumn" editable="false" minWidth="200.0" prefWidth="300.0" sortable="false" text="Time (ms)" />
```
AustinShalit: The "(ms)" is redundant

AustinShalit:
PMD has been added to the build scripts.  Please merge 'Master' into this PR.  Thanks!

JLLeitschuh:
I want to thoroughly review this and actually test this before merging it.
I don't think I'll have time to do that until at least tonight, or, if not tonight, then Wednesday night.

core/src/main/java/edu/wpi/grip/core/metrics/Timer.java
```diff
@@ -0,0 +1,131 @@
...
+ *   Timer myTimer = ...
+ *   myTimer.time(this::doSomething);
+ * </code></pre>
+ * </p>
+ */
+public class Timer {
+
+  private final EventBus eventBus;
+  private final Stopwatch stopwatch;
+  private boolean started = false;
```
AustinShalit: private variable started can be replaced with a call to stopwatch.isRunning()

https://github.com/google/guava/blob/master/guava/src/com/google/common/base/Stopwatch.java

SamCarlberg:
@JLLeitschuh This is ready for a final review

core/src/main/java/edu/wpi/grip/core/metrics/Analysis.java
```diff
@@ -0,0 +1,102 @@
...
+ * <li>The number of samples taken</li>
+ * <li>The average of the samples taken. This may not be an arithmetic mean, or even an average
+ * of <i>all</i> the samples</li>
+ * <li>Statistical analysis of the data.</li>
+ * </ul>
+ */
+@Immutable
+public final class Analysis {
+
+  private static final int windowSize = 10;
```
JLLeitschuh: All caps with underscores so that it's clearly a static final variable.
(This should probably be in checkstyle).

Also, can you document what this means?

core/src/main/java/edu/wpi/grip/core/metrics/BenchmarkRunner.java
```diff
@@ -0,0 +1,131 @@
...
+
+  @Subscribe
+  @SuppressWarnings("PMD.UnusedPrivateMethod")
+  private void onTimerEvent(TimerEvent event) {
+    this.timers.get().add(event.getTimer());
+  }
+
+  @Subscribe
+  @SuppressWarnings({"PMD.UnusedPrivateMethod", "PMD.UnusedFormalParameter"})
+  private void onRunStart(@Nullable RunStartedEvent event) {
```
JLLeitschuh: I thought you could only subscribe public methods. Am I wrong about this?

SamCarlberg: No, it's reflective access so the event bus makes the method accessible when adding it to the subscribers list.

core/src/main/java/edu/wpi/grip/core/metrics/Analysis.java
SamCarlberg: I'm not very happy with this because it makes it impossible to set the number of data points to average. This needs a serious refactor.

JLLeitschuh:
Can you add the column with the results from the last benchmark??

core/src/main/java/edu/wpi/grip/core/metrics/FixedSizeQueue.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.ConcurrentModificationException;
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+/**
+ * A queue implementation that, once full, removes the oldest element before adding a new one.
+ */
+public class FixedSizeQueue<E> extends AbstractQueue<E> {
```
JLLeitschuh: Cool! I like it!
However, google beat you to it on this one:
https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/EvictingQueue.java

SamCarlberg: Haha seriously

SamCarlberg:
>Can you add the column with the results from the last benchmark??

I think I'll make it CSV in a TextArea to make it easy to import into stuff like excel and google sheets

core/src/main/java/edu/wpi/grip/core/StepIndexer.java
```diff
@@ -0,0 +1,29 @@
...
+  int indexOf(Step step);
+
+  /**
+   * Compares two steps based on their indexes. <i>This is not consistent with {@code equals()}</i>.
+   *
+   * @param o1 the first step to compare
+   * @param o2 the second step to compare
+   */
+  @Override
+  default int compare(Step o1, Step o2) {
```
JLLeitschuh: Interesting solution.

core/src/main/java/edu/wpi/grip/core/metrics/Timer.java
JLLeitschuh: I agree

core/src/main/java/edu/wpi/grip/core/metrics/Timer.java
```diff
@@ -0,0 +1,124 @@
...
+   */
+  public void time(Runnable target) {
+    if (started) {
+      throw new IllegalStateException("This timer is already timing something");
+    }
+    try {
+      started();
+      target.run();
+    } finally {
+      stopped();
```
JLLeitschuh: Do you want to catch any potential exceptions this may throw?

core/src/main/java/edu/wpi/grip/core/metrics/Timer.java
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  /**
+   * Flags the timer as stopped. This will post a {@link TimerEvent} containing the elapsed time
+   * and analysis to the event bus.
+   *
+   * @throws IllegalStateException if this a call to this method is not preceded by a call to
+   *                               {@link #started()}.
+   */
+  public synchronized void stopped() {
```
JLLeitschuh: Should this be renamed `stop`?
Its an action, `stopped` infers a question.

ui/src/main/java/edu/wpi/grip/ui/MainWindowController.java
```diff
@@ -260,4 +277,43 @@ protected void deploy() {
...
+      AnalysisWindowController controller = loader.getController();
+      controller.setBenchmarker(benchmarkRunner);
+      controller.setStepIndexer(stepIndexer);
+      analysisStage.initOwner(root.getScene().getWindow());
+      analysisStage.setTitle("Pipeline Analysis");
+      analysisStage.getIcons().add(new Image("/edu/wpi/grip/ui/icons/grip.png"));
+      analysisStage.setOnCloseRequest(event -> eventBus.post(BenchmarkEvent.finished()));
+      eventBus.register(controller);
+      analysisStage.showAndWait();
+      eventBus.unregister(controller); // unregister to avoid memory leak
```
JLLeitschuh: How does this work on a high DPI screen?
The fonts get all weird on those sort of screens.

ui/src/main/java/edu/wpi/grip/ui/analysis/AnalysisWindowController.java
```diff
@@ -0,0 +1,339 @@
...
+      sb.append(stepNames.get(i));
+      sb.append(',');
+      sb.append(percentRunTimes.get(i));
+      sb.append(',');
+      sb.append(averageRunTimes.get(i));
+      sb.append(',');
+      sb.append(stddev.get(i));
+      sb.append('\n');
+    }
+    return sb.toString();
```
JLLeitschuh: Are you sure this is correctly formatted CSV?
This isn't using a library to create CSV and there aren't any tests for this code.

ui/src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketController.java
```diff
@@ -59,6 +62,11 @@ protected void initialize() {
     this.type.setText(this.socket.getSocketHint().getTypeLabel());
     this.handle = socketHandleViewFactory.create(this.socket);
     root.add(this.handle, 0, 0);
+    handle.connectedProperty().addListener((observable, oldValue, newValue) -> {
+      if (getContent() != null) {
+        getContent().setDisable(newValue);
```
JLLeitschuh: What is this doing and why was it added?

ui/src/main/java/edu/wpi/grip/ui/pipeline/StepController.java
```diff
@@ -192,6 +206,25 @@ private void moveStepRight() {
     pipeline.moveStep(step, +1);
   }
 
+  @Subscribe
+  @SuppressWarnings("PMD.UnusedPrivateMethod")
```
JLLeitschuh: There must be a way to disable this error when there is this annotation on the method.
Not sure how to do it.

ui/src/main/java/edu/wpi/grip/ui/analysis/AnalysisWindowController.java
```diff
@@ -0,0 +1,339 @@
...
+    void update(double time, double relativeAmount, double hotness) {
+      text.setText(String.format("%.1fms", time / 1e3));
+      progressBar.setProgress(relativeAmount);
+      if (hotness > 0) {
+        final double max = 3; // highest value before being clamped
+        double h = ((max - Math.min(hotness, max)) * 270 / max) - 45;
+        final String formatStyle = "-fx-accent: hsb(%.2f, 100%%, 100%%);";
+        progressBar.setStyle(String.format(formatStyle, h));
+      } else {
+        progressBar.setStyle("-fx-accent: hsb(180, 100%, 75%);"); // blue-gray
```
JLLeitschuh: Want to use css classes to do this?

ui/src/main/java/edu/wpi/grip/ui/analysis/AnalysisWindowController.java
```diff
@@ -0,0 +1,339 @@
...
+              lastStats.hotness(analysis.getAverage()));
+          setGraphic(null);
+          setGraphic(view);
+        }
+      }
+    });
+    table.setItems(tableItems);
+
+    benchmarkRunsField.textProperty().addListener((observable, oldValue, newValue) -> {
+      if ((oldValue.isEmpty() && "0".equals(newValue)) || !newValue.matches("[\\d]*")) {
```
JLLeitschuh: What is this regex checking?

ui/src/main/java/edu/wpi/grip/ui/MainWindowController.java
```diff
@@ -260,4 +277,43 @@ protected void deploy() {
     dialog.setResizable(true);
     dialog.showAndWait();
   }
+
+  @Subscribe
+  @SuppressWarnings({"PMD.UnusedPrivateMethod", "PMD.UnusedFormalParameter"})
+  private void runStopped(TimerEvent event) {
+    if (!(event.getTarget() instanceof PipelineRunner)) {
```
JLLeitschuh: `instanceof` is bad practice. What are you trying to do here?
Can you do something different here?

ui/src/main/java/edu/wpi/grip/ui/analysis/AnalysisWindowController.java
```diff
@@ -0,0 +1,339 @@
...
+        return;
+      }
+      benchmarkButton.setDisable(newValue.isEmpty());
+    });
+  }
+
+  @Subscribe
+  @SuppressWarnings("PMD.UnusedPrivateMethod")
+  private void onRun(TimerEvent event) {
+    if (event.getTarget() instanceof Step) {
```
JLLeitschuh: Can you avoid this instance of check?

core/src/main/java/edu/wpi/grip/core/metrics/Timer.java
SamCarlberg: No, exceptions are propagated up the stack and are caught.

ui/src/main/java/edu/wpi/grip/ui/analysis/AnalysisWindowController.java
SamCarlberg: Checks if the input is a number

ui/src/main/java/edu/wpi/grip/ui/analysis/AnalysisWindowController.java
SamCarlberg: Yes it is

ui/src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketController.java
SamCarlberg: It disables manual inputs when a connection is added

ui/src/main/java/edu/wpi/grip/ui/analysis/AnalysisWindowController.java
JLLeitschuh: I have a plugin from codecov that shows me in the PR what code is covered and what isn't.
This isn't covered by tests.
https://chrome.google.com/webstore/detail/codecov-extension/keefkhehidemnokodkdkejapdgfjmijf/related

core/src/main/java/edu/wpi/grip/core/metrics/Statistics.java
```diff
@@ -0,0 +1,119 @@
...
+        .add("n", n)
+        .add("mean", mean)
+        .add("s", s)
+        .toString();
+  }
+
+  /**
+   * Gets the number of samples.
+   */
+  public int getN() {
```
JLLeitschuh: Can this be something like `sampleCount` or `number` or `dataPoints`? `getN` seems ambiguous.

core/src/main/java/edu/wpi/grip/core/metrics/Statistics.java
```diff
@@ -0,0 +1,119 @@
...
+ * <li>Standard deviation ({@code s})</li>
+ * </ul>
+ */
+@Immutable
+public final class Statistics {
+
+  private final int n;
+  private final double sum;
+  private final double mean;
+  private final double s;
```
JLLeitschuh: Can you rename this to something like `sqrt`?

core/src/main/java/edu/wpi/grip/core/metrics/Timer.java
```diff
@@ -0,0 +1,120 @@
...
+ * Timer for code that gets run.
+ *
+ * <p>Sample usage:
+ * <pre><code>
+ *   Timer myTimer = ...
+ *   myTimer.time(this::doSomething);
+ * </code></pre>
+ * </p>
+ */
+public class Timer {
```
JLLeitschuh: Annotate with `@ThreadSafe`?

core/src/main/java/edu/wpi/grip/core/metrics/Statistics.java
```diff
@@ -0,0 +1,119 @@
...
+   * that is not in the data set used to create these statistics will most likely have a useless
+   * result.
+   *
+   * <p>If this set of statistics was calculated from less than two data points, 'hotness' doesn't
+   * mean anything and this method will always return zero.
+   *
+   * @param value the value to calculate the hotness of
+   * @return the hotness of the given value.
+   */
+  public double hotness(double value) {
```
JLLeitschuh: What is the range of hotness values? 0-10? 0 to 1? This should be documented.

ui/src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketController.java
JLLeitschuh: I thought we already did this? We should add a UI test that checks for when a connection is added you can't manually change the UI.

core/src/main/java/edu/wpi/grip/core/metrics/Statistics.java
SamCarlberg: It's the standard deviation, `s` is a standard symbol

core/src/main/java/edu/wpi/grip/core/metrics/Statistics.java
SamCarlberg: Hotness can be any number >= 0. The current implementation is just the number of standard deviations above the mean

ui/src/main/java/edu/wpi/grip/ui/pipeline/input/InputSocketController.java
SamCarlberg: It's currently implemented in each class (that's a lt of repetition). This just puts it in the common superclass to a) use a listener instead of a binding and b) let me programmatically enable and disable every socket

core/src/main/java/edu/wpi/grip/core/metrics/Statistics.java
SamCarlberg: Hotness is represented by the hue of the bar. The length of the bar represents what percentage of time the step takes. Hotness values over three standard deviations are clamped to avoid having high-sigma steps having the same color as low-sigma steps.

JLLeitschuh: Can you document that a little bit better in the javadoc?

Imagine I don't understand statistics and want to use this class. :p

ui/src/main/java/edu/wpi/grip/ui/analysis/AnalysisWindowController.java
JLLeitschuh: Why do you need to do this instanceof check?

JLLeitschuh:
LGTM!

PR #624, opened 2016-07-15
Adds CPD Plugin

AustinShalit:
Build will fail.  Take a look at the log and let me know what you think.


AustinShalit:
I am not very happy with CPD.  Most of our code is copy and paste with slight changes.  This will just add suppress warnings everywhere.

SamCarlberg:
@AustinShalit It would be nice to use FindBugs, though.

PR #618, opened 2016-07-08
Add Findbugs

AustinShalit:
Closes #615 

I still need to go through and fix failing tests.

PR #627, opened 2016-07-18
Filter Contours Ratio Rounding

osulacam:
Fixes Filter Contours to use double division instead of integer division for calculating the ratio. This allows filter contours to have minimum and maximum ratios that are not integer values.

### DIFF
--- a/core/src/main/java/edu/wpi/grip/core/operations/composite/FilterContoursOperation.java
+++ b/core/src/main/java/edu/wpi/grip/core/operations/composite/FilterContoursOperation.java
@@ -191,7 +191,7 @@ public class FilterContoursOperation implements Operation {
         continue;
       }
 
-      final double ratio = bb.width() / bb.height();
+      final double ratio = (double) bb.width() / bb.height();
       if (ratio < minRatio || ratio > maxRatio) {
         continue;
       }


SamCarlberg:
Nice. Though might it be a good a idea to use best-fit bounding boxes instead of the simple x-y-oriented bounding boxes?

Take a look at [this image](http://docs.opencv.org/master/rotatedrect.png). The current bounding box we use to calculate the ratio will be the blue rectangle, which has a *very* different ratio than the object (the oblong rectangle)

osulacam:
Best fit bounding boxes are useful but may not be what all users want. There could be an option for it but that might complicate the step.

SamCarlberg:
Yeah, it would be hard to change the existing stuff because it's already in use. I'm reluctant to add more stuff to the filter operation because there's just so much crap packed into it already and it makes the UI a mess. @JLLeitschuh what are your thoughts?

JLLeitschuh:
Uhhhhh... An `AdvancedFilterArgs` object that you can pass in?
Just a thought.

osulacam:
We could just make a separate advanced filter contours and simple filter contours operation. The simple would just have area and perimeter, while the advanced could have everything.

osulacam:
Bytedeco RotatedRect is broken so we can't actually do the best fit bounding boxes.

SamCarlberg:
How are they broken? I got them working fine when I was playing around with them.

osulacam:
They don't give a list of points for the vertices and the bounding rectangle wasn't even remotely correct so I couldn't use that either.

SamCarlberg:
`RotatedRect` contains the center point, width, height, and rotation. You can calculate the vertices from that. And how was it so far off? They've always been accurate for me. I assume you're using something like
```
for (int i = 0; i < contours.size(); i++) {
  RotatedRect rr = opencv_imgproc.minAreaRect(contours.get(i));
  // ... 
}
```

core/src/main/java/edu/wpi/grip/core/operations/composite/AdvancedFilterContoursOperation.java
```diff
@@ -165,19 +190,41 @@ public void perform() {
...
 
-      final Rect bb = boundingRect(contour);
-      if (bb.width() < minWidth || bb.width() > maxWidth) {
+      double width;
+      double height;
+      if (rotatedRect) {
+        final opencv_core.RotatedRect bb = minAreaRect(contour);
+        opencv_core.Point2f points = new opencv_core.Point2f(4);
+        bb.points(points);
+        final double rotatedWidth = Math.sqrt( Math.pow(points.position(0).x() - points.position(1)
```
SamCarlberg: Get rid of the space in `Math.sqrt( Math.pow...`

core/src/main/java/edu/wpi/grip/core/operations/composite/AdvancedFilterContoursOperation.java
```diff
@@ -165,19 +190,41 @@ public void perform() {
     for (int i = 0; i < inputContours.size(); i++) {
       final Mat contour = inputContours.get(i);
 
-      final Rect bb = boundingRect(contour);
-      if (bb.width() < minWidth || bb.width() > maxWidth) {
+      double width;
+      double height;
+      if (rotatedRect) {
+        final opencv_core.RotatedRect bb = minAreaRect(contour);
+        opencv_core.Point2f points = new opencv_core.Point2f(4);
```
SamCarlberg: `RotatedRect` and `Point2f` should be imported directly from `org.bytedeco.javacpp.opencv_core` to avoid the `opencv_core.RotatedRect` crap

core/src/main/java/edu/wpi/grip/core/operations/composite/SimpleFilterContoursOperation.java
```diff
@@ -0,0 +1,113 @@
...
+  public static final OperationDescription DESCRIPTION =
+      OperationDescription.builder()
+          .name("Simple Filter Contours")
+          .summary("Find contours matching certain criteria")
+          .category(OperationDescription.Category.FEATURE_DETECTION)
+          .icon(Icon.iconStream("find-contours"))
+          .build();
+
+  private final SocketHint<ContoursReport> contoursHint = new SocketHint.Builder<>(ContoursReport
+      .class)
```
SamCarlberg: Funky formatting

core/src/main/java/edu/wpi/grip/core/operations/composite/SimpleFilterContoursOperation.java
```diff
@@ -0,0 +1,113 @@
...
+
+
+  private final InputSocket<ContoursReport> contoursSocket;
+  private final InputSocket<Number> minAreaSocket;
+  private final InputSocket<Number> minPerimeterSocket;
+
+  private final OutputSocket<ContoursReport> outputSocket;
+
+  @SuppressWarnings("JavadocMethod")
+  public SimpleFilterContoursOperation(InputSocket.Factory inputSocketFactory, OutputSocket.Factory
```
SamCarlberg: Put parameters on separate lines to avoid a newline inside a parameter definition

core/src/main/java/edu/wpi/grip/core/operations/composite/SimpleFilterContoursOperation.java
```diff
@@ -0,0 +1,113 @@
...
+      if (arcLength(contour, true) < minPerimeter) {
+        continue;
+      }
+      outputContours.put(filteredContourCount++, contour);
+    }
+
+    outputContours.resize(filteredContourCount);
+
+    outputSocket.setValue(new ContoursReport(outputContours,
+        inputSocket.getValue().get().getRows(), inputSocket.getValue().get().getCols()));
```
SamCarlberg: Arguments should all be on one line (if it fits), or each should be on its own line

SamCarlberg:
This seems like a breaking change. Old `FilterContours` steps won't be compatible.

osulacam:
The integer division problem was fixed by #618. The rest of this PR will split up filter contours to make it easier to use.

AustinShalit:
@osulacam Is this PR ready for review?

osulacam:
@AustinShalit Yes

SamCarlberg:
I'll take a look at this tonight

JLLeitschuh:
Is it still a breaking change or has that been resolved?

osulacam:
Because this technically creates a two new operations to get rid of an old one it has to be a breaking change. It is still a necessary change that shouldn't break anything significantly.

Issue #413, opened 2016-01-22
Only show connectable sockets


JLLeitschuh:
SocketHint should have a boolean to hold whether it is connetable.

The `SocketHandleView` should check that value and be hidden based upon that boolean.

tom131313:
I had recognized 3 types of sockets and suggest 3 different appearances for inputs, outputs, and not really a socket at this time.
I like cues for usage (not knowing I had to connect the dots was the only criticism I had of my first 5 seconds with GRIP, I couldn't figure that out without help) and the male/female is always a good model.  Make the output plug mate with an input receptacle.
The 3rd (not a socket) should still be retained as an icon to denote that line is different than the other lines - like bullet points on a chart.  Thanks

AustinShalit:
@JLLeitschuh Do you think this is solved with the latest updates?

JLLeitschuh:
Yes

Issue #327, opened 2016-01-09
Another Useful option "Minimize Operation"


cpapplefamily:
If each operation had an option to only display the used Input and Output connectors with a limited label the box could be reduced to a fraction of the size saving screen restate. 

? Is this the right place for these ideas?

JLLeitschuh:
Yes, this is the right place for the ideas. If you want to discuss how to actually implement these ideas then join us in the [gitter chat](https://gitter.im/WPIRoboticsProjects/GRIP)

Could you provide a sketch for clarity?

cpapplefamily:
Full Sized
![full size](https://cloud.githubusercontent.com/assets/16618289/12214524/191d5970-b65d-11e5-8494-f4d22e79f517.PNG)

cpapplefamily:
Minimized
![minimized](https://cloud.githubusercontent.com/assets/16618289/12214527/2e30c446-b65d-11e5-958f-fba0f29aada0.png)

cpapplefamily:
Could Stack them too then.

JLLeitschuh:
Nice photoshop! This makes a lot more sense now!!!

PR #633, opened 2016-07-21
Creates expanding input socket lists for large operations

osulacam:
This makes the ui much easier to use for filter contours and a few other large operations.


AustinShalit:
Can you please paste a photo?

osulacam:
screen shot 2016-07-21 at 4 09 52 pm
screen shot 2016-07-21 at 4 10 35 pm

AustinShalit:
Nice! A few things:
- Can you make it dynamic? I have a really big display so the steps no longer use the full hight of the pipeline.
- Weird visual things happen when you collapse a section that has a connection in it (see photo)
- For some reason NTManager spams the console now.

untitled

JLLeitschuh:
Uuuuuuuuhhhhh... I'm not sure that this is the route that I would have taken to fix this UI problem.
I feel like there's a better way. I don't know what that is though.

osulacam:
I'm not that experiencde with javafx so I wouldn't be surprised if there was an easier way to do this. Brad really wanted this to get fixed and I defiantly agree that it was making the UI much more difficult to use.
I'll take a look and try to fix the issues that Austin brought up.

AustinShalit:
This is related to #413.  Maybe we go that (413's) route?

AustinShalit:
This is related to #327

osulacam:
I created a new pull request that implements the suggestion from that issue better. I felt like that way was a bit cleaner overall.

Issue #634, opened 2016-07-22
Loading of pipeline with unconnected Switch or Valve


PatrickPenguinTurtle:
If a switch or value in a Pipeline has an unconnected input socket, and the pipeline is saved. When the pipeline is loaded the following occurs:
`Thread[JavaFX Application Thread,5,main] threw an exception
## Stack Trace:
```java
java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
	at javafx.fxml.FXMLLoader$MethodHandler.invoke(FXMLLoader.java:1774)
	at javafx.fxml.FXMLLoader$ControllerMethodEventHandler.handle(FXMLLoader.java:1657)
	at com.sun.javafx.event.CompositeEventHandler.dispatchBubblingEvent(CompositeEventHandler.java:86)
	at com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:238)
	at com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:191)
	at com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:58)
	at com.sun.javafx.event.EventDispatchChainImpl.dispatchEvent(EventDispatchChainImpl.java:114)
	at com.sun.javafx.event.EventUtil.fireEventImpl(EventUtil.java:74)
	at com.sun.javafx.event.EventUtil.fireEvent(EventUtil.java:49)
	at javafx.event.Event.fireEvent(Event.java:198)
	at javafx.scene.control.MenuItem.fire(MenuItem.java:462)
	at com.sun.javafx.scene.control.skin.ContextMenuContent$MenuItemContainer.doSelect(ContextMenuContent.java:1405)
	at com.sun.javafx.scene.control.skin.ContextMenuContent$MenuItemContainer.lambda$createChildren$343(ContextMenuContent.java:1358)
	at com.sun.javafx.event.CompositeEventHandler$NormalEventHandlerRecord.handleBubblingEvent(CompositeEventHandler.java:218)
	at com.sun.javafx.event.CompositeEventHandler.dispatchBubblingEvent(CompositeEventHandler.java:80)
	at com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:238)
	at com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:191)
[...]

PR #638, opened 2016-07-25
Fixes Issue #634

PatrickPenguinTurtle:
https://github.com/WPIRoboticsProjects/GRIP/issues/634

InputSocket.Decorator equals method never checked if the other socket was also a Decorated.

So if a Decorator input socket had its equals method called with itself as a parameter, It would check if this.decorated.equals(this) which is false. Now if a Decorator socket has equals called with another Decorator, it checks if this.decorated.equals(o.decorated)


core/src/main/java/edu/wpi/grip/core/sockets/InputSocket.java
```diff
@@ -114,7 +114,14 @@ public void onValueChanged() {
 
     @Override
     public boolean equals(Object o) {
-      return o instanceof InputSocket && decorated.equals(o);
+      if (o instanceof InputSocket) {
```
AustinShalit: You don't need to check if o instanceof InputSocket

core/src/main/java/edu/wpi/grip/core/sockets/InputSocket.java
```diff
@@ -114,7 +114,14 @@ public void onValueChanged() {
...
+        if (decorated.equals(o)) {
+          return true;
+        } else if (o instanceof Decorator) {
+          return decorated.equals(((Decorator) o).decorated);
+        }
```
AustinShalit: This could be written as a single line return: 
```
return decorated.equals(o) || (o instanceof Decorator && decorated.equals(((Decorator) o).decorated));
```

PatrickPenguinTurtle: Yes but unless GRIP has a style saying "we prefer single line equals methods", multiple lines is easier to read and less prone to accidental errors.

SamCarlberg: I agree with Patrick on this

core/src/main/java/edu/wpi/grip/core/sockets/InputSocket.java
PatrickPenguinTurtle: Note that the previous version already did that.

core/src/main/java/edu/wpi/grip/core/sockets/InputSocket.java
AustinShalit: ok

core/src/main/java/edu/wpi/grip/core/sockets/InputSocket.java
AustinShalit: I don't think the previous version had to do it either.  Did not notice until now.

PatrickPenguinTurtle:
@AustinShalit Care to explain what needs fixup?

AustinShalit:
[This](https://github.com/WPIRoboticsProjects/GRIP/pull/638#discussion-diff-72119463R117) extra if statement.

PatrickPenguinTurtle:
Okay. Will fix when I log back into Mac os. In Windows hellright now.

JLLeitschuh:
LGTM!

PR #639, opened 2016-07-25
Minimize/Maximize options for all steps.

osulacam:
Implements the suggestion in #327


SamCarlberg:
Screenshots / gif / video?

AustinShalit:
Step left/right buttons are broken.  I think the button should be at the top of the step and never move vertically.

osulacam:
screen shot 2016-07-25 at 2 54 44 pm
screen shot 2016-07-25 at 2 54 17 pm

I'll try putting it on top and see if it helps.

ui/src/main/resources/edu/wpi/grip/ui/GRIP.css
```diff
@@ -113,6 +113,12 @@ Button.delete, Button.move-left, Button.move-right, .start-stoppable-button, .ex
     -fx-pref-width: 2em;
 }
 
+.expand {
+	-fx-background-color: transparent;
```
AustinShalit: This line contains a tab.

SamCarlberg:
The `Mask` step shouldn't have the expand/contract arrow because it doesn't actually do anything. I think you should only have it for steps that have user controls (sliders, checkboxes, etc)

osulacam:
I'm pretty happy with this. Let me know if you think anything can be improved.

screen shot 2016-07-25 at 5 01 02 pm
screen shot 2016-07-25 at 5 01 22 pm

AustinShalit:
Sockets are not returned to their original locations right now: https://www.youtube.com/watch?v=3We9orzIyb0&feature=youtu.be

AustinShalit:
This might be beyond the bonds of this pr but an animation would be a cool thing to add to this.

osulacam:
I agree that it would be cool but it's probably better to just get this working and add the animation later in a different pull request.

SamCarlberg:
See: http://stackoverflow.com/questions/13990664/how-to-make-a-vbox-with-animated-close-up

Should be able to call `input.getRoot().setVisible(false)` for each unconnected input, then collapse the pane containing the inputs

SamCarlberg:
The animations are really nice! But I've found [some problems with it](https://www.youtube.com/watch?v=0r6HvRpzatY)

AustinShalit:
LGTM

ui/src/main/java/edu/wpi/grip/ui/pipeline/StepController.java
```diff
@@ -82,9 +100,21 @@ private void initialize() {
         new Image(InputStream.class.cast(icon))));
     buttons.getChildren().add(0, exceptionWitnessResponderButtonFactory.create(step, "Step Error"));
 
+    if (step.getInputSockets().stream()
+        .allMatch(inputSocket -> inputSocket.getSocketHint().getView()
+            .equals(SocketHint.View.NONE))) {
+      expand.setManaged(false);
+    } else {
+      expandIcon.setImage(new Image("/edu/wpi/grip/ui/icons/up.png"));
```
JLLeitschuh: Can you cache the image in the object so it isn't loaded from disk every time you click the button?

ui/src/main/java/edu/wpi/grip/ui/pipeline/StepController.java
```diff
@@ -82,9 +100,21 @@ private void initialize() {
...
+    }
+
     // Add a SocketControlView for each input socket and output socket
     for (InputSocket<?> inputSocket : step.getInputSockets()) {
-      inputSocketMapManager.add(inputSocketControllerFactory.create(inputSocket));
+      InputSocketController tempSocket = inputSocketControllerFactory.create(inputSocket);
+      inputSocketMapManager.add(tempSocket);
+      if (!inputSocket.getSocketHint().getView().equals(SocketHint.View.NONE)) {
```
JLLeitschuh: Why this logic? Isn't it dependent upon being connected or not?

AustinShalit: This is correct.  We only want to collapse User intractable sockets (not: TEXT, SLIDER, RANGE, SELECT, CHECKBOX)

PR #643, opened 2016-07-27
Logo Update

osulacam:
A minor change to the logo to make it significantly more clear. 

I had a few people look over the old logo versus the updated logo and the reviews have been very positive.

[https://drive.google.com/file/d/0B2fTZ6gfFM8zeWVkNnhFcEtaSlk/view?usp=sharing](https://drive.google.com/file/d/0B2fTZ6gfFM8zeWVkNnhFcEtaSlk/view?usp=sharing)


JLLeitschuh:
Is it bad that I didn't even realize that our logo had the word "GRIP" in it until just now....

JLLeitschuh:
Can you update the `README.md` as well?

osulacam:
logo

core/src/test/java/edu/wpi/grip/core/sources/HttpSourceTest.java
```diff
@@ -57,7 +57,7 @@ public void setUp() throws URIException, URISyntaxException {
         contextStore,
         GripServer.IMAGE_UPLOAD_PATH);
 
-    logoFile = new File(Files.class.getResource("/edu/wpi/grip/images/GRIP_Logo.png").toURI());
+    logoFile = new File(Files.class.getResource("/edu/wpi/grip/images/old_GRIP_Logo.png").toURI());
```
JLLeitschuh: This should probably just reference the `util.Files` static variable instead of having the same string twice in the tests. Can you change this?

JLLeitschuh:
This has conflicts with master that need to be resolved.

ThomasJClark:
-1 because I don't think we should change the logo like that.

osulacam:
Everyone who has worked on grip is going to have a bias towards the old logo which is why I got input from people who are not involved.

@ThomasJClark I understand that you created the logo and are hesitant to see it changed, but I hope you can accept that this modification is not taking away from the work that you originally did.

JLLeitschuh:
I'm not feeling hugely attached to this new logo style either. Something seems off with how it looks.
Would it make sense to have the inside of the `g` and the `p` be a different color other than white? Would that look better?

SamCarlberg:
I don't think so, the problem to me is that the holes in the g and p are too small

osulacam:
I'll make and put up a few different versions when I get a chance. Then we can decide which one to go with. The goal of this change is to be as minimal as possible while still allowing people to read the logo.

AustinShalit:
I am going to close this PR because it does not look like this is going to happen in the near future.  Please reopen if you have new information.

PR #645, opened 2016-07-29
Update default roboRIO mDNS address

### LINKED ISSUE #536: Using Grip while on FMS no information publishing to network table
When we are tethered or on our own wireless network we have no issues all the number on the network table show up.
When we are attached to the field the network table shows only nulls '[]' in all the options.
Sometimes when we open the network table it only shows Root and nothing else.
Every time we have tested anything we made sure Grip was processing the image correctly and only had 1 contour found on the last image. 
We have always been able to have image feedback to our dashboard.

Anyone else have this issue? Any information would really help

AustinShalit:
closes #536 

Changes the default roboRIO mDNS address to match capitalization as discussed in #536.


JLLeitschuh:
LGTM

PR #651, opened 2016-08-02
Replaces usage of AutoValue with Kotlin

JLLeitschuh:


SamCarlberg:
I'd prefer to stay with Java as much as possible, and I don't see the advantage of Kotlin over Java in this case

bradamiller:
Why start switching code to Kotlin? Then everyone working on it will have to learn a new language. And it's pretty close to being released again. I'd prefer to see it stay Java and if we decided that we want to switch or start switching, we can do it in an organized way.

JLLeitschuh:
My logic is that I can't for the life of me get auto value to work well with the Intellij built in compiler.
What I realized is that I was trying to get the feature that Kotlin provides with a annotation processed.

PatrickPenguinTurtle:
> My logic is that I can't for the life of me get auto value to work well with the Intellij built in compiler.
What I realized is that I was trying to get the feature that Kotlin provides with a annotation processed.

So what you are saying is you want to change something so it works with a compiler that **isn't** the one GRIP is designed to work with (gradle). That certainly doesn't justify throwing in another language. 

If you add another language then anything involving that language is far less likely to have pull requests for since people don't want to bother learning a new language for this.

JLLeitschuh:
> so it works with a compiler that isn't the one GRIP is designed to work with (gradle)

Gradle isn't a compiler its a build tool that works with Java, Groovy, Scala, Kotlin, C++. Kotlin was designed by jetbrains to interop with java as it generates JVM bytecode.
The kotlin language reduces the amount of boilerplate code that is required to do specific things (this being a perfect example of reducing boilerplate).

> If you add another language then anything involving that language is far less likely to have pull requests for since people don't want to bother learning a new language for this.

You are not required to use Kotlin when writing new code for GRIP. It just gives developers another tool that they can take advantage of if they would like to use it.

bradamiller:
I'm concerned that it will force people to use Intellij even if they were previously using eclipse for doing their builds. I know that's the case with Patrick and sometimes Toby. And it's one module out of the whole code base that would cause people to have to switch..

JLLeitschuh:
There's an eclipse plugin maintained by jetbrains. I've seen it in action and it works fine. They are going to continue to support eclipse because it is widely used.

https://github.com/JetBrains/kotlin-eclipse

SamCarlberg:
Does it have Netbeans support? Can you build and run from he command line? Honestly this is exactly what [Lombok](http://www.projectlombok.org) was designed for

JLLeitschuh:
Netbeans is in the works: https://github.com/Baratynskiy/kotlin-netbeans

I'm not arguing for a rewrite of GRIP in kotlin. My thought is that it is a tool that can make future development easier.

PatrickPenguinTurtle:
> You are not required to use Kotlin when writing new code for GRIP. It just gives developers another tool that they can take advantage of if they would like to use it.

Yes and 6 months down the line (or further) when there is a significant amount of kotlin in GRIP:
1) It becomes impossible to understand without knowing kotlin.
2) It becomes impossible to fix anything without knowing kotlin. 
3) Due to 1) and 2) people stop contributing to GRIP.

Further you missed my point with : 

> My logic is that I can't for the life of me get auto value to work well with the Intellij built in compiler.
What I realized is that I was trying to get the feature that Kotlin provides with a annotation processed.

The reason you want this change (from what you have said) is because you are having an issue with a build system that isn't the build system (compiler was the wrong word) that GRIP is designed for. Why do you even want to compile with something other than what Gradle uses to compile (probably ant but I am not sure).

SamCarlberg:
It's not a build system thing, it's just that IntelliJ doesn't play nice with AutoValue and can't find the generated `AutoValue_Foo` class declarations, so any class that uses AutoValue gets flagged as 'cannot find class `AutoValue_Whatever`'

bradamiller:
I have no problem with Kotlin, and think that it might be a good language to use in the future, but I think we should plan this and get buy-in from all the developers rather than slipping in a module in the middle of all the Java code just to solve a build problem. Thereby forcing the decision on everyone. It means that everyone has to (at least) upgrade their development tools to support Kotlin, and potentially learn it, just to make it easier to solve the problem with the build.

SamCarlberg:
>The reason that I'd rather steer clear of Lombok is that it seems like a tool that was built on top of a language that is need of new features.

Whereas Kotlin is an _entirely new_ language that someone would have to get familiar with to do development on GRIP. It also introduces your original complaint with lombok because it "hides functionality" by creating getters that aren't present in the source code (which lombok does with the `@Getter`, `@Data`, and `@Value` class annotations). 

But this _does not need to use Kotlin or Lombok at all_. There is no reason _not_ to use idiomatic Java in such a simple use case as this.

SamCarlberg:
This PR also doesn't address `@AutoValue` on `ContoursReport`

JLLeitschuh:
From a technological standpoint there is no reason to avoid using kotlin as far as I can see, the language works. The problems that we are having with this are more personal as I can tell from the above comments. This is totally understandable.

> Yes and 6 months down the line (or further) when there is a significant amount of kotlin in GRIP:
 1. It becomes impossible to understand without knowing kotlin.
 2. It becomes impossible to fix anything without knowing kotlin. 
 3. Due to 1) and 2) people stop contributing to GRIP.

I think your concern is valid and makes sense however I counter that Kotlin has a very low barrier to entry if you understand Java. Kotlin was designed to be a language that was very familiar to java developers but with more syntactic sugar so that you can be more expressive while writing fewer lines of code. I would argue that Lombok has the same barrier to entry that Kotlin does.

I'm in no way trying to force this upon anyone, this is just a potential solution.

We are all developers here, we all want to learn something new. Students working on this project at WPI can use this as a new language on their resume.
The barrier to entry argument could be made about any of the technology choices we have made. We are using `JavaFX` for a UI which has it's own set of edge cases that you have to understand, we are using dependency injection with `Guice` because of issues regarding the `EventBus`. By using Kotlin you are able to be more expressive with your code while writing fewer lines of it. I can understand why there might be some fear regarding scaring people off. But I think you can encourage people to give it a shot with the "look, it's pretty simple to understand if you know java. Give it a shot and we can help if you need it" response. I did the same thing with Sam and Austin when they were trying to understand `Guice`.


> This PR also doesn't address @AutoValue on ContoursReport

You're right. I meant to get to that one tonight. I forgot to fix it because for some reason Intellij was fine with it.

PatrickPenguinTurtle:
> It's not a build system thing, it's just that IntelliJ doesn't play nice with AutoValue and can't find the generated AutoValue_Foo class declarations, so any class that uses AutoValue gets flagged as 'cannot find class AutoValue_Whatever'

And this is different from eclipse not playing nice with Guice, Injection, almost all of the @ whatevers,  all of the bytedeco stuff, and some other things that I can't think of now? 

We shouldn't base our design decisions on what modules play well with our IDEs when they are not responsible for building our code. If there is a reason that Kotlin is better for our means than what we currently use and is better by enough to be worth the extra hassle, then by all means lets us it. But so far the only reason I have heard is a note about being more sweet (I'm not sorry for that pun) and that it plays well with IntelliJ while AutoValue doesn't. 

Also does it have RoboRio support? I get that it is a java extension (maybe not precise technical term) but that doesn't necessarily mean it runs on any platform java runs on without issues.

JLLeitschuh:
> Also does it have RoboRio support? I get that it is a java extension (maybe not precise technical term) but that doesn't necessarily mean it runs on any platform java runs on without issues.

Yes, it's a JVM based language that was designed to work with android (it compiles to JVM 6 bytecode).

> And this is different from eclipse not playing nice with Guice, Injection, almost all of the @ whatevers, all of the bytedeco stuff, and some other things that I can't think of now?

I had no idea you were having problems. People have managed to get eclipse to work with GRIP if you need help with something let us know.

> We shouldn't base our design decisions on what modules play well with our IDEs when they are not responsible for building our code. If there is a reason that Kotlin is better for our means than what we currently use and is better by enough to be worth the extra hassle, then by all means lets us it. But so far the only reason I have heard is a note about being more sweet (I'm not sorry for that pun) and that it plays well with IntelliJ while AutoValue doesn't.

I agree. What features would you like to know about? There are plenty of really cool features of the language that could make your life easier while working on code generation. For example, [string interpolation](https://kotlinlang.org/docs/reference/idioms.html#string-interpolation), far superior [null safety](https://kotlinlang.org/docs/reference/null-safety.html) than java, documentation using [markdown format instead of javadoc](https://kotlinlang.org/docs/reference/kotlin-doc.html#kdoc-syntax), [immutable data classes](https://kotlinlang.org/docs/reference/data-classes.html), [type inference](https://kotlinlang.org/docs/reference/basic-syntax.html#defining-local-variables) (meaning you don't have to keep writing the type on the left side of a variable assignment), ect... Take a look and ask about any features you don't understand. I'd be more than happy to explore this with you.

AustinShalit:
I am going to close this PR because it does not look like this is going to happen in the near future. Please reopen if you have new information.

PR #661, opened 2016-08-09
A null file will cancel application exit

AustinShalit:


JLLeitschuh:
LGTM! I don't know if this is something we could have had a test for.

AustinShalit:
We can not test for this because it is OS specific.

PR #676, opened 2016-10-05
GRIP as a web service

JLLeitschuh:
In order to compile this you will need to [clone this repository](https://github.com/gwizard/gwizard) and run `mvn install` to install the newest snapshot of the repo in your local m2 repository.

To test:

Make a `put` rest call against `http://localhost:8080/web/sources`
Copy the UUID of the image output socket.
In a browser tab open the address `http://localhost:8080/web/sockets/{UUID from last request}/images`


core/src/main/java/edu/wpi/grip/core/Step.java
```diff
@@ -82,7 +82,7 @@ public OperationDescription getOperationDescription() {
   }
 
   /**
-   * @return a UUID to distinguish this class
+   * @return A UUID to distinguish this class.
```
SamCarlberg: First words in Javadocs for `@` lines (`@param`, `@return`, etc.) should be lowercase

AustinShalit:

PR #678, opened 2016-10-06
Add cascade classifier operation

SamCarlberg:
Also adds a new source for cascade classifiers

screenshot from 2016-10-10 15-14-28
### Goals
- [x] Code generation
- [x] File source for choosing XML configuration files
- [ ] Add operation to extract ROIs from the image (can be a separate PR)
- [x] Add additional inputs for finer control over the classifier


ui/src/main/java/edu/wpi/grip/ui/preview/RectangleSocketPreviewView.java
```diff
@@ -0,0 +1,110 @@
...
+      }
+      final Mat convertInput = input;
+      final int numRegions = rectangles.size();
+      platform.runAsSoonAsPossible(() -> {
+        final Image image = this.imageConverter.convert(convertInput);
+        this.imageView.setImage(image);
+        this.infoLabel.setText("Found " + numRegions + " regions of interest");
+      });
+    }
+  }
```
JLLeitschuh: This code seems eerily familiar, could this be refactored into the core (maybe behind an interface) with just a `render` method that takes a boolean `showInputImage` and also has a method like `getDescriptionString`?

SamCarlberg: I copy-pasted from the `LinesReport` preview, that's probably what you're thinking of. I'll pretty it up with a refactor later

JLLeitschuh: Cool, if you move the `render` into core then it becomes incredibly easy for me to render as a web server too. Because then the code for creating the preview image will be in the core.

SamCarlberg: Good idea

core/src/main/java/edu/wpi/grip/core/sources/FileSource.java
```diff
@@ -0,0 +1,87 @@
...
+
+import java.util.List;
+import java.util.Properties;
+
+/**
+ * A source for a single file's path. This will only supply the <i>path</i> to the file; it is up
+ * to the operations to resolve the file associated with the path.
+ */
+@XStreamAlias("grip:File")
+public class FileSource extends Source {
```
JLLeitschuh: I would disagree with this pattern. I would make this a classifiers source.
Just like we don't expect steps to resolve paths for images that are loaded from the filesystem this too should behave the same.

core/src/main/java/edu/wpi/grip/core/util/MatUtils.java
```diff
@@ -1,46 +0,0 @@
...
-   *
-   * @param image           the image
-   * @param showOriginal    true if the original image should be shown
-   * @param dataSupplier    supplier for the data being drawn on the image
-   * @param drawingFunction the function to draw on the image with
-   * @param <T>             the type of the data being drawn
-   *
-   * @return a mat with the data drawn on it
-   */
-  public static <T> Mat draw(Mat image,
```
JLLeitschuh: I did like this solution, I just would have passed the `tmp` mat into it.

SamCarlberg:
It'd be nice to have a `Size` input socket to avoid having one or two `New Size` steps for each cascade classifier. @JLLeitschuh thoughts?

core/src/main/java/edu/wpi/grip/core/operations/composite/CascadeClassifierOperation.java
```diff
@@ -0,0 +1,132 @@
...
+  private final SocketHint<Size> minSizeHint =
+      new SocketHint.Builder<>(Size.class)
+          .identifier("Min size")
+          .initialValue(new Size(0, 0))
+          .build();
+  private final SocketHint<Size> maxSizeHint =
+      new SocketHint.Builder<>(Size.class)
+          .identifier("Max size")
+          .initialValue(new Size(0, 0))
+          .build();
```
JLLeitschuh: Want to put these in the socket hints builder?

core/src/main/java/edu/wpi/grip/core/operations/composite/CascadeClassifierOperation.java
```diff
@@ -0,0 +1,132 @@
...
+    final String fileName = filePath.getValue().get();
+    if (!fileName.equals(lastFile)) {
+      // Don't load the same file multiple times in a row
+      try {
+        classifier.load(fileName);
+        lastFile = fileName;
+      } catch (RuntimeException e) {
+        // Error with the config file, reset the classifier and throw an exception
+        classifier.load(lastFile);
+        throw new IllegalArgumentException("Invalid XML in configuration file", e);
```
JLLeitschuh: The classifier should not be loaded from within the operation. The source should be the object instantiating the `CascadeClassifier` object

JLLeitschuh: Think about this, if you ever want to add a HTTP source that creates a classifier then this won't work. Similarly, if someone sends one via network tables this won't work either.

SamCarlberg: That's a good idea, though I think that would have a negative impact on code generation because it would require users to manually instantiate the classifiers instead of having the pipeline handle everything

core/src/main/java/edu/wpi/grip/core/operations/composite/CascadeClassifierOperation.java
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * Operation for identifying parts of an image with a cascade classifier.
+ */
+public class CascadeClassifierOperation implements Operation {
+
+  public static final OperationDescription DESCRIPTION =
+      OperationDescription.builder()
+          .name("Cascade Classifier")
+          .summary("Runs a cascade classifier on an image")
+          .icon(Icon.iconStream("opencv"))
```
SamCarlberg: It's basically a one-to-one mapping of an OpenCV function, so it's okay as is

core/src/main/java/edu/wpi/grip/core/operations/composite/CascadeClassifierOperation.java
SamCarlberg: No need, the default already initializes to (0, 0)

core/src/main/java/edu/wpi/grip/core/operations/composite/CascadeClassifierOperation.java
JLLeitschuh: Cooooode gen!

core/src/main/java/edu/wpi/grip/core/operations/composite/CascadeClassifierOperation.java
SamCarlberg: > It's basically a one-to-one mapping of an OpenCV function

All the other OpenCV operations use the OpenCV icon. So this will use the OpenCV icon, too.

JLLeitschuh:
Is this ready to be merged?

SamCarlberg:
Yes

JLLeitschuh:
LGTM!

PR #680, opened 2016-10-07
Fix massive memory leak in previews (#679)

### LINKED ISSUE #679: Previewing a lines output makes memory use balloon
This is caused by tens of thousands of `Pointer$NativeDeallocator` being created every second (~30k-40k) because two new points are being created for every line. Each point has a deallocator object, so this results in a ton of useless objects.

### Resolution

Reuse a single `Point` object and modify it's coordinates before drawing the line endpoints

Note: this will need to be fixed for the next release

SamCarlberg:
closes #679 

Fixes the bug in #679


JLLeitschuh:
LGTM!

PR #682, opened 2016-10-11
Replace List<Number> with Range

### LINKED ISSUE #677: Don't use List to hold a min-max range for inputs
We should either use apache's `Pair<Double, Double>` or we could write our own range class that uses primitive doubles e.g.

```java
public class Range {
  private double min;
  private double max;

  // constructors, getters, setters...

}
```

So we'd have `InputSocket<Pair<Double, Double>>` or `InputSocket<Range>` instead of `InputSocket<List<Double>>`

SamCarlberg:
Closes #677


AustinShalit:
Why not use guava? https://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/Range.html

SamCarlberg:
Because it's immutable and I don't want to be generating excessive amounts of short-lived objects

core/src/main/java/edu/wpi/grip/core/Range.java
```diff
@@ -0,0 +1,94 @@
...
+  /**
+   * Creates a new range with the given bounds.
+   *
+   * @param min the lower end of the range
+   * @param max the upper end of the range
+   *
+   * @throws IllegalArgumentException if min > max
+   */
+  public Range(double min, double max) {
+    checkArgument(min <= max, "Min must be <= max");
```
nightpool: Would probably be more usable if, instead of throwing checkArgument exceptions, this just did the intelligent thing and set min and max to the smallest and the largest, respectively.

JLLeitschuh: I'm not sure if I would agree with that.

JLLeitschuh:
I agree with @AustinShalit on this one. I think using Guava makes sense.
These short lived objects will be cleaned up by the JVM and probably will stay in eden.

ZeusChops:
Sorry, this looks like a dead thread but I found some time to 'walk around GRIP' on Github. I am not so sure about Guava. Looking into it, I think Guava uses Comparable (.compareTo()). Not to say that anything is wrong with this, but if anything is mistyped, this custom Range class will force compilation issues (meaning that it will be easily revisable and noticed faster than letting Travis CI check it). Also, .compareTo() can be added to anything, and has been built in to Strings, Arrays, int, double, char, and plenty of other Objects in the Java library. CompareTo also has a tendency to be one of two comparisons, based on how the Java library is written. Where the object either checks bit values or else the object compares memory addresses which can make code confusing, and/or mismatched.

PR #691, opened 2016-10-24
Fix a potential deadlock on opening a project

SamCarlberg:
This deadlock is caused by `SocketImpl.setValueOptional` being a synchronized method and posting a SocketChangedEvent that the UI thread tries to handle by locking on the socket when getting it's value.


JLLeitschuh:
LGTM!

PR #692, opened 2016-10-26
Add versioning

SamCarlberg:
If a save file cannot be loaded, no changes happen to the pipeline and the user is notified. This will happen for any save file, pre-versioning or otherwise.

When saving a loaded project, the original version string will be changed to the version of GRIP that saved it. Pre-versioning save files will also be upgraded to the versioned format and have the current version of GRIP

This depends on #693 for user alerts


JLLeitschuh:
Does this really have to use inheritance to solve this problem?
I'm not really a fan of putting a `AbstractPipelineEntry` in the inheritance hierarchy  of `Step`.
Let's chat about this.

ui/src/main/java/edu/wpi/grip/ui/MainWindowController.java
```diff
@@ -60,6 +64,7 @@
 /**
  * The Controller for the application window.
  */
+@SuppressWarnings("PMD.GodClass") // temporary
```
JLLeitschuh: Add an issue to track this or make sure that this isn't in the final version you want to merge.

SamCarlberg: There's already an issue open for this

core/src/main/java/edu/wpi/grip/core/Source.java
```diff
@@ -101,6 +111,28 @@ public final void initializeSafely() {
...
+
+  /**
+   * Cleans up this source when it gets removed from the pipeline. The default implementation is a
+   * NOP, but subclasses should override this method if they need to free any resources or
+   * de-register callbacks.
+   */
+  @Override
+  @SuppressWarnings("PMD")
+  protected void cleanUp() {
+    // Default to NOP
```
JLLeitschuh: Why not make this a default method?

core/src/main/java/edu/wpi/grip/core/http/HttpPipelineSwitcher.java
```diff
@@ -38,8 +39,14 @@ protected void handleIfPassed(String target,
...
-    project.open(new String(IOUtils.toByteArray(request.getInputStream()), "UTF-8"));
-    response.setStatus(HttpServletResponse.SC_CREATED);
-    baseRequest.setHandled(true);
+    try {
+      project.open(new String(IOUtils.toByteArray(request.getInputStream()), "UTF-8"));
+      response.setStatus(HttpServletResponse.SC_CREATED);
+      baseRequest.setHandled(true);
+    } catch (InvalidSaveException e) {
+      // 406 - Not Acceptable if given an invalid save
+      response.setStatus(HttpServletResponse.SC_NOT_ACCEPTABLE);
```
JLLeitschuh: Maybe attach a header that contains the message from the exception?

core/src/main/java/edu/wpi/grip/core/AbstractPipelineEntry.java
```diff
@@ -0,0 +1,151 @@
...
+  @Override
+  public final void setId(String id) throws NullPointerException, IllegalArgumentException {
+    checkNotNull(id, "The ID cannot be null");
+    boolean inDeserialization = Arrays.stream(new Exception().getStackTrace())
+        .map(e -> e.getClassName())
+        .anyMatch(n -> n.matches(".*(Step|Source)Converter"));
+    if (!inDeserialization) {
+      throw new IllegalStateException(
+          "This method may only be called during project deserialization");
+    }
```
JLLeitschuh: Why not just make this package private? Or hide it in some other way?

core/src/main/java/edu/wpi/grip/core/AbstractPipelineEntry.java
```diff
@@ -0,0 +1,151 @@
...
+  @Override
+  public String toString() {
+    return getId();
+  }
+
+  /**
+   * Pool of used IDs.
+   */
+  @VisibleForTesting
+  static class IdPool extends HashMap<Class<? extends PipelineEntry>, Set<String>> {
```
JLLeitschuh: Guava has a type for this:
https://github.com/google/guava/wiki/NewCollectionTypesExplained#implementations

core/src/main/java/edu/wpi/grip/core/Palette.java
```diff
@@ -24,10 +29,54 @@
...
+      );
+      checkDuplicates(
+          operationData,
+          "output socket names",
+          operation.getOutputSockets(), s -> s.getSocketHint().getIdentifier()
+      );
+      checkDuplicates(operationData, "socket IDs", sockets, Socket::getUid);
+    } finally {
+      operation.cleanUp();
+    }
```
JLLeitschuh: Not a huge fan of creating an operation when the user hasn't asked the UUID to create one.

Perhaps put this into a test that loads every operation?

core/src/main/java/edu/wpi/grip/core/PipelineEntry.java
```diff
@@ -0,0 +1,61 @@
...
+  List<OutputSocket> getOutputSockets();
+
+  /**
+   * Gets the socket with the given ID in this entry.
+   *
+   * @param uid the UID of the socket to get
+   *
+   * @throws NoSuchElementException if there is no socket with the given UID,
+   */
+  Socket getSocketByUid(String uid) throws NoSuchElementException;
```
JLLeitschuh: ~~Perhaps using the UUID type would be nice.~~

Never mind. I understand now.
Is there some sort of launch time guarantee of being unique?

core/src/main/java/edu/wpi/grip/core/Step.java
```diff
@@ -24,7 +24,7 @@
  * sockets, and it runs the operation whenever one of the input sockets changes.
  */
 @XStreamAlias(value = "grip:Step")
-public class Step {
+public class Step extends AbstractPipelineEntry {
```
JLLeitschuh: Why not use composition instead of inheritance?

core/src/main/java/edu/wpi/grip/core/sockets/Socket.java
```diff
@@ -26,6 +26,17 @@
   SocketHint<T> getSocketHint();
 
   /**
+   * Gets a String that uniquely identifies this socket. This only needs to be unique for the set
+   * of sockets containing this one (e.g. only per-operation or per-source); it does not need to be
+   * <i>universally unique</i>. However, <strong>this is not allowed to change</strong>; even if
+   * the name, view, or even the type changes, the UID has to be constant in order for projects
+   * saved from different versions of GRIP to be compatible.
+   *
+   * @implSpec <strong>This value MAY NOT change</strong>
```
JLLeitschuh: `This value MUST NOT EVER change`

JLLeitschuh: Why is this on the socket and not the socket hint?

core/src/main/java/edu/wpi/grip/core/sources/HttpSource.java
```diff
@@ -87,10 +85,10 @@
       GripServer server,
       ContextStore store,
       @Assisted String path) {
-    super(exceptionWitnessFactory);
+    super(makeId(HttpSource.class), exceptionWitnessFactory);
```
JLLeitschuh: Why do you need this duplicate code in every source?
Why not just have the super constructor pull this information?

core/src/test/java/edu/wpi/grip/core/serialization/ProjectTest.java
```diff
@@ -131,6 +132,39 @@ public void testSerializeEmptyPipeline() throws Exception {
...
+    String xml = "<grip:Project>\n"
+        + "  <grip:Pipeline>\n"
+        + "    <sources/>\n"
+        + "    <steps>\n"
+        + "      <grip:Step name=\"\"/>\n"
+        + "    </steps>\n"
+        + "    <connections/>\n"
+        + "    <settings/>\n"
+        + "  </grip:Pipeline>"
+        + "</grip:Project>";
```
JLLeitschuh: This makes me wish for kotlin multiline strings:
https://kotlinlang.org/docs/reference/basic-types.html#string-literals

JLLeitschuh (commented):
Just adding some comments. I don't know what state this is in but I figured I'd provide some feedback for while you are working on it.

core/src/main/java/edu/wpi/grip/core/Step.java
JLLeitschuh: https://en.wikipedia.org/wiki/Composition_over_inheritance

core/src/main/java/edu/wpi/grip/core/PipelineEntry.java
SamCarlberg: Part of the contract is that the socket ID has to be unique within its parent

JLLeitschuh: Can you add a unit test that guarantees this contract for all operations?

PR #693, opened 2016-10-26
Add mechanism for alerting the user when they try to do something wrong

SamCarlberg:
e.g. trying to generate code with operations that don't support it

Post a WarningEvent to the eventbus and an alert dialog will appear.
### Still to do
- [ ] Tests!
- [x] Make alerts resize to fit content


ui/src/main/resources/edu/wpi/grip/ui/warning_alert.css
```diff
@@ -0,0 +1,3 @@
+html {
+    font: 14px Roboto;
```
AustinShalit: Is this HDPI display compatible?

ui/src/main/java/edu/wpi/grip/ui/WarningAlert.java
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * Markdown processor. Converts markdown to HTML.
+   */
+  private static final PegDownProcessor mdProcessor = new PegDownProcessor();
+  private final WebView webView = new WebView();
+  private final WebEngine webEngine = webView.getEngine();
+
+  // Width and height of the alert dialog
+  private static final double WIDTH = 360;
+  private static final double HEIGHT = 180;
```
AustinShalit: Are these HDPI display compatible?

ui/src/main/java/edu/wpi/grip/ui/WarningAlert.java
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  /**
+   * Initializes the web view and engine prior to setting its content.
+   */
+  private void initWebView() {
+    webView.setBlendMode(BlendMode.DARKEN); // make the background transparent
+    webEngine.setUserStyleSheetLocation(
+        getClass().getResource("warning_alert.css").toExternalForm()
+    );
```
AustinShalit: This style feels weird.  Idk what the guide says but I feel like this should be one line or broken differently.

ui/src/main/java/edu/wpi/grip/ui/WarningAlert.java
```diff
@@ -0,0 +1,69 @@
...
+import javafx.scene.web.WebEngine;
+import javafx.scene.web.WebView;
+import javafx.stage.StageStyle;
+import javafx.stage.Window;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * An alert that warns the user when they try to do something unsupported, such as trying to
+ * generate code with an operation that doesn't support code gen.
```
AustinShalit: `doesn't` -> `does not`

ui/src/main/java/edu/wpi/grip/ui/WarningAlert.java
SamCarlberg: I doubt it

ui/src/main/java/edu/wpi/grip/ui/WarningAlert.java
SamCarlberg: It's either this or embed the CSS inline with the HTML

ui/src/main/java/edu/wpi/grip/ui/WarningAlert.java
SamCarlberg: This will be removed once I get the dialogues to properly fit the rendered content

ui/src/main/java/edu/wpi/grip/ui/WarningAlert.java
AustinShalit: I was talking more about the java code rather than the HTML.  I think line 65 should be moved to the end of line 64.

SamCarlberg: Oh. This is what I normally do, it makes it easier to see what each method call is used for

ui/src/main/resources/edu/wpi/grip/ui/warning_alert.css
JLLeitschuh: You need to use `em`
https://css-tricks.com/why-ems/

JLLeitschuh: We use `em` in a lot of places.

SamCarlberg:
I got rid of the webview + markdown stuff because it was impossible to consistently get the height of the rendered HTML. Sometimes the height would be an appropriate value and look fine, and other times it would want to be 100 pixels taller for exactly the same content and completely screw up the pane.

SamCarlberg:
@JLLeitschuh @AustinShalit Can one of you test this? I'd like to get this in ASAP

AustinShalit:
Testing steps?

PR #696, opened 2016-10-31
Fix #695

### LINKED ISSUE #695: Ubuntu 16.10: Cancel on error quits GRIP UI
NullPointerException when removing objects from pipeline. Two case scenarios:

Case 1:
   -Setup pipeline, connect some steps
   -Start deleting steps from pipeline that are used by other steps (slowly)
   -Receive an error about a NullPointerException from the pipeline
   -Hit cancel on the message, then the UI quits out unexpectedly

Case 2:
   -Start up Ubuntu without a camera connected (or unplug all cameras from ubuntu before launching GRIP)
   -Add camera to GRIP without physical connection
   -Verify that the Camera is 'connected' using the flashing red 'X' next to the step in the sources pane
   -Remove camera from the pipeline, UI crashes (there might be another UI error window that displays, but I have not yet confirmed that)

Device: mid-2015 MacBook Pro
Type: Virtual Machine, guest
OS: Ubuntu 16.10 x64

VirtualBox configuration:
RAM: 8GB
CPU: i7 2.2GHz (quad-core, not default)
PAX/NE enabled (not default)
VRAM: 128MB (not default)
1 Screen (default)
HiDPI (2.0)
Network: NAT (default)
Audio: ICH AC97 (not default)
Execution Cap: 100% (default)
[VirtualBox Hard Drive download](https://drive.google.com/open?id=0B86Py5WkgumYVjUxSTM0QXlNTGc)

SamCarlberg:
Closes #695 

695 brought up two problems:

1. Removing a step with connections from the pipeline would cause an exception to be thrown; and  
2. Removing a failing camera source would also cause an exception (unrelated to point 1)

### DIFF
--- a/core/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
+++ b/core/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
@@ -285,12 +285,24 @@ public class CameraSource extends Source implements RestartableService {
 
   @Override
   public void stopAndAwait() {
-    stopAsync().cameraService.stopAndAwait();
+    try {
+      stopAsync().cameraService.stopAndAwait();
+    } catch (IllegalStateException e) {
+      if (!cameraService.state().equals(State.FAILED)) {
+        throw e;
+      }
+    }
   }
 
   @Override
   public void stopAndAwait(long timeout, TimeUnit unit) throws TimeoutException {
-    stopAsync().cameraService.stopAndAwait(timeout, unit);
+    try {
+      stopAsync().cameraService.stopAndAwait(timeout, unit);
+    } catch (IllegalStateException e) {
+      if (!cameraService.state().equals(State.FAILED)) {
+        throw e;
+      }
+    }
   }
 
   @Override
--- a/core/src/test/java/edu/wpi/grip/core/sources/CameraSourceTest.java
+++ b/core/src/test/java/edu/wpi/grip/core/sources/CameraSourceTest.java
@@ -42,7 +42,6 @@ import java.util.concurrent.TimeoutException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-import static com.google.common.truth.Truth.assertThat;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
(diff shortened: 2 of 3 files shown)


SamCarlberg:
@JLLeitschuh Can you take a look a the CameraSourceTest to make sure I didn't screw it up?

core/src/test/java/edu/wpi/grip/core/sources/CameraSourceTest.java
```diff
@@ -193,18 +200,24 @@ public FrameGrabber create(String addressProperty) throws MalformedURLException
...
     waiter3.await();
-    try {
-      source.stopAndAwait();
-      fail("This should have failed");
-    } catch (IllegalStateException e) {
-      assertThat(e.getCause()).isNotNull();
-      assertThat(e.getCause().getCause()).isNotNull();
-      assertThat(e.getCause().getCause()).hasMessage(GRABBER_START_MESSAGE);
-    }
+    source.stopAndAwait();
```
JLLeitschuh: Stop and await should fail because starting throws an exception. Right?

SamCarlberg: Uh, that's what this PR is fixing.

core/src/main/java/edu/wpi/grip/core/sources/CameraSource.java
```diff
@@ -285,12 +285,24 @@ public CameraSource stopAsync() {
...
 
   @Override
   public void stopAndAwait(long timeout, TimeUnit unit) throws TimeoutException {
-    stopAsync().cameraService.stopAndAwait(timeout, unit);
+    try {
+      stopAsync().cameraService.stopAndAwait(timeout, unit);
+    } catch (IllegalStateException e) {
+      if (!cameraService.state().equals(State.FAILED)) {
+        throw e;
+      }
```
SamCarlberg: It means that the service crashed while running, or an exception was thrown on startup (e.g. the webcam doesn't exist or isn't available). All this does is stop if it's in terminal state (TERMINATED or FAILED), instead of throwing another exception if startup failed

PR #698, opened 2016-11-01
Add source for video files

SamCarlberg:
Currently supports:

- [x] mp4
- [x] mpeg
- [x] avi


core/src/main/java/edu/wpi/grip/core/sources/VideoFileSource.java
```diff
@@ -0,0 +1,198 @@
...
+          }
+          synchronized (workingMat) {
+            m.copyTo(workingMat);
+          }
+          m.release();
+          isNewFrame.set(true);
+          eventBus.post(new SourceHasPendingUpdateEvent(VideoFileSource.this));
+          Thread.sleep((long) (1e3 / fps));
+        } catch (InterruptedException | FrameGrabber.Exception e) {
+          getExceptionWitness().flagException(e);
```
JLLeitschuh: Handle the `InterruptedException` like an `InterruptedException` should be handled.
http://www.yegor256.com/2015/10/20/interrupted-exception.html

red-green:
I don't know if/when this will get merged but I'm very interested in using video files for testing pipelines instead of static images.

core/src/main/java/edu/wpi/grip/core/sources/VideoFileSource.java
```diff
@@ -0,0 +1,194 @@
...
+          }
+          m.release();
+          isNewFrame.set(true);
+          eventBus.post(new SourceHasPendingUpdateEvent(VideoFileSource.this));
+          Thread.sleep((long) (1e3 / fps));
+        } catch (FrameGrabber.Exception e) {
+          getExceptionWitness().flagException(e);
+        } catch (InterruptedException e) {
+          Thread.currentThread().interrupt();
+          getExceptionWitness().flagException(e, "Interrupted while waiting for next frame");
```
JLLeitschuh: You probably don't want to flag and you probably want to exit the while loop.

core/src/main/java/edu/wpi/grip/core/sources/VideoFileSource.java
```diff
@@ -0,0 +1,194 @@
...
+import java.io.IOException;
+import java.util.List;
+import java.util.Properties;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * A source for a video file input.
+ */
+@XStreamAlias("grip:VideoFile")
+public class VideoFileSource extends Source {
```
JLLeitschuh: There seems to be duplicate code between this and the Camera source. Can any of that be refactored into a base class?

Also, I notice you are using a thread. We had issues with threads and their weirdness. Any reason you didn't implement this as a restarting service?

SamCarlberg: There are some fundamental differences between them. Camera frame grabbers don't get null frame mats from the conversion; video files do. The grabber for the video file needs to update at a specific rate specified by the file itself; camera sources just grab as fast as possible.

And I'm using a Thread (scheduled executor now) to grab the frames because of these two issues.

JLLeitschuh:
Can you rewind or perhaps restart the video?
What about advancing one frame at a time? Would either of these features be useful?

nightpool:
@jlleitschuh scope creep

SamCarlberg:
That _is_ something that would be worthwhile to include. It would act similar to the multi-image file source, but with a scrubber and play/pause button as well.

SamCarlberg:
There's something weird with the FFMpegFrameGrabber where it'll segfault, seemingly at random. Seems to only happen when scrubbing through a long video?

`.mpeg` files don't seem to work very well, either. Scrubbing is laggy and I keep getting errors in the console like
```
[mpegts @ 0x7fce787cda00] Invalid timestamps stream=0, pts=333257999, dts=333261000, size=221
```

I only have a ~7MB mpeg file though, maybe a smaller one would work better.

JLLeitschuh:
Seems like you are re-inventing the wheel here:
https://github.com/ReactiveX/RxJava

This library also provides synchronization guarantees.

SamCarlberg:
I wanted to keep it similar to JavaFX's Observables

JLLeitschuh:
Do you want to provide any synchronization guarantees about adding/removing observers?

SamCarlberg:
There are in `SynchronizedObservable`. Adding or removing an observer will block until all the observers are notified when the value changes.

JLLeitschuh:
Are you sure you don't want to just use RX java's observable?

PR #699, opened 2016-11-02
Reduce runtime memory use

SamCarlberg:
Reduces memory use by watershed (release temp objects and reuse memory)

I can run in headless mode with -Xmx15m and UI mode with -Xmx150m (was previously seeing up to 890MB heap in UI with watershed)


SamCarlberg:
So just reducing the max heap size will cause more garbage collections on the young generation, but doesn't seem to run full GC's any more often.

The biggest problem with memory is that if the GC leaves dead native objects (Mats, Rects, etc.) floating around, the native memory won't get freed up and causes the process's memory use to explode -- 290MB of used heap space can correspond to __10GB__ of memory used by the process (this could go even higher, but my 16GB of RAM was used up and swapping was getting nasty). This was just running the core jar.

Calling `System.gc()` after every pipeline run _hugely_ alleviates this problem. With the same pipeline and settings as above, I'm seeing heap use between 5MB and 7MB and a total process use of ~200MB. The downside to this is that a full GC runs about _20 times per second_. 

Calling `System.gc` periodically (I tested with 15 second intervals) helped, but heap use was peaking around 50MB and the process was using ~1.6GB of memory. Not great.

Calling `System.gc()` every 10 pipeline runs, however, seems to be a good solution. Heap use is between 6MB and 8MB, process seems to be steady at about 230-240MB, and a full GC about 2-3 times per second.

---

### Heap and CPU profiles

#### `-Xmx30m`, no GC calls

screenshot from 2016-11-04 12-19-03

#### `-Xmx2G`, no GC calls

screenshot from 2016-11-04 12-19-07

#### `-Xmx2G`, `System.gc()` after every pipeline run

screenshot from 2016-11-04 12-19-10

#### `-Xmx2G`, `System.gc()` every 15 seconds

screenshot from 2016-11-04 12-19-12


#### `-Xmx2G`, `System.gc()` after every 10 pipeline runs

screenshot from 2016-11-04 12-19-15

JLLeitschuh:
I'd like to get some other people's thoughts on this before we merge it.

@saudet Is calling `release` on these objects okay or should we be calling something else?

SamCarlberg:
`deallocate()` only seems to work if the `Deallocator` object hasn't been set (which it seems to be) or on garbage collection. But it looks like calling `release()` on a mat will immediately free up the native memory if there's no longer a reference to it

saudet:
deallocate() should also free memory immediately, it doesn't?

To free the Pointer object itself from the Java heap, we need to wait after
the GC though...

SamCarlberg:
So it looks like the only way to address this is to periodically call `System.gc()`, or to make Eden space small to force more GC runs to clean up unused mat objects and deallocate their native memory.

The first one has the advantage of not putting any limits on the heap size, but it will trigger a major stop-the-world GC, which can take a while. It also keeps memory use very low in both UI and headless modes.

The second option will only trigger minor GCs in Eden and S1/S2 that don't take much time, but that will impose limits on other parts of the program. For example, previewing large images in the UI takes a _ton_ of memory on the heap -- a 12MP picture can take upwards of 150MB to preview -- and won't be able to fit in new gen, making the app hang while the JVM keeps trying to unsuccessfully allocate enough memory for it. #711 addresses this problem, but there may be other parts of the program or future features that would be limited by a tiny new gen space.

I'm not happy with either of these options, but I don't see any other alternatives.

saudet:
For cases like these, it's useful to reuse (pre)allocated memory as much as possible, to avoid constantly allocating and deallocating. I am not familiar with the details of this project, but that doesn't look possible?

SamCarlberg:
I don't think it's possible for us to use preallocated memory. We also can't make eden small because that puts limits on the size of the images we can preview in the UI (big images = big buffers that need to be allocated). So the only solution I see is to periodically call `System.gc()` to clean up the old Mats leftover from runs of the pipeline.

SamCarlberg:
@JLLeitschuh I'd like you to try this out and see if you can find any serious problems

saudet:
@SamCarlberg BTW, recent versions of JavaCPP check before each allocation the amount of physical memory used and call `System.gc()` only when that amount is crossed, so it's possible to limit these calls to a minimum that way. We can adjust that amount by setting the `org.bytedeco.javacpp.maxphysicalbytes` system property to the desired value:
http://bytedeco.org/javacpp/apidocs/org/bytedeco/javacpp/Pointer.html#maxPhysicalBytes

SamCarlberg:
That's good to know. I'll try it out, see how it compares to running `System.gc()` every few passes.

core/src/main/java/edu/wpi/grip/core/operations/composite/WatershedOperation.java
```diff
@@ -64,12 +65,22 @@
   private final InputSocket<ContoursReport> contoursSocket;
   private final OutputSocket<ContoursReport> outputSocket;
 
+  private static final int maxMarkers = 253;
```
JLLeitschuh: Shouldn't this be `MAX_MARKERS`?

core/src/main/java/edu/wpi/grip/core/operations/composite/WatershedOperation.java
```diff
@@ -64,12 +65,22 @@
...
+  private final List<Mat> markerPool = new ArrayList<>(maxMarkers);
```
JLLeitschuh: Can this be an immutable list?

core/src/main/java/edu/wpi/grip/core/Cleaner.java
```diff
@@ -0,0 +1,68 @@
...
+    runsSinceLastGc++;
+    if (lastRun == 0) {
+      lastRun = System.nanoTime();
+    }
+    if (runsSinceLastGc >= MAX_RUNS_BEFORE_GC
+        && System.nanoTime() - lastRun >= MIN_DELAY * 1e6) {
+      runsSinceLastGc = 0;
+      lastRun = System.nanoTime();
+      System.gc();
+    }
```
JLLeitschuh: Use StopWatch?

build.gradle
```diff
@@ -435,6 +441,12 @@ project(":ui") {
         scopes.PROVIDED.plus += [configurations.ideProvider]
     }
 
+    if (project.hasProperty('jvmArgs')) {
+        run {
+            jvmArgs = (project.jvmArgs.split("\\s+") as List)
+        }
+    }
```
JLLeitschuh: Move this to a common config block

core/src/main/java/edu/wpi/grip/core/operations/composite/WatershedOperation.java
```diff
@@ -64,12 +65,24 @@
...
   public WatershedOperation(InputSocket.Factory inputSocketFactory,
                             OutputSocket.Factory outputSocketFactory) {
     srcSocket = inputSocketFactory.create(srcHint);
     contoursSocket = inputSocketFactory.create(contoursHint);
     outputSocket = outputSocketFactory.create(outputHint);
+    List<Mat> tmpPool = new ArrayList<>();
+    for (int i = 0; i < MAX_MARKERS; i++) {
+      tmpPool.add(new Mat());
+    }
+    markerPool = ImmutableList.copyOf(tmpPool);
```
JLLeitschuh: Use the `ImmutableListBuilder`

SamCarlberg: That's no better than what's currently happening.

build.gradle
SamCarlberg: Done

core/src/main/java/edu/wpi/grip/core/Cleaner.java
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * Stopwatch to keep track of the elapsed time since the last gc call.
+   */
+  private final Stopwatch stopwatch = Stopwatch.createUnstarted();
+
+  /**
+   * The maximum number of runs allowed before calling System.gc().
+   */
+  private static final int MAX_RUNS_BEFORE_GC = 5;
```
AustinShalit: Should this be called minimum instead?

core/src/main/java/edu/wpi/grip/core/operations/composite/WatershedOperation.java
AustinShalit: If we use Guava 21.0:
`markerPool = Stream.generate(Mat::new).limit(MAX_MARKERS).collect(ImmutableList.toImmutableList());`

core/src/main/java/edu/wpi/grip/core/Cleaner.java
SamCarlberg: Yeah, I should change that

PR #702, opened 2016-11-03
Add better support for command line arguments

SamCarlberg:
Set the save file with `-f <path>` or `--file <path>`  
Set the HTTP server port with `-p <port>` or `--port <port>`  
Print the version with `-v` or `--version`  
Print help text with `-h` or `--help`  
Set headless mode with `--headless` (UI only)

Can pass command line args through gradle with `-Pargs="<args>"`

Note: this breaks backwards compatibility with loading files from the command line.


JLLeitschuh:
Did you try this in a deployed application?

SamCarlberg:
If by "deployed" you mean "core app", yes

Issue #710, opened 2016-11-09
Resize too-large images to fit the preview pane


SamCarlberg:
So instead of showing full-size 1920x1080 px images (for example), the preview should be resized so the whole image fits in the preview pane without needing to scroll.

Related: #706

SamCarlberg:
Closed by #711

PR #711, opened 2016-11-10
Dynamically resize image previews to fit the pane

### LINKED ISSUE #706: Previewing really big images uses a ton of memory and looks bad
(4 bytes per pixel) x (a lot of pixels) = a lot of memory

We should prompt the user to add a resize step, as well as one of:

- Refuse to display the image, and instead show a message like "The image is too large to preview"
- Show a scaled-down version of the image and show the scale factor (1/2 size, 1/4 size, 1/8 size, ...)

SamCarlberg:
Closes #706 and #710

[Example](https://gfycat.com/AcrobaticEdibleBalloonfish)


SamCarlberg:
Merging master should fix that. This branch is pretty out-of-date.

PR #715, opened 2016-11-20
Avoid name resolution issue in C++ code generation

phurley:
Found a simple issue with the C++ code generation, where a class method named blur, was causing a name resolution error. The easiest fix was to explicitly use the cv:: namespace in the generated code. It would probably be best to not even use the using namespace (especially as currently done in a header file) and always used explicit namespaces in the generated code.


SamCarlberg:
If there are other places where the full `cv::` namespace should be specified, you should change those too.

You also need to remove the `using namespace cv` line from the template

SamCarlberg:
I'm going to assume that this was closed in favor of #734

phurley:
Yes, it was a much bigger change, with a much larger surface, so it was easier. Sorry for any confusion.

PR #734, opened 2016-12-02
Remove using namespace from generated header, fix the fall out

phurley:
Lots of simple changes to the C++ code generation for C++. In particular the generated code has 

using namespace cv;
using namespace std;

in the generated header file, which is a pretty bad idea. I checked all of the generated code options that previously worked still do. I will try and create a few issues/pulls where I found other issues. I also fixed a couple issues with the existing code generator where some generated code was failing due to name resolution.


SamCarlberg:
Can you put the pipeline in its own namespace as well? Even just `namespace grip` would be good.

phurley:
Made the change namespace change, it requires a minor change to using the generated code, which could be considered a breaking change, but it is cleaner.

SamCarlberg:
LGTM. Are there any other similar problems you've noticed that you think should be included?

ui/src/main/resources/edu/wpi/grip/ui/codegeneration/cpp/Pipeline.vm
```diff
@@ -114,6 +117,9 @@ void $fileName::set$step.name()${step.num()}(bool value){
 #parse($toParse)
 
 #end
+
+} // end grip namespace
```
SamCarlberg: This should go after the `#if($testing)` block

ui/src/main/resources/edu/wpi/grip/ui/codegeneration/cpp/Pipeline.h.vm
```diff
@@ -86,6 +91,9 @@ class $fileName #if($testing): public AbsPipeline #end{
 #end
 #end
 };
+
+} // end namespace grip
```
SamCarlberg: This should go after the `#if($testing)` block

phurley: Done, and committed, tests pass.

SamCarlberg:
@phurley Is there anything else you'd like to add to this PR before merging it?

phurley:
No, I think this is ready to merge. I do not have currently have any more fixes for the code generator. I hope to make a more careful walk through the generated code over the weekend, but any issues will not be related to namespaces, and should be made in a different pull with a corresponding issue if necessary.

SamCarlberg:
Okay

PR #746, opened 2016-12-10
gradle improvements

JLLeitschuh:

Various improvements to gradle to make it [better, faster, stronger](https://www.youtube.com/watch?v=GDpmVUEjagg).

Unrelated/offtopic:
There should be far less of [this](https://youtu.be/dQw4w9WgXcQ) in the build.


build.gradle
```diff
@@ -599,5 +616,7 @@ task jacocoRootReport(type: JacocoReport, group: 'Coverage reports') {
 }
 
 task wrapper(type: Wrapper) {
-    gradleVersion = '3.0'
+    gradleVersion = '3.2.1'
+    // We want the sources with the wrapper.
+    distributionType = DistributionType.ALL
```
JLLeitschuh: This should finally fix intellij constantly running the wrapper task and changing the wrapper properties

build.gradle
```diff
@@ -23,6 +25,7 @@ import edu.wpi.first.wpilib.opencv.installer.Installer
 import edu.wpi.first.wpilib.opencv.installer.platform.Platform
 
 plugins {
+    id 'com.gradle.build-scan' version '1.3'
```
AustinShalit: Why?

JLLeitschuh: Because its cool and gives us insights on the build when it fails while in CI.

PR #753, opened 2016-12-13
Discourage use of the deploy tool

SamCarlberg:
An alert will show every time, even in the same session, for maximum annoyance

### DIFF
--- a/ui/src/main/java/edu/wpi/grip/ui/MainWindowController.java
+++ b/ui/src/main/java/edu/wpi/grip/ui/MainWindowController.java
@@ -308,6 +308,13 @@ public class MainWindowController {
 
   @FXML
   protected void deploy() {
+    eventBus.post(new WarningEvent(
+        "Deploy has been deprecated",
+        "The deploy tool has been deprecated and is no longer supported. "
+            + "It will be removed in a future release.\n\n"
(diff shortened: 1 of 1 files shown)


AustinShalit:
When this is merged.  Lets close all deploy issues.  https://github.com/WPIRoboticsProjects/GRIP/labels/deploy

JLLeitschuh:
I disagree with the decision of removing the tool.
@EdWard680 Don't people rely upon it for deploying grip to a PI?
Does code gen work on a PI just as fast?

SamCarlberg:
> I disagree with the decision of removing the tool.

We're not removing it, just showing an alert that discourages its use because code generation is the better solution.

> Does code gen work on a PI just as fast?

Do you mean, "does the generated code run just as fast as GRIP?" or "is the workflow just as fast for using generated code as it is for deploying GRIP?"

EdWard680:
@JLLeitschuh I would agree that code-gen should be encouraged, and the tool currently doesn't do anything with code-gen, however I don't think this should be a deprecation of the concept of a full "deploy." I think teams would find the most use from a modified deploy tool which deploys an executable/script built from grip code gen, but until something like that were to be made, this seems like a good deprecation.

AustinShalit:
@bradamiller Thoughts on this?

PR #757, opened 2016-12-22
Use Grgit instead of command-line Git invocation

JacisNonsense:
This should fix all accounts where Git is not on the path. For cases where Git is not present, we return the commit as `<no commit>` and the Git tag/version as `v0.0.0`


JLLeitschuh:
Rerunning build. The failures aren't your fault.

AustinShalit:
https://travis-ci.org/WPIRoboticsProjects/GRIP/jobs/186028549#L2369

PR #758, opened 2016-12-22
Log when HTTP server can't be started (also shows alert in UI)

SamCarlberg:
Change default port from 8080 to 2084


AustinShalit:
If it cannot start the server, can we still let GRIP start? Exiting seems like the wrong solution.

SamCarlberg:
We might be able to show a confirmation dialog to ask if we should continue without the server, but that would make it impossible to use any of the HTTP stuff. I don't think that'd be good.

AustinShalit:
That would be ok for the normal end user.  If I am at an event using GRIP and something gets onto the port GRIP uses, I would be very upset that GRIP is not launching.  The HTTP stuff is not required for GRIP to work.

PR #760, opened 2016-12-28
opencv-python package is broken; disable python codegen tests

SamCarlberg:
This should be reverted once the package works again.

See #761

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ script:
   # will not be able to run MainWindowTest.testDragOperationFromPaletteToPipeline
   - |
     if [[ "$TRAVIS_OS_NAME" == "osx" ]];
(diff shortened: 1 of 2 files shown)


SamCarlberg:
Closing this because the problem is with `opencv-python` only supporting up to 3.5, while homebrew installs 3.6

SamCarlberg:
Unfortunately, there doesn't seem to be an easy way to get homebrew to install a supported version of Python for the `opencv-python` package. We'll just have to disable the python tests until a package for 3.6 is released (which is currently waiting on a numpy release)

AustinShalit:
Other options: http://apple.stackexchange.com/questions/237430/how-to-install-specific-version-of-python-on-os-x

PR #790, opened 2017-01-17
Improve logging

SamCarlberg:
The current logs are about 1100 lines of debugging stuff from guava and about zero lines of stuff from GRIP. This PR reverses that situation.

### DIFF
--- a/core/src/main/java/edu/wpi/grip/core/Connection.java
+++ b/core/src/main/java/edu/wpi/grip/core/Connection.java
@@ -7,7 +7,9 @@ import edu.wpi.grip.core.events.SourceRemovedEvent;
 import edu.wpi.grip.core.events.StepRemovedEvent;
 import edu.wpi.grip.core.sockets.InputSocket;
 import edu.wpi.grip.core.sockets.OutputSocket;
+import edu.wpi.grip.core.sockets.Socket;
 
+import com.google.common.base.MoreObjects;
 import com.google.common.eventbus.EventBus;
 import com.google.common.eventbus.Subscribe;
 import com.google.inject.Inject;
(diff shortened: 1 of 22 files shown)


core/src/main/java/edu/wpi/grip/core/events/EventLogger.java
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Class for logging events as they're posted to the event bus.
+ */
+@Singleton
+public class EventLogger {
+
+  private static final Logger logger = Logger.getLogger(EventLogger.class.getName());
+
+  @Subscribe
+  public void eventPosted(LoggableEvent event) {
```
JLLeitschuh: Do you want to make it so that this event can be handled by multiple threads simultaneously?
If so you need to add that other annotation (not sure what its called)

SamCarlberg: `@AllowConcurrentEvents` I think

core/src/main/java/edu/wpi/grip/core/Connection.java
```diff
@@ -112,6 +114,27 @@ public void remove() {
...
 
+  @Override
+  public String toString() {
+    return MoreObjects.toStringHelper(this)
+        .add("src", simpleString(outputSocket))
+        .add("dst", simpleString(inputSocket))
+        .toString();
+  }
+
+  private static String simpleString(Socket<?> socket) {
```
JLLeitschuh: Why is this method in the connection class instead of a helper method on the `Socket` object??
Throw this in a default method on the `Socket` interface.

core/src/main/java/edu/wpi/grip/core/events/EventLogger.java
SamCarlberg: I'm not going to add the `@AllowConcurrentEvents` annotation; the event bus is synchronous so only one event will be posted at a time anyway.

PR #798, opened 2017-01-23
Fix various issues with generated C++

### LINKED ISSUE #797: Generated C++ doesn't compile
- The .cpp still declares the `setsourceN` functions
- VIsionPipeline include is wrong (should be `#include "vision/VisionPipeline.h"`)
- FindBlobs is missing a `cv::` namespace qualifier
- Includes the contrib module (and several others) even when it's not used, and may not compile on systems that don't or can't have it installed (e.g. FRC)

SamCarlberg:
Closes #797


SamCarlberg:
@JLLeitschuh C++ code gen tests are broken. But I've been generating code into an FRC project and compiling it there.

PR #818, opened 2017-02-05
Publish core jar

### LINKED ISSUE #787: Release core.jar as a separate release
We will need to have one per OS because we only include the library for the OS we are building on.

AustinShalit:
closes #787


JLLeitschuh:
You need to do the same thing for windows given that our shadow jar doesn't include binaries for all OS'es.

Maybe we could create a jar that contains all of them. That should be pretty easy with gradle.

But in the short term @AustinShalit can you add a config change to the appveyor script too?

JLLeitschuh:
@AustinShalit can you take care of this?

JLLeitschuh:
Does this work on windows? Just checking. Its fine if you haven't tested it.

AustinShalit:
Hold off.  Still testing

JLLeitschuh:
Status?

AustinShalit:
Broken

PR #822, opened 2017-02-10
Continuously backup changes and load them when GRIP is launched

SamCarlberg:


JLLeitschuh:
The only problem I forsee is if the backup file turns out to be corrupted or mal-formed then it makes it impossible to load GRIP without deleting this file.

SamCarlberg:
Well this is a pain... different save files of the same project aren't guaranteed to be textually identical, even if they're functionally identical.

AustinShalit:
Can you use time modified?

SamCarlberg:
> Can you use time modified?

The problem with that is if someone makes a change and undoes it, the backup will be more recent and GRIP will think the save file's different than what's been saved, even though the backup is identical. Comparing the text seems to work now, though. Changed the connections set to a `LinkedHashSet` to preserve insertion order.

ui/src/main/java/edu/wpi/grip/ui/ProjectBackupLoader.java
```diff
@@ -0,0 +1,94 @@
...
+    } else if (GripFileManager.BACKUP_FILE.exists()) {
+      // Load the backup, if possible
+      loadBackup();
+      project.setFile(Optional.empty());
+    }
+  }
+
+  private File lastSaveFile() throws IOException {
+    List<String> lines = Files.readAllLines(GripFileManager.LAST_SAVE_FILE.toPath());
+    if (lines.size() == 1) {
```
JLLeitschuh: Why if there is only one line in the file is it considered valid? I'm confused.

ui/src/main/java/edu/wpi/grip/ui/ProjectBackupLoader.java
```diff
@@ -0,0 +1,94 @@
...
+    } else {
+      logger.warning("Unexpected data in last_save file: " + lines);
+      return null;
+    }
+  }
+
+  private void loadBackup() {
+    try {
+      logger.info("Loading backup file");
+      project.open(GripFileManager.BACKUP_FILE);
```
JLLeitschuh: If this fails won't the pipeline be in a bogus state and be unusable? this could even crash GRIP because the APP was in a bad state.
If this happens then the user can't open GRIP without deleting the load backup file correct?

ui/src/main/java/edu/wpi/grip/ui/ProjectBackupLoader.java
```diff
@@ -0,0 +1,94 @@
...
+    } catch (XStreamException | IOException e) {
+      logger.log(Level.WARNING, "Could not load backup file", e);
+    }
+    eventBus.post(DirtiesSaveEvent.DIRTIES_SAVE_EVENT);
```
JLLeitschuh: Should this also fire this event if the deserialization fails?

JLLeitschuh (changes requested):
I like the overall idea however, this sort of thing could expose an underlying design flaw in GRIP where loading a save loads the file into the active pipeline as it runs. This could put GRIP in a bogus state.

A better solution for GRIP would have been to have XStream read in a data model that then gets copied into what is the running pipeline once everything has been validated correctly.

That being said, I think this could work as it is but I'd want to know it had been tested and was resilient to failures.

Have you tried making the backup something that is bogus (for example half the file isn't there because GRIP is closed mid write of the stream).

Also, adding comments where I indicated would be really helpful.

core/src/main/java/edu/wpi/grip/core/events/DirtiesSaveEvent.java
```diff
@@ -8,6 +8,13 @@
  */
 public interface DirtiesSaveEvent {
 
+  DirtiesSaveEvent DIRTIES_SAVE_EVENT = new DirtiesSaveEvent() {
```
SamCarlberg: Fields and methods declared in interfaces are always `public static final`

ui/src/main/java/edu/wpi/grip/ui/ProjectBackupLoader.java
SamCarlberg: It does... That's why the event is _after_ the whole try-catch block.

core/src/main/java/edu/wpi/grip/core/Pipeline.java
```diff
@@ -61,7 +61,7 @@
    */
   private final List<Source> sources = new ArrayList<>();
   private final List<Step> steps = new ArrayList<>();
-  private final Set<Connection> connections = new HashSet<>();
+  private final Set<Connection> connections = new LinkedHashSet<>();
```
SamCarlberg: I'm actually going to revert this. Just looking at the last modified time now.

ui/src/main/java/edu/wpi/grip/ui/ProjectBackupLoader.java
SamCarlberg: Welp... guess I'm going to rebase this onto #692, which fixes all of those problems

core/src/main/java/edu/wpi/grip/core/serialization/Project.java
```diff
@@ -88,8 +95,27 @@ public void initialize(StepConverter stepConverter,
...
+    file.ifPresent(f -> logger.info("Setting project file to: " + f.getAbsolutePath()));
     this.file = file;
+    try {
+      if (file.isPresent()) {
+        Files.write(file.get().getAbsolutePath(),
+            GripFileManager.LAST_SAVE_FILE,
+            Charset.defaultCharset());
+      } else {
+        // No project file, delete the last_save file
+        GripFileManager.LAST_SAVE_FILE.delete();
```
SamCarlberg: Basically, the `LAST_SAVE_FILE` stores the full path to the most recently loaded save file. If no file was set, then the `LAST_SAVE_FILE` is deleted to prevent auto-loading a stale file.

ui/src/main/java/edu/wpi/grip/ui/ProjectBackupLoader.java
SamCarlberg: Or, since #692 is pretty big and breaking... maybe leave this as is and have #692 address this specific problem

PR #832, opened 2017-03-01
Add support for custom CSS skins

SamCarlberg:
I'm opening this PR just to make this visible to anyone who wants to give input during development. There is still a _lot_ of room for improvement, visually.

This also needs to get reworked to store arbitrarily many themes in ~/GRIP and offer a UI to choose between them, possibly with previews (like how Slack does it)

#### Empty
Empty

#### Add a source
Add webcam

#### Warning
Warning alert

#### With steps and connections
With steps

#### Deploy dialog
Deploy

#### Analysis window
Analysis

#### Settings
screenshot from 2017-03-01 00-15-18

#### About
screenshot from 2017-03-01 00-16-34


JLLeitschuh:
🎆  Yes!!! I like this!!! 🎆

bradamiller:
Sam

I was curious how much work you figured it would take to add the UI to be able to select themes and if that was all that it needed to be finished? It seems reasonable to start out with the two (old look and dark mode), then think about more or get contributions.

ui/src/main/resources/edu/wpi/grip/ui/GRIP-dark-controls.css
```diff
@@ -0,0 +1,86 @@
...
+    -fx-background-color: gripgreen;
+}
+
+.range-slider .track, .slider .track {
+    -fx-background-color: base-dark;
+}
+
+.range-slider .high-thumb, .range-slider .low-thumb, .slider .thumb {
+    -fx-background-color: gripsandy;
+}
```
JLLeitschuh: Newline at end of file.

SamCarlberg: Done. Thought I'd already done that

nightpool:
looking really good! some feedback:

image

The white borders around menu bars are really distracting. Compare the subtler shadow effect on the light theme:

image

---

image

Very little contrast in the background for the title of each section. The light grey is very close to the light green of the section borders, making it hard to distinguish the background of that title section from the borders, and thus muddying the borders in general. Compare to the high contrast of the light theme:

image

I would recommending getting rid of the backgrounds in this section entirely, adding dark borders where they don't already exist (preview and palette, sources already has a border for some reason?) and making the text bigger (because the white has less contrast on the grey then the black does on the tan background, so you should compensate by increasing the text size)

----

The search bar not having any borders is kinda weird:
image
image

---

there should be more space between these two buttons if there isn't going to be a border between them:

image

---

The stop button here should be bigger. it's really hard to see/understand right now:

image

Also, the "pulsing" play button is very light/hard to see:

grip-playbutton

----

some weird white lines on the "benchmark" screen:

image

and in the "file" menu:

image

those should probably both be dark

---

If we're doing borderless, this dialogue needs more spacing too:

image

(add network table source)

nightpool:
In general, there are two main "overall" design issues that need to be addressed. 

 - The dark theme is flat, while the light theme is a more traditional skeuomorphic look-and-feel. this is good—I like it a lot more—but it requires a little more care in choosing spacing, where borders go, where we're using shadows, etc. Right now we're pretty inconsistent on where we're using borders. For example, we have a solid thick black border around preview output, and in the pipeline, but no borders at all in the palette or around headers. this leads to an inconsistent visual metaphor and a confusing user experience.

 - Everything is pretty weirdly blue. I really like it for the "selected" state, but I feel like the body text needs to be more desaturated—it feels like i'm using a 70s 2-color CRT terminal. Also the body text is honestly slightly too bright. Tweaking the main text color should fix both of these problems.

JLLeitschuh:
@nightpool Thanks for taking the time to review this PR and really give a solid review of the coloring and design.

SamCarlberg:
So these tests are failing on Travis, Appveyor, and my Linux laptop... but not on my Windows 10 desktop. I'm very tempted to just `@Ignore` all those tests since it seems (to me) to be an issue with testfx

SamCarlberg:
Closing this PR. I may open a PR in the future with a better theme

PR #837, opened 2017-03-22
Add the GRIP MQP pdf and other presentation docs

JLLeitschuh:
This adds the presentation documents to the GRIP project so that they are under source control and live with the project.

I debated between this and creating a separate repository.
I think this makes a lot of sense because the documentation should live with the project that is being documented.


JLLeitschuh:
Appveyor seems to have missed the boat on this build. I don't think it's totally necessary for it to pass.

AustinShalit:
Want to try to amend and force push to see if we can trigger it?

JLLeitschuh:
Not really, I'll probably just merge it myself.

Issue #783, opened 2017-01-11
Webcam source memory leak


johnyburd:
The title pretty much says it all.

![memory](https://cloud.githubusercontent.com/assets/4121471/21852418/f0b24e7c-d7e0-11e6-97b0-7122b102dd0c.png)
![webcam](https://cloud.githubusercontent.com/assets/4121471/21852419/f0b3dcb0-d7e0-11e6-8ec3-d9e67ade9ceb.png)

Whenever I have webcam selected as the source, the memory usage constantly goes up until my computer crashes.
I'm running arch linux and this happens with or without a preview selected.

johnyburd:
I saw on the raspberry pi page that there was a bug with USB webcams, but I have also reproduced this on an HTTP stream and even a static image

SamCarlberg:
How much memory do you have installed? GRIP is configured to use up to 200MB on the Java heap but will tend to use more overall due to OpenCV's native memory use

johnyburd:
10GB

SamCarlberg:
And what version of GRIP?

johnyburd:
I also noticed the find line, contour, and blob modules all agrevate the problem.

johnyburd:
The version is the latest from GitHub

SamCarlberg:
If you go to the `Help > About` menu, what version does it report?

johnyburd:
1.5.1

SamCarlberg:
Then it's most likely a problem with Arch. I can't reproduce this on Elementary OS

johnyburd:
That is very strange. Any work around that isn't changing Os's?

SamCarlberg:
If you have a JDK installed, you can launch VisualVM to see how much of the heap is being used and how large the heap is. If the heap is small (< 50%) compared to the total memory use you see in the system monitor, then OpenCV objects aren't getting cleaned up.

thomassross:
Having the same issue on Debian Stretch

SamCarlberg:
@thomassross are you building with Gradle or using the native app from our releases page?

thomassross:
@SamCarlberg I'm using the latest deb release on the releases page.

thomassross:
I am not getting this issue with a static image instead of a webcam.

SamCarlberg:
That's expected. Changing images allocates more memory. If you're comfortable with it, you can install VisualVM and the GC plugin for it. Garbage collection should be running about once per second

[...]

PR #838, opened 2017-03-23
Release converted images in GrabberService after they're no longer used

SamCarlberg:
Helps relieve some memory pressure.
Could help #783


JLLeitschuh:
The frame grabber writes memory into the same mat every read. Look at the Webcam frame grabber.

This will break something.

JLLeitschuh:
Looking at the tests... It seems that it did break something.

SamCarlberg:
It only breaks the tests though, it works fine in the real world. Would probably have to compare the frames to see if a change was made

SamCarlberg:
I'm going to scrap this for the time being

PR #866, opened 2017-11-09
Use cscore for PublishVideoOperation

SamCarlberg:
Updates JavaCV to 1.3.3 and javacpp-opencv to 3.2.0 in order to keep the same memory model between bindings.

Allows up to ten publish video operations on ports 1180 to 1189.

Note: this increases the size of the core jar by ~67MB from bundling the cscore and opencv binaries.

#### GRIP pipeline
grip-cscore-pipeline

#### Video streams in Shuffleboard using cscore
grip-streams-in-shuffleboard

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -225,10 +225,10 @@ project(":core") {
 
     dependencies {
         compile group: 'com.google.code.findbugs', name: 'jsr305', version: '3.0.1'
-        compile group: 'org.bytedeco', name: 'javacv', version: '1.1'
-        compile group: 'org.bytedeco.javacpp-presets', name: 'opencv', version: '3.0.0-1.1'
-        compile group: 'org.bytedeco.javacpp-presets', name: 'opencv', version: '3.0.0-1.1', classifier: os
-        compile group: 'org.bytedeco.javacpp-presets', name: 'opencv-3.0.0-1.1', classifier: 'linux-frc'
+        compile group: 'org.bytedeco', name: 'javacv', version: '1.3.3'
+        compile group: 'org.bytedeco.javacpp-presets', name: 'opencv', version: '3.2.0-1.3'
+        compile group: 'org.bytedeco.javacpp-presets', name: 'opencv', version: '3.2.0-1.3', classifier: os
+        //compile group: 'org.bytedeco.javacpp-presets', name: 'opencv-3.0.0-1.1', classifier: 'linux-frc'
         compile group: 'org.bytedeco.javacpp-presets', name: 'videoinput', version: '0.200-1.1', classifier: os
         compile group: 'org.bytedeco.javacpp-presets', name: 'ffmpeg', version: '0.200-1.1', classifier: os
         compile group: 'org.python', name: 'jython', version: '2.7.0'
@@ -253,6 +253,13 @@ project(":core") {
         compile group: 'org.ros.rosjava_messages', name: 'grip_msgs', version: '0.0.1'
         compile group: 'edu.wpi.first.wpilib.networktables.java', name: 'NetworkTables', version: '3.1.2', classifier: 'desktop'
         compile group: 'edu.wpi.first.wpilib.networktables.java', name: 'NetworkTables', version: '3.1.2', classifier: 'arm'
+
+        // cscore dependencies
+        compile group: 'edu.wpi.first.cscore', name: 'cscore-java', version: '1.1.0-beta-2'
+        compile group: 'edu.wpi.first.cscore', name: 'cscore-jni', version: '1.1.0-beta-2', classifier: 'all'
+        runtime group: 'edu.wpi.first.wpiutil', name: 'wpiutil-java', version: '3.+'
+        compile group: 'org.opencv', name: 'opencv-java', version: '3.2.0'
+        compile group: 'org.opencv', name: 'opencv-jni', version: '3.2.0', classifier: 'all'
     }
 
     mainClassName = 'edu.wpi.grip.core.Main'
--- a/core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
+++ b/core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
@@ -6,6 +6,7 @@ import edu.wpi.grip.core.sockets.InputSocket;
 import edu.wpi.grip.core.sockets.OutputSocket;
 import edu.wpi.grip.core.sockets.SocketHints;
 import edu.wpi.grip.core.util.Icon;
+import edu.wpi.grip.core.util.OpenCvShims;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
@@ -90,7 +91,7 @@ public class PublishVideoOperation implements Operation {
   // applications connected to the same NetworkTable server (eg Shuffleboard)
   private final ITable cameraPublisherTable = NetworkTable.getTable("/CameraPublisher"); // NOPMD
   private final ITable ourTable;
-  private Mat publishMat = null;
+  private final Mat publishMat = new Mat();
   private long lastFrame = -1;
 
   @SuppressWarnings("JavadocMethod")
@@ -160,16 +161,7 @@ public class PublishVideoOperation implements Operation {
       throw new IllegalArgumentException("Input image must not be empty");
     }
 
-    // "copy" the input data to an OpenCV mat for cscore to use
-    // This basically just wraps the mat pointer in a different wrapper object
-    // No copies are performed, but it means we have to be careful about making sure we use the
-    // same version of JavaCV and OpenCV to minimize the risk of binary incompatibility.
-    // This copy only needs to happen once, since the operation input image is always the same
-    // object that gets copied into.  The data address will change, however, if the image is resized
-    // or changes type.
-    if (publishMat == null || publishMat.nativeObj != input.address()) {
-      publishMat = new Mat(input.address());
-    }
+    OpenCvShims.copyJavaCvMatToOpenCvMat(input, publishMat);
     // Make sure the output resolution is up to date. Might not be needed, depends on cscore updates
     serverSource.setResolution(input.size().width(), input.size().height());
     serverSource.putFrame(publishMat);
--- a/core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
+++ b/core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
@@ -7,6 +7,7 @@ import edu.wpi.grip.core.sockets.OutputSocket;
 import edu.wpi.grip.core.sockets.SocketHints;
 import edu.wpi.grip.core.util.Icon;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
@@ -17,15 +18,12 @@ import edu.wpi.cscore.VideoMode;
 import edu.wpi.first.wpilibj.networktables.NetworkTable;
 import edu.wpi.first.wpilibj.tables.ITable;
 
-import org.apache.commons.lang.SystemUtils;
 import org.bytedeco.javacpp.opencv_core;
 import org.opencv.core.Mat;
 
-import java.io.BufferedReader;
-import java.io.IOException;
-import java.io.InputStreamReader;
 import java.lang.reflect.Field;
-import java.util.Arrays;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
 import java.util.Deque;
 import java.util.LinkedList;
 import java.util.List;
@@ -115,8 +113,16 @@ public class PublishVideoOperation implements Operation {
       server.setSource(serverSource);
 
       ourTable = cameraPublisherTable.getSubTable("GRIP-" + totalStepCount);
-      ourTable.putStringArray("streams",
-          new String[]{"mjpeg:http://" + getHostName() + ":" + ourPort + "/?action=stream"});
+      try {
+        InetAddress localHost = InetAddress.getLocalHost();
+        ourTable.putStringArray("streams",
+            new String[]{
+                generateStreamUrl(localHost.getHostName(), ourPort),
+                generateStreamUrl(localHost.getHostAddress(), ourPort)
+            });
+      } catch (UnknownHostException e) {
+        ourTable.putStringArray("streams", new String[0]);
+      }
     } else {
       server = null;
       serverSource = null;
@@ -177,6 +183,10 @@ public class PublishVideoOperation implements Operation {
     }
   }
 
+  private static String generateStreamUrl(String host, int port) {
+    return String.format("mjpeg:http://%s:%d/?action=stream", host, port);
+  }
+
   /**
    * Copies the data from a JavaCV Mat wrapper object into an OpenCV Mat wrapper object so it's
    * usable by the {@link CvSource} for this operation.
@@ -195,7 +205,8 @@ public class PublishVideoOperation implements Operation {
    * @param openCvMat the OpenCV Mat wrapper object to copy into
    * @throws RuntimeException if the OpenCV native pointer could not be set
    */
-  private static void copyJavaCvToOpenCvMat(opencv_core.Mat javaCvMat, Mat openCvMat)
+  @VisibleForTesting
+  static void copyJavaCvToOpenCvMat(opencv_core.Mat javaCvMat, Mat openCvMat)
       throws RuntimeException {
     // Make the OpenCV Mat object point to the same block of memory as the JavaCV object.
     // This requires no data transfers or copies and is O(1) instead of O(n)
@@ -208,42 +219,6 @@ public class PublishVideoOperation implements Operation {
         logger.log(Level.WARNING, "Could not set native object pointer", e);
         throw new RuntimeException("Could not copy the image", e);
       }
-    }
-  }
-
-  /**
-   * Multi platform method for getting the hostname of the local computer. cscore's
-   * {@link CameraServerJNI#getHostname() getHostName() function} only works on Linux, so we need to
-   * implement the method for Windows and Mac ourselves.
-   */
-  private static String getHostName() {
-    if (SystemUtils.IS_OS_WINDOWS) {
-      // Use the Windows `hostname` command-line utility
-      // This will return a single line of text containing the hostname, no parsing required
-      ProcessBuilder builder = new ProcessBuilder("hostname");
-      Process hostname;
-      try {
-        hostname = builder.start();
-      } catch (IOException e) {
-        logger.log(Level.WARNING, "Could not start hostname process", e);
-        return "";
-      }
-      try (BufferedReader in =
-               new BufferedReader(new InputStreamReader(hostname.getInputStream()))) {
-        return in.readLine() + ".local";
-      } catch (IOException e) {
-        logger.log(Level.WARNING, "Could not read the hostname process output", e);
-        return "";
-      }
-    } else if (SystemUtils.IS_OS_LINUX) {
-      // cscore already defines it for linux
-      return CameraServerJNI.getHostname();
-    } else if (SystemUtils.IS_OS_MAC) {
-      // todo
-      return "TODO-MAC";
-    } else {
-      throw new UnsupportedOperationException(
-          "Unsupported operating system " + System.getProperty("os.name"));
     }
   }
 
--- a/core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
+++ b/core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
@@ -10,34 +10,42 @@ import edu.wpi.grip.core.util.Icon;
 import com.google.common.collect.ImmutableList;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import edu.wpi.cscore.CameraServerJNI;
+import edu.wpi.cscore.CvSource;
+import edu.wpi.cscore.MjpegServer;
+import edu.wpi.cscore.VideoMode;
+import edu.wpi.first.wpilibj.networktables.NetworkTable;
 
-import org.bytedeco.javacpp.BytePointer;
-import org.bytedeco.javacpp.IntPointer;
+import org.bytedeco.javacpp.opencv_core;
+import org.opencv.core.Mat;
 
-import java.io.DataInputStream;
-import java.io.DataOutputStream;
-import java.io.IOException;
-import java.net.ServerSocket;
-import java.net.Socket;
+import java.lang.reflect.Field;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-import static org.bytedeco.javacpp.opencv_core.Mat;
-import static org.bytedeco.javacpp.opencv_imgcodecs.CV_IMWRITE_JPEG_QUALITY;
-import static org.bytedeco.javacpp.opencv_imgcodecs.imencode;
+import static org.bytedeco.javacpp.opencv_core.CV_8S;
+import static org.bytedeco.javacpp.opencv_core.CV_8U;
 
 /**
  * Publish an M-JPEG stream with the protocol used by SmartDashboard and the FRC Dashboard.  This
  * allows FRC teams to view video streams on their dashboard during competition even when GRIP has
- * exclusive access to the camera.  In addition, an intermediate processed image in the pipeline
- * could be published instead. Based on WPILib's CameraServer class:
- * https://github.com/robotpy/allwpilib/blob/master/wpilibj/src/athena/java/edu/wpi/first/wpilibj
- * /CameraServer.java
+ * exclusive access to the camera. Uses cscore to host the image streaming server.
  */
 public class PublishVideoOperation implements Operation {
 
   private static final Logger logger = Logger.getLogger(PublishVideoOperation.class.getName());
+
+  static {
+    try {
+      // Loading the CameraServerJNI class will load the appropriate platform-specific OpenCV JNI
+      CameraServerJNI.getHostname();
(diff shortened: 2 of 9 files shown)


build.gradle
```diff
@@ -225,10 +225,10 @@ project(":core") {
...
     dependencies {
         compile group: 'com.google.code.findbugs', name: 'jsr305', version: '3.0.1'
-        compile group: 'org.bytedeco', name: 'javacv', version: '1.1'
-        compile group: 'org.bytedeco.javacpp-presets', name: 'opencv', version: '3.0.0-1.1'
-        compile group: 'org.bytedeco.javacpp-presets', name: 'opencv', version: '3.0.0-1.1', classifier: os
-        compile group: 'org.bytedeco.javacpp-presets', name: 'opencv-3.0.0-1.1', classifier: 'linux-frc'
+        compile group: 'org.bytedeco', name: 'javacv', version: '1.3.3'
+        compile group: 'org.bytedeco.javacpp-presets', name: 'opencv', version: '3.2.0-1.3'
+        compile group: 'org.bytedeco.javacpp-presets', name: 'opencv', version: '3.2.0-1.3', classifier: os
+        //compile group: 'org.bytedeco.javacpp-presets', name: 'opencv-3.0.0-1.1', classifier: 'linux-frc'
```
SamCarlberg: Commented out because there's no `opencv-3.2.0-1.3` with the `linux-frc` classifier and I'm worried about different memory models between binding versions (JavaCV with 3.0.0, OpenCV with 3.2.0) . This could get removed entirely if we remove the deploy functionality.

JLLeitschuh:
Travis fails, this is a non-starter for me.
We need to introduce the docker based build system to resolve this first before this gets merged.

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
```diff
@@ -167,25 +125,51 @@ public PublishVideoOperation(InputSocket.Factory inputSocketFactory) {
...
+    // Make sure the output resolution is up to date
+    serverSource.setResolution(size.width(), size.height());
+
+    // Make the OpenCV Mat object point to the same block of memory as the JavaCV object.
+    // This requires no data transfers or copies and is O(1) instead of O(n)
+    if (javaCvMat.address() != openCvMat.nativeObj) {
+      try {
+        Field nativeObjField = Mat.class.getField("nativeObj");
+        nativeObjField.setAccessible(true);
+        nativeObjField.setLong(openCvMat, javaCvMat.address());
```
JLLeitschuh: This seems like a terrible idea, but I understand what you are doing here.
Do you actually see speed benefits from doing this?
@saudet Is there any reason why doing this assignment could shoot us in the foot?

This is taking the memory address of the OpenCV mat and basically telling JavaCV to use it's memory instead.

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
```diff
@@ -46,110 +59,55 @@
...
-        connected = false;
-      }
-    }
-  };
+  private final MjpegServer server;
+  private final CvSource serverSource;
+
+  // Write to the /CameraPublisher table so the MJPEG streams are discoverable by other
+  // applications connected to the same NetworkTable server (eg Shuffleboard)
+  private static final ITable cameraPublisherTable = NetworkTable.getTable("/CameraPublisher");
```
JLLeitschuh: Can you not load this statically? Instead do it at instantiation time.

SamCarlberg: What would that gain?

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
```diff
@@ -10,34 +10,47 @@
...
+
+  static {
+    try {
+      // Loading the CameraServerJNI class will load the appropriate platform-specific OpenCV JNI
+      CameraServerJNI.getHostname();
+    } catch (Throwable e) {
+      logger.log(Level.SEVERE, "CameraServerJNI load failed! Exiting", e);
+      System.exit(31);
+    }
+  }
```
JLLeitschuh: Does this need to happen statically? Can't this be done when the operation is constructed?

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
JLLeitschuh: If it fails then GRIP doesn't crash immediately.

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
```diff
@@ -10,34 +10,47 @@
...
 
   private static final Logger logger = Logger.getLogger(PublishVideoOperation.class.getName());
+
+  static {
+    try {
+      // Loading the CameraServerJNI class will load the appropriate platform-specific OpenCV JNI
+      CameraServerJNI.getHostname();
+    } catch (Throwable e) {
+      logger.log(Level.SEVERE, "CameraServerJNI load failed! Exiting", e);
+      System.exit(31);
```
SamCarlberg: This should set a flag and make the operation throw an exception in `perform()` instead of suddenly exiting the entire application

JLLeitschuh: Yea, we want to limit the number crashes that don't display the crash dialog.

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
SamCarlberg: So instead of crashing to desktop at initialization when creating the operations list, it'll crash to desktop when adding a new PublishVideo operation. I guess that's better because at least users will see it happens when adding the operation

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
```diff
@@ -167,25 +142,109 @@ public PublishVideoOperation(InputSocket.Factory inputSocketFactory) {
...
+      }
+    }
   }
+
+  /**
+   * Multi platform method for getting the hostname of the local computer. cscore's
+   * {@link CameraServerJNI#getHostname() getHostName() function} only works on Linux, so we need to
+   * implement the method for Windows and Mac ourselves.
+   */
+  private static String getHostName() {
```
JLLeitschuh: Replace with: `InetAddress.getLocalHost().getHostName()`???

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
```diff
@@ -167,25 +142,109 @@ public PublishVideoOperation(InputSocket.Factory inputSocketFactory) {
 
   @Override
   public void perform() {
-    if (!connected) {
-      return; // Don't waste any time converting images if there's no dashboard connected
+    final long now = System.nanoTime(); // NOPMD
+
+    if (!cscoreLoaded) {
+      throw new IllegalStateException(
+          "cscore could not be loaded. The image streaming server cannot be started.");
```
JLLeitschuh: 👍

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
```diff
@@ -10,146 +10,121 @@
...
       OperationDescription.builder()
           .name("Publish Video")
           .summary("Publish an M_JPEG stream to the dashboard.")
           .category(OperationDescription.Category.NETWORK)
           .icon(Icon.iconStream("publish-video"))
           .build();
-  private static final int PORT = 1180;
-  private static final byte[] MAGIC_NUMBER = {0x01, 0x00, 0x00, 0x00};
+  private static final int INITIAL_PORT = 1180;
+  private static final int MAX_STEP_COUNT = 10; // limit ports to 1180-1189
```
JLLeitschuh: 👍

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
```diff
@@ -167,25 +142,109 @@ public PublishVideoOperation(InputSocket.Factory inputSocketFactory) {
...
-    synchronized (imageLock) {
-      imencode(".jpeg", inputSocket.getValue().get(), imagePointer,
-          new IntPointer(CV_IMWRITE_JPEG_QUALITY, qualitySocket.getValue().get().intValue()));
-      imageLock.notifyAll();
+    copyJavaCvToOpenCvMat(input, publishMat);
+    // Make sure the output resolution is up to date. Might not be needed, depends on cscore updates
+    serverSource.setResolution(input.size().width(), input.size().height());
+    serverSource.putFrame(publishMat);
+    if (lastFrame != -1) {
+      long dt = now - lastFrame;
```
JLLeitschuh: I've seen this actually be `0` even though you are using nano seconds. (This is highly unlikely but we saw it in the camera source code in one bug) This will cause a DIV by zero exception.

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
```diff
@@ -10,146 +10,121 @@
...
+   */
+  private static final boolean cscoreLoaded;
+
+  static {
+    boolean loaded;
+    try {
+      // Loading the CameraServerJNI class will load the appropriate platform-specific OpenCV JNI
+      CameraServerJNI.getHostname();
+      loaded = true;
+    } catch (Throwable e) {
```
JLLeitschuh: Can native bindings ever throw an `InterruptedException`? If so, make sure that this handles it correctly.

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
SamCarlberg: It'll actually be `Double.Infinity` per IEEE spec for floating-point numbers (`1e9 / 0L = Infinity`)

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
SamCarlberg: It's most likely to be `UnsatisfiedLinkError` or a related error. No interrupted exceptions here.

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
SamCarlberg: Oooh. That's much better 👍

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
JLLeitschuh: `(int) 1e9 / 0L` will throw `java.lang.ArithmeticException: / by zero` because of the cast.

JLLeitschuh: In kotlin: `(1e9 / 0L).toInt()` is equal to `2147483647` which is an insanely high frame rate. Is that what you want to set it to?

SamCarlberg: It's technically infinite since the new frame comes at the same time as the previous one. But it doesn't happen here because it's rate limited by the pipeline and the update rate of the pipeline sources. I seriously doubt `dt` will ever be less than a few hundred microseconds.

JLLeitschuh: Okay, that's fine then.

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
```diff
@@ -167,25 +142,109 @@ public PublishVideoOperation(InputSocket.Factory inputSocketFactory) {
...
+   *
+   * <p>A caveat to this approach is that the memory layout used by the OpenCV binaries bundled with
+   * both wrapper libraries <i>must</i> be identical. Using the same OpenCV version for both
+   * libraries should be enough.</p>
+   *
+   * @param javaCvMat the JavaCV Mat wrapper object to copy from
+   * @param openCvMat the OpenCV Mat wrapper object to copy into
+   * @throws RuntimeException if the OpenCV native pointer could not be set
+   */
+  private static void copyJavaCvToOpenCvMat(opencv_core.Mat javaCvMat, Mat openCvMat)
```
JLLeitschuh: Can you add some unit tests for this function to make sure that it gets covered. I don't want someone in the future to be updating dependencies and have this break because this code doesn't get run by a test.

build.gradle
JLLeitschuh: With this commented out as it is is the deploy functionality essentially nerfed? If so, then we should remove the deployer for this season.

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
```diff
@@ -7,149 +7,130 @@
...
+      serverSource = new CvSource("GRIP CvSource " + totalStepCount,
+          VideoMode.PixelFormat.kMJPEG, 0, 0, 0);
+      server.setSource(serverSource);
+
+      ourTable = cameraPublisherTable.getSubTable("GRIP-" + totalStepCount);
+      try {
+        InetAddress localHost = InetAddress.getLocalHost();
+        ourTable.putStringArray("streams",
+            new String[]{
+                generateStreamUrl(localHost.getHostName(), ourPort),
```
SamCarlberg: @PeterJohnson Would it be prudent to add a ".local" to the end of the local hostname e.g. `team-190-driverstation.local`? I'm not sure how mDNS would handle it on an FRC field

core/src/test/java/edu/wpi/grip/core/operations/composite/PublishVideoOperationTest.java
```diff
@@ -0,0 +1,74 @@
...
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
+
+public class PublishVideoOperationTest {
+
+  @BeforeClass
+  public static void initialize() {
+    // Make sure the OpenCV JNI is loaded
+    blackHole(PublishVideoOperation.DESCRIPTION);
```
JLLeitschuh: Derpy but sure.

core/src/test/java/edu/wpi/grip/core/operations/composite/PublishVideoOperationTest.java
```diff
@@ -0,0 +1,74 @@
...
+import org.opencv.core.Mat;
+
+import java.nio.ByteBuffer;
+
+import static org.hamcrest.Matchers.greaterThanOrEqualTo;
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
+
+public class PublishVideoOperationTest {
```
JLLeitschuh: 👍 Thanks for adding this test for us!

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
```diff
@@ -115,8 +113,16 @@ public PublishVideoOperation(InputSocket.Factory inputSocketFactory) {
...
-          new String[]{"mjpeg:http://" + getHostName() + ":" + ourPort + "/?action=stream"});
+      try {
+        InetAddress localHost = InetAddress.getLocalHost();
+        ourTable.putStringArray("streams",
+            new String[]{
+                generateStreamUrl(localHost.getHostName(), ourPort),
+                generateStreamUrl(localHost.getHostAddress(), ourPort)
+            });
+      } catch (UnknownHostException e) {
+        ourTable.putStringArray("streams", new String[0]);
```
JLLeitschuh: Should we log something here?

SamCarlberg: Maybe? The documentation isn't very clear on what circumstances will cause this exception to be thrown. It might never be thrown for the local host since it appears to only happen if the host is unreachable.

JLLeitschuh: Interesting. @PeterJohnson do you have any thoughts on what might cause this on the local system or if this is a non-issue?

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
PeterJohnson: Note `InetAddress.getLocalHost()` may not give you the address you want if the machine has multiple network adapters (https://stackoverflow.com/questions/5813194/inetaddress-getlocalhost-does-not-return-expected-ip-address-from-c-windows-s).  You probably want to use `NetworkInterface.getNetworkInterfaces()` instead.  cscore provides the similar `CameraServerJNI.getNetworkInterfaces()` as well as the NetworkInterfacesChanged event for listening to network interface changes, but at the moment these only work on Linux.  In either case, appending ".local" is almost certainly the wrong thing to do; you should publish the raw IP instead; appending ".local" will require the computer running GRIP to publish its mDNS address, etc.

The streams entry is an array for exactly this reason, so you can e.g. generate an entry for each network address the machine has (e.g. on the robot side we publish with both the USB and Ethernet IP addresses).

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
PeterJohnson: It sounds like how the Java getLocalHost function works is it gets the local hostname, then resolves it into an address.  So it's definitely possible to fail: https://stackoverflow.com/questions/1881546/inetaddress-getlocalhost-throws-unknownhostexception

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
saudet: First it'd say, what is missing from JavaCV that you need to use the other set of bindings? IMO, we should fix that so we don't need to do things like that.

And even when that's necessary, we don't need to use reflection.

Just passing the pointer address around is sufficient, as long as you make sure they are from the same version of OpenCV, compiled with the same compiler, the same flags, etc and to make sure memory isn't prematurely deallocated and that it gets deallocated by the same binaries used to allocate it.

JLLeitschuh: Is there a way of creating a `Mat` just using the memory address??

saudet:
https://docs.opencv.org/java/2.4.2/org/opencv/core/Mat.html#Mat(long) ?

JLLeitschuh:
@SamCarlberg The above solves our problem.

SamCarlberg:
Well, kind of. I'm still wary about binary incompatibilities between the JavaCV builds and the WPI OpenCV builds. The perfect solution would be to have cscore accept raw image data (bytes, width, height) to avoid needing to use the OpenCV bindings altogether. Something like manually creating [cscore images](https://github.com/wpilibsuite/cscore/blob/master/src/main/native/cpp/Image.h).

That'd let us keep the OpenCV bindings as a runtime dependency for cscore instead of a compile-time dependency for GRIP.

PeterJohnson:
I'm working on a raw image API for cscore for exactly that reason.

saudet:
We can still share the data of the image without sharing the whole Mat
struct. That would be a much better solution.

SamCarlberg:
Doesn't that still run the risk of binary incompatibilities?

saudet:
Just sharing image data? No worries at all, just need to make sure
deallocation is done right.

saudet:
On Windows though, symbols are loaded globally for the process, but I think
the Java bindings from OpenCV don't expose the C++ API anyway, so should be
alright.

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
```diff
@@ -160,7 +171,16 @@ public void perform() {
...
-    copyJavaCvToOpenCvMat(input, publishMat);
+    // "copy" the input data to an OpenCV mat for cscore to use
+    // This basically just wraps the mat pointer in a different wrapper object
+    // No copies are performed, but it means we have to be careful about making sure we use the
+    // same version of JavaCV and OpenCV to minimize the risk of binary incompatibility.
+    // This copy only needs to happen once, since the operation input image is always the same
+    // object that gets copied into.  The data address will change, however, if the image is resized
+    // or changes type.
+    if (publishMat == null || publishMat.nativeObj != input.address()) {
+      publishMat = new Mat(input.address());
```
JLLeitschuh: Can you keep the test you had for `copyJavaCvToOpenCvMat ` arround for making sure that we know if the memory models are incompatible when we bump versions in the future?

SamCarlberg: Sure.

JLLeitschuh:
Wow, serious throwback to an old PR. Thanks @saudet!

ThadHouse:
I'm probably going to completely rewrite most of this using the RawFrame API I'm adding to CScore, but I want to leave it open for now for reference for some of the more weird ui stuff that I don't know.

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
JLLeitschuh: @SamCarlberg Ping on this? Thoughts on making sure the memory isn't inconsistently GCed?

ThadHouse: See my comment at the bottom. In the rewrite, all the memory mapping between different opencv versions will be gone. There will be no wpilib opencv interface anymore.

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
ThadHouse: getNetworkInterfaces works on windows and mac now, so we will be able to use the cscore functions for that. My rewrite will do this

ThadHouse:
@SamCarlberg do you want to move the unrelated stuff out of this PR into another PR? Like the DoublePointer stuff? Or has that already been done in other PR's?

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
ThadHouse: As an example, The following class would be added to GRIP. It will use the JavaCV version of OpenCV, and always copy from the raw image into the requested Mat. It is memory safe, and does not depend at all on the internal representation of how an OpenCV Mat looks. There's some tricks on the cscore jni side to avoid allocations when needed, but everything is in spec, no reflection or undefined behavior.

https://github.com/wpilibsuite/allwpilib/blob/13e15663b6570028a9222b4990d141d16fda3a72/cscore/src/dev/java/edu/wpi/cscore/RawCVMatSink.java

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
ThadHouse: Unless we explicitly throw interrupted from the JNI level, nothing in JNI otherwise would throw it. And since its a dumb idea to interrupt a thread anyway, we would never throw it internally. Pretty much all that shows up is UnsatisfiedLinkError, or one of the few runtime exceptions we throw.

Note this should be using forceLoad(), not getHostname() (I don't think we had forceLoad() then), but I will my sure my version does that.

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
ThadHouse: Theres actually nothing to catch, because internally in the JNI loading classes we call System.exit(1). There would be no exception ever caught here. It would be impossible to catch.

ThadHouse: After a little discussion, we will leave the IO Exception that the static constructor can throw in flight, that way its catchable. That will be a change on our side. We will do the same thing to ntcore as well.

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
JLLeitschuh: @SamCarlberg Do we want to wait for that update then?

SamCarlberg: I think so. Thad's offered to do do the rewrite with the new cscore stuff. Maybe we should also use the cameraserver library for doing the stream hosting instead of manually doing the Networktable side

ThadHouse: I wouldn't be able to bring it in directly (because of the opencv dependency) but I could bring in most of it. There's a lot extra we wouldnt need there as well.

ThadHouse: I wouldn't be able to bring it in directly (because of the opencv dependency) but I could bring in most of it. There's a lot extra we wouldnt need there as well.

JLLeitschuh:
@SamCarlberg is this PR outdated now that the one from @ThadHouse is merged?

ThadHouse:
My PR only did inputs, not output.

PR #874, opened 2017-12-28
Add morphology operation

twinters007:


JLLeitschuh:
I'm good with this. I assume you tested this and it worked the way you wanted?

twinters007:
Yeah. Although I'm not sure how you feel about my hard coded enums. I couldn't figure out where to get that value from since the other ones use 'generated' code but I can't find it.

twinters007:
How do I add things to code generation? It's not very useful without it

ui/src/main/resources/edu/wpi/grip/ui/codegeneration/java/operations/CV_morphologyEx.vm
```diff
@@ -0,0 +1,24 @@
+	/**
```
twinters007: Can an admin verify this?

ui/src/main/resources/edu/wpi/grip/ui/codegeneration/python/operations/CV_morphologyEx.vm
```diff
@@ -0,0 +1,14 @@
+    @staticmethod
```
twinters007: I tested this and it works

core/src/main/java/edu/wpi/grip/core/operations/opencv/NewKernelOperation.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  @Override
+  public void perform() {
+    final int widthValue = widthSocket.getValue().get().intValue();
+    final int heightValue = heightSocket.getValue().get().intValue();
+    final int kernelType = typeSocket.getValue().get().value;
+
+    outputSocket.setValue(opencv_imgproc.getStructuringElement(kernelType, new Size(widthValue,
+                          heightValue)));
```
JLLeitschuh: Won't this allocate a new matrix every time this perform method is called.
Don't you instead want to pass in an existing matrix so the memory can be reused?

twinters007: You're right. Ideally you'd create the kernel once, but idk how to get that behavior..

JLLeitschuh: There should be an overload that takes a `dest` as a parameter.

twinters007: I'm not sure I understand.

twinters007:
So i updated the python code to alleviate this I think. Instead of putting the function in process, it puts it in `__init__`. However the java code seems to allocate every variable in process which is...weird. I'm not entirely sure if this is a good idea or not, but it's more close to the code I would write freehand

JLLeitschuh:
@SamCarlberg Do you have any time to finish up this PR? I don't really at this exact moment.

core/src/main/java/edu/wpi/grip/core/operations/opencv/NewKernelOperation.java
ArchdukeTim: It's december of the next year, which means it's my turn to comment 😂 

Going back and looking through this, I understand what we are going for, but not sure how that functionality is created in GRIP. The kernel creating isn't part of the pipeline, just the setup, although I think i'm combining the code generation with the actual application..lemme see what I can come up with for the application part and then I'll worry about code gen later

PR #860, opened 2017-08-05
Re-writing Grip Build to be Strongly Typed

JLLeitschuh:
[//]: # (Please ensure that the "Allow edits from maintainers" checkbox is checked. Thanks!)

Work in progress

PR #900, opened 2019-01-11
Upgrade to Gradle 5.0

SamCarlberg:
- The buildSrc project has been removed
  - The generated classes have been pulled into `:core`
  - Since each run gave the same output, this subproject was effectively redundant
- Each subproject now has its own buildscript
- Switch to the Kotlin DSL
  - The Linux launcher has to stay on the Groovy DSL due to the C++ model not working with the Kotlin one

Supersedes and closes #860


AustinShalit:
And yes... there is a git directory.

.travis.yml
```diff
@@ -36,7 +36,7 @@ install:
 script:
   # Only run code generation tests on OSX -- linux requires sudo to install OpenCV dependencies, and that environment
   # will not be able to run MainWindowTest.testDragOperationFromPaletteToPipeline
-  - ./gradlew check jacocoTestReport jacocoRootReport --stacktrace -Pheadless=true -PlogTests --scan;
```
JLLeitschuh: Why the removal of the `--scan`?

SamCarlberg: It was causing Travis to fail from having to accept the license agreement. Since Travis is getting removed anyway, I'm not worried about temporarily reducing its workload to keep the important stuff working until then

build.gradle.kts
```diff
@@ -0,0 +1,201 @@
+import org.ajoberstar.grgit.Grgit
+import com.github.spotbugs.*
+
+buildscript {
+    repositories {
+        jcenter()
+        maven {
+            setUrl("https://github.com/WPIRoboticsProjects/opencv-maven/raw/mvn-repo")
```
JLLeitschuh: HTTPS support. Wonderful!

Github might get pissy though if we keep using them as a maven repo.

build.gradle.kts
```diff
@@ -0,0 +1,201 @@
+import org.ajoberstar.grgit.Grgit
+import com.github.spotbugs.*
```
JLLeitschuh: Can we avoid wildcard imports?

build.gradle.kts
```diff
@@ -0,0 +1,201 @@
...
+
+repositories {
+    mavenCentral()
+    jcenter()
+}
+
+tasks.withType<Wrapper>().configureEach {
+    gradleVersion = "5.0"
+    distributionType = Wrapper.DistributionType.ALL
+}
```
JLLeitschuh: I usually keep this at the bottom. Why `5.0` instead of `5.1`? The memory leak issue?

build.gradle.kts
```diff
@@ -0,0 +1,201 @@
...
+    }
+}
+
+var grgit: Grgit? = null
+if (rootProject.file(".git").exists()) {
+    project.apply {
+        plugin("org.ajoberstar.grgit")
+    }
+    grgit = Grgit.open()
+}
```
JLLeitschuh: You could do this as:
```kotlin
val grgit: Grgit? by lazy {
	// blah....
}
```

build.gradle.kts
```diff
@@ -0,0 +1,201 @@
...
+        plugin("org.gradle.checkstyle")
+        plugin("net.ltgt.errorprone")
+        plugin("com.github.spotbugs")
+    }
+    repositories {
+        mavenCentral()
+        jcenter()
+        maven {
+            name = "WPILib Maven Release"
+            setUrl("http://first.wpi.edu/FRC/roborio/maven/release")
```
JLLeitschuh: Booo. Not a fan of `http` here.

Can you use the new Gradle repository filers so that only artifacts that are actually on these servers get searched in these repos?

https://docs.gradle.org/current/userguide/declaring_repositories.html#declaring_a_repository_filter

SamCarlberg: first.wpi.edu unfortunately does not have an HTTPS cert
The repository filter looks nice, though

build.gradle.kts
```diff
@@ -0,0 +1,201 @@
...
+        sourceSets = setOf(javaSources["main"], javaSources["test"])
+        excludeFilter = file("$rootDir/findBugsSuppressions.xml")
+        effort = "max"
+    }
+
+    tasks.withType<SpotBugsTask> {
+        reports {
+            xml.isEnabled = false
+            emacs.isEnabled = true
+        }
```
JLLeitschuh: You can get spotbugs to produce an `html` report if we don't want to deal with the emacs one anymore.

build.gradle.kts
```diff
@@ -0,0 +1,201 @@
...
+
+    tasks.named<JacocoReport>("jacocoTestReport") {
+        reports {
+            html.isEnabled = true
+            xml.isEnabled = true
+        }
+    }
+
+    tasks.withType<Javadoc> {
+        source(tasks.named<JavaCompile>("compileJava").map { it.source })
```
JLLeitschuh: 👍

build.gradle.kts
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    javaSubprojects {
+        val sourceSets = (this as ExtensionAware).extensions.getByName("sourceSets") as SourceSetContainer
+        dependsOn(tasks["test"])
+        val srcFiles = files(sourceSets["main"].allSource.srcDirs)
+        additionalSourceDirs(srcFiles)
+        sourceDirectories.from(srcFiles)
+        classDirectories.from(files(sourceSets["main"].output))
+        executionData.from(tasks.named<JacocoReport>("jacocoTestReport").map { it.executionData })
```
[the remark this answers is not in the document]
SamCarlberg: I have no idea. This is just a port of the old task (which wasn't ever running AFAICT) to Kotlin.

settings.gradle
```diff
@@ -1,4 +1,25 @@
...
+     */
+    final String groovyName = "${project.name}.gradle"
+    final String kotlinName = "${project.name}.gradle.kts"
+    project.buildFileName = groovyName
+    if (!project.buildFile.isFile()) {
+        project.buildFileName = kotlinName
+    }
+    assert project.buildFile.isFile(): "File named $groovyName or $kotlinName must exist."
+    project.children.each { setUpChildProject(it) }
+}
```
JLLeitschuh: This file can be Kotlin based now if you want it to be. Up to you though.

ui/preloader/preloader.gradle.kts
```diff
@@ -0,0 +1,13 @@
...
+}
+
+application {
+    mainClassName = "edu.wpi.grip.preloader.GripPreloader"
+}
+
+tasks.named<JavaExec>("run") {
+    classpath = sourceSets["main"].runtimeClasspath
+    main = application.mainClassName
+    args = listOf("windowed")
```
JLLeitschuh: Windowed?

ui/ui.gradle.kts
```diff
@@ -0,0 +1,195 @@
...
+
+dependencies {
+    compile(project(":core"))
+    compile(project(":ui:preloader"))
+    //ideProvider project(path= ":core", configuration= "compile")
+    compile(group = "org.controlsfx", name = "controlsfx", version = "8.40.11")
+    compile(group = "com.hierynomus", name = "sshj", version = "0.16.0")
+    compile(group = "org.apache.velocity", name = "velocity", version = "1.7")
+
+    val coreTestOutput = project(":core").dependencyProject.sourceSets["test"].output
```
JLLeitschuh: Eventually we should fix this better.

ui/ui.gradle.kts
```diff
@@ -0,0 +1,195 @@
...
+    testCompile(files(coreTestOutput))
+    testCompile(files(coreTestOutput.resourcesDir))
+    testCompile(group = "org.testfx", name = "testfx-core", version = "4.0.5-alpha")
+    testCompile(group = "org.testfx", name = "testfx-junit", version = "4.0.5-alpha")
+    testRuntime(group = "org.testfx", name = "openjfx-monocle", version = "1.8.0_20")
+    testCompile(group = "org.opencv", name = "opencv-java", version = "3.1.0")
+}
+
+evaluationDependsOn(":core")
+evaluationDependsOn(":ui:preloader")
```
JLLeitschuh: Oof, we need to fix those. Suuuper hacky. If it works for now, fine. But that's really frowned upon.

ui/ui.gradle.kts
```diff
@@ -0,0 +1,195 @@
...
+}
+
+if (project.hasProperty("generation") || project.hasProperty("genonly")) {
+    val syst = osdetector.os
+    tasks.withType<Test>() {
+        val platform = Installer.getPlatform()
+        Installer.setOpenCvVersion("3.1.0")
+        val jniLocation = project.properties.getOrDefault("jniLocation", platform.defaultJniLocation())
+        Installer.installJni("$jniLocation")
+        val defaultLibPath = System.getProperty("java.library.path");
```
JLLeitschuh: Can this installation stuff happen in a task instead of in the Gradle configuration phase?

build.gradle.kts
SamCarlberg: Darn, the filtering isn't available in 5.0

ui/preloader/preloader.gradle.kts
SamCarlberg: For launching the preloader by itself. See the preloader application class

build.gradle.kts
JLLeitschuh: Is there any sort of hash validation we can do before we just start arbitrarily running code that was downloaded insecurely?

Is there some way we can validate the SHA 256 hash of it first or something?

SamCarlberg: I don't think there's anything we can do beyond what I assume Gradle does already with using the .md5 or .sha hash files in the Maven repo.

This is also out of scope for this PR

JLLeitschuh: Fair enough.

ui/ui.gradle.kts
SamCarlberg: These lines have been removed

ui/ui.gradle.kts
SamCarlberg: Done

PR #901, opened 2019-01-11
Disable code generation tests on Travis

SamCarlberg:
The mac builds hang on installing GCC or make. Disabling this makes the builds work, though the code generation tests will no longer run.


JLLeitschuh:
Wait, so have we now disabled it on both mac and linux? This doesn't bode well.

AustinShalit:
It is ok because azure will be added in the near future.

PR #903, opened 2019-01-13
Set up CI with Azure Pipelines

SamCarlberg:
The pipeline is configured to use Linux, 32- and 64-bit Windows, and macOS.
Code generation tests will run on mac and both Windows platforms, but not Linux
Native installer packages are generated and published as build artifacts

Preview at https://dev.azure.com/SamCarlberg/GRIP/_build

Upgrade Gradle to 4.10.2 to allow use of the `-Dorg.gradle.java.home` property on the Linux

Note: conflicts with #900 due to some changes to the build script

---

An Azure pipeline for this project will be created after this PR is merged. This will replace the Travis and Appveyor integrations


JLLeitschuh:
@SamCarlberg How do you want to handle releases which used to be completely executed through and uploaded by Travis CI?

SamCarlberg:
Using Azure releases

PR #915, opened 2019-01-23
Add contour angle to contours report

SamCarlberg:
Angle is from minAreaRect, and is in the closed range `[-90, 0]`
Rework contours operations to be a bit more functional
Make filter contours a bit more optimized

Add a `LazyInit` class for lazily creating objects, instead of using a bizarre pattern with `Optional`
Add a utility class for streaming Java-fied `std::vector` types such as `MatVector` as Java `Stream` objects. This currently just supports `MatVector`, but can be added to in the future.

### DIFF
--- a/core/src/main/java/edu/wpi/grip/core/util/LazyInit.java
+++ b/core/src/main/java/edu/wpi/grip/core/util/LazyInit.java
@@ -1,0 +1,44 @@
+package edu.wpi.grip.core.util;
+
+import java.util.Objects;
+import java.util.function.Supplier;
+
+/**
+ * A holder for data that gets lazily initialized.
+ *
+ * @param <T> the type of held data
+ */
+public class LazyInit<T> {
+
+  private T value = null;
+  private final Supplier<? extends T> factory;
+
+  /**
+   * Creates a new lazily initialized data holder.
+   *
+   * @param factory the factory to use to create the held value
+   */
+  public LazyInit(Supplier<? extends T> factory) {
+    this.factory = Objects.requireNonNull(factory, "factory");
+  }
+
+  /**
+   * Gets the value, initializing it if it has not yet been created.
+   *
+   * @return the held value
+   */
+  public T get() {
+    if (value == null) {
+      value = factory.get();
+    }
+    return value;
+  }
+
+  /**
+   * Clears the held value. The next call to {@link #get()} will re-instantiate the held value.
+   */
+  public void clear() {
+    value = null;
+  }
+
+}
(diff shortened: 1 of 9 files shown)


core/src/main/java/edu/wpi/grip/core/util/LazyInit.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.Objects;
+import java.util.function.Supplier;
+
+/**
+ * A holder for data that gets lazily initialized.
+ *
+ * @param <T> the type of held data
+ */
+public class LazyInit<T> {
```
JLLeitschuh: So there are no thread safety guarantees offered? Is this intentional?

JLLeitschuh:
Just want to make sure we're not unintentionally introducing any memory leaks with this logic. Are there any native pointers that are created in these new stream operations that need to be cleaned up?

SamCarlberg:
The stream operation itself doesn't allocate anything - it's just standard indexed iteration. A new `RotatedRect` array is created for each contour report object, but the impact in native memory is minimal and the garbage collector can easily handle it.

JLLeitschuh:
Reminder, the garbage collector has no idea how large native objects truly are, so it can leave them around for far longer than you think it will.

If there is a clean place to de-allocate them after usage, please do so. Otherwise, it's fine.

SamCarlberg:
Right. The C++ class is small (only holds a few numbers), so the memory impact is low. We also prompt the garbage collector every few seconds so I'm not too worried

core/src/main/java/edu/wpi/grip/core/util/LazyInit.java
JLLeitschuh: ping!

core/src/main/java/edu/wpi/grip/core/util/LazyInit.java
```diff
@@ -0,0 +1,44 @@
...
+
+  private T value = null;
+  private final Supplier<? extends T> factory;
+
+  /**
+   * Creates a new lazily initialized data holder.
+   *
+   * @param factory the factory to use to create the held value
+   */
+  public LazyInit(Supplier<? extends T> factory) {
```
JLLeitschuh: If you don't make this expose a constructor, it doesn't tightly couple this to needing to be newed.

core/src/main/java/edu/wpi/grip/core/util/LazyInit.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+    return value;
+  }
+
+  /**
+   * Clears the held value. The next call to {@link #get()} will re-instantiate the held value.
+   */
+  public void clear() {
+    value = null;
+  }
```
JLLeitschuh: Why should you be able to clear a lazy?

PR #916, opened 2019-01-24
Update networktables to ntcore 2019.2.1

SamCarlberg:
Removes dependency on old networktables

Call flush after entire pipeline runs to keep all published values synchronized as much as possible - old behavior was to flush for every publisher (_before_ the publisher would make its changes)

Removes the auto-shutdown behavior for NTHandler since clients now only maintain entries set on them (eg from publishers). I believe this behavior was present to prevent GRIP from keeping phantom entries around that it had no business using.

Since NetworkTables now requires libc 2.2.7, GRIP will not run on Ubuntu-based distros older than 18.04.


azure-pipelines.yml
```diff
@@ -6,13 +6,22 @@
...
 
+resources:
+  containers:
+    - container: wpilib2019
+      image: wpilib/roborio-cross-ubuntu:2019-18.04
+
 jobs:
   - job: Linux
     pool:
       vmImage: 'Ubuntu 16.04'
```
mrtidy: I see that this was changed to 18.04 in commit 5e8ee9d but changed back to 16.04 in 63797a9  Since other changes also happened in 63797a9 it wasn't clear to me if this reset was intentional.

SamCarlberg: It was. Azure pipelines only has Ubuntu 16.04, not 18.04

JLLeitschuh:
This totally cans Travis CI right?

core/src/main/java/edu/wpi/grip/core/operations/network/networktables/NTManager.java
```diff
@@ -104,9 +105,21 @@
 
         }
       }
-    }, true);
+    }, 0xFF);
```
JLLeitschuh: Can you label this "magic constant" with a comment explaining it?

core/src/main/java/edu/wpi/grip/core/operations/network/networktables/NTManager.java
```diff
@@ -39,17 +46,14 @@
  */
 @Singleton
 public class NTManager implements Manager, MapNetworkPublisherFactory, MapNetworkReceiverFactory {
-  /*
-   * Nasty hack that is unavoidable because of how NetworkTables works.
-   */
-  private static final AtomicInteger count = new AtomicInteger(0);
```
JLLeitschuh: I forget why this was required. Can you remind me why we had it?

SamCarlberg: Basically, just tracking usages of the NetworkTables connection. The client would only run if there was at least one source or one publishing operation that used it - if not, then the client would be shut down. I believe this was to prevent GRIP from keeping old entries around, but ntcore makes it so that clients only keep around entries that they have written to (the old library would make the client keep around _all_ entries and re-send them to a server on a reconnect).

SamCarlberg:
I think Travis can be removed now. It's just doing the javadoc push to the github site, but I think that's broken anyway. Azure has everything else.

There has been some chatter about compiling ntcore with compatibility for 14.04 or 16.04... I'm tempted to hold out to see if that goes anywhere - or maybe wait until Azure has an 18.04 image (issue [here](https://github.com/Microsoft/azure-pipelines-image-generation/issues/506)) instead of mucking around with Docker

core/src/main/java/edu/wpi/grip/core/operations/network/networktables/NTManager.java
JLLeitschuh: Ping!

core/src/main/java/edu/wpi/grip/core/operations/network/networktables/NTManager.java
```diff
@@ -117,51 +132,60 @@
...
 
-    synchronized (NetworkTable.class) {
-      NetworkTable.shutdown();
-      NetworkTable.setIPAddress(projectSettings.getPublishAddress());
+    synchronized (ntLock) {
+      ntInstance.stopClient();
+      ntInstance.deleteAllEntries();
+      ntInstance.startClient();
+      ntInstance.setServer(projectSettings.getPublishAddress());
+    }
```
JLLeitschuh: Is there any chance that any of these things could hang?

core/src/test/java/edu/wpi/grip/core/sources/NetworkTableEntrySourceTest.java
```diff
@@ -41,16 +43,34 @@ public NetworkTableEntrySourceTest() {
 
   @Before
   public void setUp() {
-    testingNtManager = p -> new MockNTReceiver(p, testTable);
-
-    testTable.putBoolean(BOOLEAN_PATH, true);
-    testTable.putDouble(NUMBER_PATH, TEST_NUMBER);
-    testTable.putString(STRING_PATH, TEST_STRING);
+    ntInstance = NetworkTableInstance.create();
```
JLLeitschuh: Hate this dam singleton pattern.

SamCarlberg: It's not a singleton? There's a default instance, but `create()` will return a new instance. Makes it useful for tests since there's no global state

JLLeitschuh: Oh. Cool!

PR #922, opened 2019-02-09
Save pipeline connections in a consistent order

### LINKED ISSUE #919: Valve and switch operations are flaky in saves
Pipeline connections are saved in a `HashSet`, which gives nondeterministic ordering of connections. XStream saves them in encounter order - which is a problem for the valve and switch operations whose output types rely on the input type.

When loading a save file, the _output_ connection may be deserialized first, before the input, which causes an exception to be thrown if the output connection is connected to an input socket that takes anything but `Object` (i.e. any socket except one on another valve or switch)

SamCarlberg:
Fixes a problem with logical operations (and linked input sockets in general) having output connections deserialized prior to input connections, which lead to crashes when loading save files with operations that use linked input sockets.

Old save files using logical operations remain unusable, but creating a save file with those operations will guarantee a readable save file.

Fixes #919

### DIFF
--- a/core/src/main/java/edu/wpi/grip/core/serialization/PipelineConverter.java
+++ b/core/src/main/java/edu/wpi/grip/core/serialization/PipelineConverter.java
@@ -1,0 +1,174 @@
+package edu.wpi.grip.core.serialization;
+
+import edu.wpi.grip.core.Connection;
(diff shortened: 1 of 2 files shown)


JLLeitschuh (changes requested):
Please add a test if possible. Other than that, this looks good!

PR #926, opened 2019-03-22
Added crop functionality

TheTripleV:
A CropOperation has been added that allows an image to be cropped down to a smaller size. 
Users can specify a location and size for the crop. They can also specify the location of the rectangle from which the crop propogates: top left, bottom left, top right, bottom right, center.

Note: I don't know how I would go about writing/testing the code generation.

Example usage:
image


core/src/main/java/edu/wpi/grip/core/operations/composite/CropOperation.java
```diff
@@ -0,0 +1,123 @@
...
+    final Origin origin = originSocket.getValue().get();
+
+    final Rect regionOfInterest = new Rect(
+            x.intValue() + (int)(origin.xOffsetMultiplier * width.intValue()),
+            y.intValue() + (int)(origin.yOffsetMultiplier * height.intValue()),
+            width.intValue(),
+            height.intValue()
+            );
+
+    final Mat output = new Mat(input, regionOfInterest).clone();
```
JLLeitschuh: This can lead to a memory leak. @SamCarlberg Do you know of a good class to use as an example of how this should be done to avoid the memory leak?

JLLeitschuh:
This is awesome! We'll try to get to it. Can you work on fixing the build so that CI passes?

core/src/main/java/edu/wpi/grip/core/operations/composite/CropOperation.java
SamCarlberg: I don't think the clone is necessary. The  constructor used here uses the same memory as the base image IIRC so there's only minimal overhead. I'll have to take a look when I'm at my computer

JLLeitschuh: Well, we don't want this operation modifying the original input matrix.

TheTripleV:
I fixed the 2 [WhitespaceAfter] errors I believe. However, the [CustomImportOrder] errors are the same ones I see in ResizeOperation (I used ResizeOperation as a starter file).

Also, based on another fork I just saw, I refactored the perform method to not create a new mat every iteration.

TheTripleV:
Right now, I'm just running the Google Checkstyle. Is there a custom checkstyle file for GRIP?

core/src/main/java/edu/wpi/grip/core/operations/composite/CropOperation.java
```diff
@@ -87,14 +88,13 @@ public void perform() {
...
-            y.intValue() + (int)(origin.yOffsetMultiplier * height.intValue()),
+            x.intValue() + (int) (origin.xOffsetMultiplier * width.intValue()),
+            y.intValue() + (int) (origin.yOffsetMultiplier * height.intValue()),
             width.intValue(),
             height.intValue()
             );
 
-    final Mat output = new Mat(input, regionOfInterest).clone();
-
+    input.apply(regionOfInterest).copyTo(output);
```
JLLeitschuh: Does `apply` modify the input? If so, it might be more appropriate to do:
```
input.copyTo(output);
output.apply(regionOfInterest);
```

TheTripleV: `apply` doesn't modify the input. It just returns the specified submatrix of the original Mat.

image
I added a blur off of the input of the crop and it seems to work just fine.

JLLeitschuh: This is 100% the test I was thinking of having you run to ensure that this worked correctly. Thank you for ensuring this. Would you be so kind as to leave a comment explaining this so that 6 months from now when I look at this code I know this is intentional? A simple link to this thread and a comment explaining what you said above should be enough.

JLLeitschuh:
> Right now, I'm just running the Google Checkstyle. Is there a custom checkstyle file for GRIP?

I haven't actually developed in this codebase in so long that I honestly forget. @SamCarlberg can you help out here?

SamCarlberg:
Yes there is: https://github.com/WPIRoboticsProjects/GRIP/blob/master/checkstyle.xml

Run `gradlew checkstyleMain` to run checkstyle on the production sourcesets, or `gradlew check` to run all code checks (checkstyle, PMD, ...) as well as unit tests. This is what we have CI use to check pull requests.

SamCarlberg:
@TheTripleV With the recent addition of CUDA support, you'll need to change the `Mat` sockets to use `MatWrapper`

I can add the code generation in a followup if you're not comfortable with it

TheTripleV:
Cool. I'll take a look at it after my finals next week. I'm just unsure on running tests on GRIP in general. For the operation files, I see test cases/files written. But for code generation, do we just run each piece of generated code manually?

TheTripleV:
I updated the crop operation to use MatWrapper. Interestingly, the operation performs just fine without a clone. This is because every operation that occurs after a crop makes a copy of the data in the MatWrapper. 

Should I leave the `.clone()` out for performace (less memory copies) or should it be in for safety's sake?

JLLeitschuh:
@SamCarlberg ^^

SamCarlberg:
You can safely omit calling `clone()`

core/src/main/java/edu/wpi/grip/core/operations/composite/CropOperation.java
```diff
@@ -0,0 +1,129 @@
...
+            y.intValue() + (int) (origin.yOffsetMultiplier * height.intValue()),
+            width.intValue(),
+            height.intValue()
+            );
+
+    //apply() returns a sub-matrix; It does not modify the input Mat: https://github.com/WPIRoboticsProjects/GRIP/pull/926
+    if (input.isCpu()) {
+      output.set(input.getCpu().apply(regionOfInterest)); // .clone() optional
+    } else {
+      output.set(input.getGpu().apply(regionOfInterest)); // .clone() optional
```
JLLeitschuh: I don't think the `.clone()` comments are necessary.

TheTripleV:
I added Java Code Generation. One weird thing I haven't been able to figure out is how it sets default values.
```
	public void process(Mat source0) {
		// Step Crop0:
		Mat cropInput = source0;
		double cropX = 100;
		double cropY = 100;
		double cropWidth = 50;
		double cropHeight = 50;
		Origin cropOrigin = Center;
		crop(cropInput, cropX, cropY, cropWidth, cropHeight, cropOrigin, cropOutput);

	}
```
In this step, `Origin cropOrigin = Center;` should be `Origin cropOrigin = Origin.get("Center");` similar to Blur's code generation.

SamCarlberg:
There's a hardcoded macro in `ui/src/main/resources/edu/wpi/grip/ui/codegeneration/java/macros.vm` for doing the enum lookups:

```
#elseif ($input.type().equals("BlurType"))
		$input.type() ${tMeth.name($input.name())} = ${input.type()}.get("${input.value()}");
```

You should be able to add an 'or' there to generate the expected enum lookup

```
#elseif ($input.type().equals("BlurType") || $input.type().equals("Origin"))
```

JLLeitschuh:
Sorry for letting this sit for so long. Happy to merge if it's in a working state.

JLLeitschuh:
Thanks for your contribution @TheTripleV!!

PR #932, opened 2019-04-02
Update to Java 11

### LINKED ISSUE #890: Update to support JDK 10
@SamCarlberg & @AustinShalit how do we want to handle this? The rest of WPI lib is working towards updating to run under JDK 10.

WPILib bundles a JAR version of GRIP with the plugins that we need to ensure works with JDK 10 correct?
As far as I know, everyone else uses the installer which bundles the JRE with the application meaning we won't have an issue there.

Thoughts?

SamCarlberg:

Since the native packager tool was removed in Java 11, building native installers requires JDK 13 with a `jpackage` tool to be installed. Builds can be downloaded [here](https://jdk.java.net/jpackage/) for 64-bit Windows, Mac, and Linux. No win32 build is yet available.

Updates PMD, since older versions cannot run on Java 11. Most of the changes from this PR are due to new PMD rules.

Errorprone and spotbugs are disabled due to problems with Java 11

Grgit has been removed and replaced with a manual call to `git describe --tags` to get the version number.

UI tests are disabled due to TestFX not working on Java 11 when headless, and is bugged even when not headless (some tests require manual input).

Native installers can be created with `./gradlew :ui:jpackage`. Generates a `.exe` for Windows, `.dmg` for Mac, and `.deb` for Linux in `ui/build/installer/`. The location of the JDK install can be specified with the property `jdk13`, eg `./gradlew :ui:jpackage -Pjdk13=/path/to/jdk13/home`

 File associations are available on Windows (so opening a `.grip` file will launch GRIP), but are not yet working for Mac or Linux. Updates to the `jpackage` tool will hopefully let us add that functionality further down the line.

Closes #890


build.gradle.kts
```diff
@@ -22,8 +20,8 @@ plugins {
     id("com.github.johnrengelman.shadow") version "4.0.3"
     id("com.google.osdetector") version "1.4.0"
     id("org.ajoberstar.grgit") version "2.0.0" apply false
-    id("net.ltgt.errorprone") version "0.0.16"
-    id("com.github.spotbugs") version "1.6.4"
+    //id("net.ltgt.errorprone") version "0.0.16"
+    //id("com.github.spotbugs") version "1.6.4"
```
JLLeitschuh: Spotbugs doesn't work on 11?

buildSrc/src/main/kotlin/JpackageExec.kt
```diff
@@ -0,0 +1,313 @@
...
+ *
+ * Because JDK 13 has not yet been released, and because `jpackage` is not yet included in the
+ * early-access builds, a compatible JDK must be downloaded from
+ * [https://jdk.java.net/jpackage/](https://jdk.java.net/jpackage/). The location of the JDK must be
+ * specified with [jdkHome] (eg `/opt/java/jdk-13/`)
+ *
+ */
+open class JpackageExec : DefaultTask() {
+
+    private val objectFactory = project.objects
```
JLLeitschuh: Annotate as `@Internal`

buildSrc/src/main/kotlin/JpackageExec.kt
```diff
@@ -0,0 +1,313 @@
...
+            applicationVendor.ifPresent { vendor ->
+                args.addAll("--vendor", vendor)
+            }
+            identifier.ifPresent { id ->
+                args.addAll("--identifier", id)
+            }
+            fileAssociations.ifPresent { propsFile ->
+                args.addAll("--file-associations", propsFile.asFile.absolutePath)
+            }
+            args.addAll("--installer-type", installerType.get())
```
JLLeitschuh: Consider extracting all of this to a private function?

buildSrc/src/main/kotlin/PlatformMapper.kt
```diff
@@ -0,0 +1,54 @@
...
+    }
+    val osArch = System.getProperty("os.arch")
+    val arch: String =
+            if (osArch.contains("x86_64") || osArch.contains("amd64")) {
+                "64"
+            } else if (osArch.contains("x86")) {
+                "32"
+            } else {
+                throw UnsupportedOperationException(osArch)
+            }
```
JLLeitschuh: Formatting inconsistency with the `when` statement above.

buildSrc/src/main/kotlin/PlatformMapper.kt
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Generates a classifier string for a platform-specific JavaFX artifact.
+*/
+fun javaFxClassifier(platform: NativePlatforms) = when (platform) {
+    WIN32 -> "win32"
+    WIN64 -> "win"
+    MAC -> "mac"
+    LINUX -> "linux"
+}
```
JLLeitschuh: Any particular reason why these can't just all live as fields on the enum?

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
```diff
@@ -80,7 +80,7 @@
 
           final int fps = socketInputStream.readInt();
           final int compression = socketInputStream.readInt();
-          final int size = socketInputStream.readInt();
+          // final int size = socketInputStream.readInt(); // Unused
```
JLLeitschuh: Consider removing? That being said, not reading the int might fail to progress the input stream to read the next bit of data. Are you sure this isn't breaking anything?

JLLeitschuh:
Is there any way to preserve running the UI tests with Travis on JDK 8 if we make it so the build can compile on both platforms?

buildSrc/src/main/kotlin/PlatformMapper.kt
SamCarlberg: Different libraries use different classifiers for native components. Since there's at least three different standards (javafx, javacpp, and wpilib), and possibly more in the future, and some libraries don't have artifacts for all platforms, better to split it out like this.

core/src/main/java/edu/wpi/grip/core/operations/composite/PublishVideoOperation.java
SamCarlberg: Ah, good point. I'll have to check that

buildSrc/src/main/kotlin/PlatformMapper.kt
JLLeitschuh: Okay.

SamCarlberg:
Running UI tests on Java 8 would probably be possible - just don't add the JavaFX deps and use the appropriate versions of controlsfx and testfx. I'll have to do some digging

buildSrc/src/main/kotlin/JpackageExec.kt
SamCarlberg: Since the property is `private`, shouldn't that be unnecessary? There's no public accessor method like there are on no-modifier (public) properties.

buildSrc/src/main/kotlin/JpackageExec.kt
SamCarlberg: All of what?

build.gradle.kts
SamCarlberg: Updating the Gradle plugin seems to do the trick... However, there are several new false positives

JLLeitschuh:
> Running UI tests on Java 8 would probably be possible - just don't add the JavaFX deps and use the appropriate versions of controlsfx and testfx. I'll have to do some digging

I believe that preserving UI tests is worth the extra overhead.

Without them, we lose a significant percentage of our test coverage.

core/src/main/java/edu/wpi/grip/core/sockets/SocketHints.java
```diff
@@ -75,6 +77,7 @@ private SocketHints() { /* no op */ }
 
   /* PRIVATE PARTIALLY CONSTRUCTED BUILDERS BELOW */
 
+  @SuppressFBWarnings(value = "UPM", justification = "False positive")
```
JLLeitschuh: "UMP"?

SamCarlberg: **U**nused **P**rivate **M**ethod. This method is _clearly_ used elsewhere in this file, so it's a false positive

ui/ui.gradle.kts
```diff
@@ -187,80 +201,88 @@ tasks.register<Copy>("collectDependencies") {
...
+        description = "Generates a runtime image for a native installer."
+
+        val outputDir = project.buildDir.resolve("jlink").absolutePath
+        outputs.file(outputDir)
+        delete(outputDir)
+
+        setCommandLine(
+                Jvm.current().javaHome.resolve("bin/jlink").absolutePath,
+                "--add-modules",
+                "java.base,java.desktop,java.datatransfer,java.logging,java.management,java.naming,java.prefs,java.scripting,java.sql,java.xml,jdk.dynalink,jdk.scripting.nashorn,jdk.unsupported",
```
JLLeitschuh: Can you turn this into a variable as a list and then when its used, `joinToString`?
That way, if we need to add more, it's easier/clearer what's been added/removed.

core/src/main/java/edu/wpi/grip/core/sockets/SocketHints.java
JLLeitschuh: Huh. Okay then. Thanks!

ui/ui.gradle.kts
SamCarlberg: Makes sense

JLLeitschuh:
Why the 14% code coverage drop? Also, it's good to see the @codecov-io bot back on this PR.

SamCarlberg:
Codecov is currently running on Azure, so UI tests aren't included in the reports

JLLeitschuh:
> Codecov is currently running on Azure, so UI tests aren't included in the reports

😢 Ah well.

buildSrc/src/main/kotlin/JpackageExec.kt
JLLeitschuh: The arg creation.

JLLeitschuh:
Status on this?

SamCarlberg:
Currently getting UI tests and JaCoCo running on its own VM with Java 8

JLLeitschuh:
Ping me when you want me to take a final look. Thanks for investing the time to keep the UI tests running. They are really valuable to have!

PR #933, opened 2019-04-10
Add CUDA acceleration

SamCarlberg:

## Overview
Adds CUDA support to the GRIP runtime. A CUDA-enabled build will set a flag in `MANIFEST.MF` to let the runtime know that it is is using OpenCV builds with CUDA extensions; this is required since the CUDA-enabled OpenCV binaries cannot load if there is no compatible CUDA runtime installed on the system.

Add a `MatWrapper` class for semi-transparently working with an image on CPU and GPU memory. Data is lazily copied between host and device when needed.

Add fixed exit codes for use with `SafeShutdown`, to keep the codes organized. Add an exit code for missing CUDA rutnime for CUDA-enabled builds.

Add a `flagChanged` method to `OutputSocket` to avoid hacky calls to `socket.setValue(socket.getValue().get())`

Add a partial `OutputSocket` implementation with `CudaSocket` - contains a boolean value to flag if CUDA acceleration is available. Will always contain `false` if CUDA is not available, either as a missing runtime or if a CPU-only build of OpenCV is used. Displayed as a regular checkbox, which will be disabled if CUDA is unavailable.

### TODO
- [x] Windows CUDA runtime detection
- [x] Mac CUDA runtime detection
- [x] Linux CUDA runtime detection
- [x] Confirm expected behavior of CUDA-enabled builds on systems without CUDA
- [ ] Investigate more usecases for CUDA acceleration (eg cascade classifiers)

## Updated Operations with CUDA Support
- BlurOperation: Can CUDA-accelerate Gaussian blurs and bilateral filters. Box and median blurs are still CPU-only. Bilateral filter is 10x-100x faster in CUDA depending on blur radius
- CannyEdgeOperation
- DesaturateOperation
- NormalizeOperation (single-channel inputs only)
- CV Absdiff
- CV add
- CV addWeighted
- CV bitwise_and
- CV bitwise_not
- CV bitwise_or
- CV bitwise_xor
- CV compare
- CV subtract
- CV cvtColor
- CV Sobel (~15x speedup versus CPU)
- CV Threshold

Note: speedups of CPU versus CUDA are as measured on my desktop computer

| Component | Name | Speed |
|---|---|---|
| CPU | i7-7700k | 4800MHz |
| RAM | 2x16GB DDR4 | 2133MHz |
| GPU | GTX 980Ti | 1450MHz core/3750MHz mem |


core/core.gradle.kts
```diff
@@ -15,15 +15,21 @@ application {
 val os = osdetector.classifier.replace("osx", "macosx").replace("x86_32", "x86")
 val arch = osdetector.arch.replace("x86_64", "x64")
 
+val withCuda = project.hasProperty("cuda") || project.hasProperty("WITH_CUDA")
+
+if (withCuda) {
+    version = "$version-cuda"
+}
```
JLLeitschuh: Is the plan to just build a different version of GRIP with CUDA support?

core/src/main/java/edu/wpi/grip/core/MatWrapper.java
```diff
@@ -0,0 +1,391 @@
...
+ *
+ * <p>Data is lazily copied between host and device memory when needed. Wrappers that
+ * are only accessed from CPU operations will never have their data stored in device memory.
+ * Accessing a wrapper from CUDA land with {@link #getGpu()} will first copy the existing data from
+ * host memory to device, if the wrapper was most recently accessed from CPU code (from either
+ * {@link #getCpu()} or {@link #rawCpu()}). This behavior also applies in the reverse, for wrappers
+ * accessed from CPU land when they have been most recently used from CUDA code.
+ */
+@SuppressWarnings("PMD.GodClass")
+public final class MatWrapper {
```
JLLeitschuh: Is this class thread safe? Does it need to be?

core/src/main/java/edu/wpi/grip/core/MatWrapper.java
```diff
@@ -0,0 +1,391 @@
...
+   * @param ifGpu the function to call if this is backed by a mat in GPU memory
+   * @param <T>   the type of the property to extract
+   */
+  private <T> T extract(Function<Mat, T> ifCpu, Function<GpuMat, T> ifGpu) {
+    if (isCpu) {
+      return ifCpu.apply(cpuMat);
+    } else {
+      return ifGpu.apply(gpuMat);
+    }
+  }
```
JLLeitschuh: 👍 Clean & straightforward!

core/core.gradle.kts
SamCarlberg: Yeah. CUDA-enabled OpenCV will fail to load if there's no CUDA runtime available, and I'm not sure if it's possible to trick JavaCPP into loading the non-CUDA version at runtime

core/src/main/java/edu/wpi/grip/core/MatWrapper.java
SamCarlberg: No, and I don't think it needs to be. `Mat` is not thread safe and we've been fine so far with our existing synchronization constructs.

core/core.gradle.kts
JLLeitschuh: Classloader hackyness? A thought you could experiment with after this PR is merged.

core/src/main/java/edu/wpi/grip/core/sockets/CudaSocket.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Optional;
+
+import javax.annotation.Nullable;
+
+/**
+ * A type of input socket that lets an operation know that it should prefer to use a
+ * CUDA-accelerated code path. If no compatible CUDA runtime is available, sockets of this type
+ * will <i>always</i> have a value of {@code false} and cannot be changed.
+ */
+public class CudaSocket extends InputSocketImpl<Boolean> {
```
JLLeitschuh: `CudaInputSocket`?

core/src/main/java/edu/wpi/grip/core/util/SafeShutdown.java
```diff
@@ -14,6 +14,31 @@
...
+    /**
+     * The HTTP server cannot start (typically due to the port already being in use).
+     */
+    public static final int HTTP_SERVER_COULD_NOT_START = 0x02;
+
+    /**
+     * CUDA is required by OpenCV but no compatible runtime is available on the system.
+     */
+    public static final int CUDA_UNAVAILABLE = 0x04;
+  }
```
JLLeitschuh: Prefer an enum with `code` field on it over this.

core/src/test/java/edu/wpi/grip/core/serialization/ProjectTest.java
```diff
@@ -269,24 +271,25 @@ public void testPerformSerializedPipeline() throws Exception {
 
   @Test
   @SuppressWarnings("unchecked")
+  @Ignore("opencv_core.compare segfaults")
```
JLLeitschuh: ❗️ Plans to fix this before merging?

JLLeitschuh:
So far, everything looks good. Poke me when you want me to take another pass.

core/src/test/java/edu/wpi/grip/core/serialization/ProjectTest.java
SamCarlberg: Fixed

core/src/main/java/edu/wpi/grip/core/Loggers.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.logging.FileHandler;
+import java.util.logging.Handler;
+import java.util.logging.Level;
+import java.util.logging.LogManager;
+import java.util.logging.LogRecord;
+import java.util.logging.Logger;
+import java.util.logging.SimpleFormatter;
+import java.util.logging.StreamHandler;
+
+public final class Loggers {
```
JLLeitschuh: Solid change.

core/src/main/java/edu/wpi/grip/core/GripCudaModule.java
```diff
@@ -0,0 +1,61 @@
...
+      logger.log(Level.WARNING, "Could not read manifest to determine CUDA acceleration", e);
+    }
+    if (usingCuda) {
+      bind(AccelerationMode.class).to(CudaAccelerationMode.class);
+    } else {
+      bind(AccelerationMode.class).to(NullAccelerationMode.class);
+    }
+
+    bind(CudaVerifier.class).in(Scopes.SINGLETON);
+  }
```
JLLeitschuh: Do you want to move this logic to a construction phase? Like a method `GripCudaModule.create()` method, that way there's no dynamic logic inside of `configure`?

Related: https://github.com/google/guice/wiki/AvoidConditionalLogicInModules

core/src/test/java/edu/wpi/grip/core/cuda/CudaVerifierTest.java
```diff
@@ -0,0 +1,49 @@
+package edu.wpi.grip.core.cuda;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class CudaVerifierTest {
```
JLLeitschuh: Cool!

core/core.gradle.kts
```diff
@@ -66,6 +68,21 @@ tasks.withType<Jar>().configureEach {
...
+        file.delete()
+        file.parentFile.mkdirs()
+        file.createNewFile()
+        Files.write("Hardware-Acceleration: ${if (withCuda) "CUDA" else "None"}\n", file, StandardCharsets.UTF_8)
+    }
+}
+
+tasks.withType<JavaCompile>().configureEach {
+    dependsOn("generateCudaFile")
+}
```
JLLeitschuh: Might just want to use a properties file in the `resources` directory that you configure using the `WriteProperties` task.
https://docs.gradle.org/current/dsl/org.gradle.api.tasks.WriteProperties.html

core/src/main/java/edu/wpi/grip/core/MatWrapper.java
```diff
@@ -100,6 +110,13 @@ private MatWrapper(GpuMat gpuMat) {
     this.isCpu = false;
   }
 
+  private MatWrapper(Mat cpuMat, GpuMat gpuMat) {
+    Objects.requireNonNull(cpuMat, "cpuMat");
+    Objects.requireNonNull(gpuMat, "gpuMat");
+    this.cpuMat = cpuMat;
+    this.gpuMat = gpuMat;
+  }
```
JLLeitschuh: Instead of having so many overloads for the constructor, you can instead just have all of those creation methods call to this one constructor.

Might reduce the amount of code.

Also, you can do:
```
this.cpuMat = requireNonNull(cpuMat, "cpuMat");
```

core/core.gradle.kts
SamCarlberg: That looks perfect!

core/src/main/java/edu/wpi/grip/core/MatWrapper.java
SamCarlberg: Makes sense. It is a lot of clutter.

core/src/test/java/edu/wpi/grip/core/cuda/CudaVerifierTest.java
SamCarlberg: IOC and composition makes this a pretty trivial class 🙂

PR #937, opened 2019-04-18
Manually set GH_TOKEN env var

SamCarlberg:
Should fix Javadoc publishing this time

### DIFF
--- a/.azure-scripts/push-javadoc-to-gh-pages.sh
+++ b/.azure-scripts/push-javadoc-to-gh-pages.sh
@@ -15,15 +15,15 @@ echo -e "Publishing javadoc...\n"
 cp -R build/docs/javadoc $HOME/javadoc-latest
 
 cd $HOME
-git config user.email "azuredevops@microsoft.com"
-git config user.name "azure-pipelines"
+git config --global user.email "azuredevops@microsoft.com"
+git config --global user.name "azure-pipelines"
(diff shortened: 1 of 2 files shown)


.azure-scripts/push-javadoc-to-gh-pages.sh
```diff
@@ -15,15 +15,15 @@ echo -e "Publishing javadoc...\n"
...
+git config --global user.email "azuredevops@microsoft.com"
+git config --global user.name "azure-pipelines"
 git clone --quiet --branch=gh-pages https://${GH_TOKEN}@github.com/WPIRoboticsProjects/grip gh-pages #> /dev/null
 
 cd gh-pages
 git rm -rf ./javadoc
 cp -Rf $HOME/javadoc-latest ./javadoc
 git add -f .
-git commit -m "Lastest javadoc on successful Azure build ${BUILD_BUILD_NUMBER} auto-pushed to gh-pages"
+git commit -m "Lastest javadoc on successful Azure build ${BUILD_BUILDNUMBER} auto-pushed to gh-pages"
```
JLLeitschuh: That's weird. But sure.

SamCarlberg: The build variable is `Build.BuildNumber`, but when it's transformed to an environment variable they just replace `.` with `_` and make the whole thing uppercase. So it's kinda awkward like this

PR #939, opened 2019-04-19
Use native configurations for native artifacts

SamCarlberg:
Should slightly cut down on install sizes by only including the appropriate native libraries.

### DIFF
--- a/core/core.gradle.kts
+++ b/core/core.gradle.kts
@@ -12,14 +12,13 @@ application {
     mainClassName = "edu.wpi.grip.core.Main"
 }
 
-val os = osdetector.classifier.replace("osx", "macosx").replace("x86_32", "x86")
-val arch = osdetector.arch.replace("x86_64", "x64")
-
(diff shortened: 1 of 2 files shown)


core/core.gradle.kts
```diff
@@ -12,24 +12,23 @@ application {
...
-val os = osdetector.classifier.replace("osx", "macosx").replace("x86_32", "x86")
-val arch = osdetector.arch.replace("x86_64", "x64")
-
 val withCuda = project.hasProperty("cuda") || project.hasProperty("WITH_CUDA")
 
 if (withCuda) {
     version = "$version-cuda"
 }
 
+createNativeConfigurations()
```
JLLeitschuh: Where is this method from?

SamCarlberg: [buildSrc/NativeConfigurations.kt](https://github.com/WPIRoboticsProjects/GRIP/blob/master/buildSrc/src/main/kotlin/NativeConfigurations.kt)

JLLeitschuh: Why isn't this added as part of a plugin?

PR #944, opened 2019-04-25
Make previews always fit vertically in the preview area

SamCarlberg:
No more previews clipping below the bottom that forced users to scroll up and down to see information displayed below the image (eg contour count)

Reimplement TitledPane and ImageView to get proper behavior. JavaFX `TitledPane` leaves a gap around the image when it's shrunk, and `ImageView`  is atrocious at resizing to fit its parent.

Remove the overloads in `ImageConverter` that take desired height, since the UI is now responsible for doing the resize calculations.

This is a better implementation of #711 and fix for #710

# Screenshots

**Full size previews**
GRIP Computer Vision Engine | Edited_012

**Partially shrunk**
GRIP Computer Vision Engine | Edited_014

**Minimum size previews**
GRIP Computer Vision Engine | Edited_015

### DIFF
--- a/ui/src/main/java/edu/wpi/grip/ui/preview/BlobsSocketPreviewView.java
+++ b/ui/src/main/java/edu/wpi/grip/ui/preview/BlobsSocketPreviewView.java
@@ -83,7 +83,7 @@ public final class BlobsSocketPreviewView extends ImageBasedPreviewView<BlobsRep
       final Mat output = tmp;
       final int numBlobs = blobsReport.getBlobs().size();
       platform.runAsSoonAsPossible(() -> {
-        final Image image = this.imageConverter.convert(output, getImageHeight());
+        final Image image = this.imageConverter.convert(output);
         this.imageView.setImage(image);
         this.infoLabel.setText("Found " + numBlobs + " blobs");
       });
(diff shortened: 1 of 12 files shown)


JLLeitschuh:
Ooof. Not 100% sure I'm on board with this change.

When I resize an image, I kinda want the image itself to be the same size across the entire preview pane. I really don't care that much about the size of the outer box.

What's the reasoning here? Do you really think it's a better UX experience?

SamCarlberg:
The problem with the current behavior is that extra information is hidden below the scrollpane divider, even though the images are shrunk. This makes it difficult, for example, to see the number of contours when the preview area is vertically small. The current behavior does shrink the images so they're (usually) always entirely visible, but this PR makes the _entire_ preview always visible.

JLLeitschuh:
@AustinShalit @JacisNonsense @bradamiller Seeking feedback on the UI changes in this PR.

I just want to make sure that we get solid feedback from users.

bradamiller:
@JLLeitschuh I'm not sure what the alternative would be since the information below the image (in this case, the number of contours) needs to also be visible. To keep all the images the same size, we would have to shrink all of them, leaving white space on the others to match the space taken by the text.

JLLeitschuh:
> To keep all the images the same size, we would have to shrink all of them, leaving white space on the others to match the space taken by the text.

That's an interesting thought, although, I don't think it would look good visually.

I guess the question is, is it more important to have consistent sizes in images or is the consistent size of the preview panes more important.

I don't really have an answer to this question.

Octogonapus:
@JLLeitschuh My thoughts are the same as @bradamiller's. If you want all the images to be the same size, then boxes with extra info will cause a gap. I think it looks fine as-is.

AustinShalit:
I think this change looks good.  My brain is ok dealing with the scaling difference when looking at the images and I would rather this than part of the image always being cut off.

JacisNonsense:
Looks good to me, should help with those tiny KoP screens.

PR #946, opened 2019-05-08
Fix UI tests on Azure

SamCarlberg:
Switch to Linux for running the UI tests
Install OpenJFX and openjdk through `apt`

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -203,13 +203,16 @@ jobs:
       vmImage: 'xcode9-macos10.13'
 
     steps:
+      - script: |
+          mkdir build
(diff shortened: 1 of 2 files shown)


SamCarlberg:
@JLLeitschuh Thoughts on why the code coverage dropped so much?

JLLeitschuh:
No idea. This doesn't seem like the kind of thing that would cause it. Perhaps there is binary jacoco comparison going on?

SamCarlberg:
I think it _might_ be because there was no coverage report for the latest commit on master. I'm not sure, though

SamCarlberg:
Oh, the UI tests aren't running because Java 11 is being used instead of Java 8 :man_facepalming:

SamCarlberg:
ZuluFX is an OpenJDK + OpenJFX build provided by Azul Systems (https://www.azul.com/). We used the embedded Zulu JRE on the RoboRIO for the 2018 FRC season

PR #960, opened 2020-01-30
Fix java 14 url to point to always existing version

ThadHouse:
This should mean we no longer need to update the version every time oracle removes one.


buildSrc/src/main/kotlin/JpackageExec.kt
```diff
@@ -234,7 +234,7 @@ open class JpackageExec : DefaultTask() {
             fileAssociations.ifPresent { propsFile ->
                 args.addAll("--file-associations", propsFile.asFile.absolutePath)
             }
-            args.addAll("--package-type", installerType.get())
+            args.addAll("--type", installerType.get())
```
SamCarlberg: Why is this change necessary?

ThadHouse: Because they keep changing flags in jpackage. That's why we're going to a fixed version we're hosting.

PR #967, opened 2020-09-08
GitHub Actions Build

AustinShalit:
This also fixes the release process.

I'd like to merge this PR because it is equivalent to the Azure build but when testing I ran into a memory issue that prevents MacOS builds from launching. That is an independent issue.

You can view a test release created by this build process here: https://github.com/AustinShalit/GRIP/releases/tag/v0.0.0-test. The artifacts were automatically populated.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,73 @@
...
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: joschi/setup-jdk@v2
+        id: setup-java-14
+        with:
+          java-version: 14
+          architecture: ${{ matrix.architecture }}
+      - uses: joschi/setup-jdk@v2
```
JLLeitschuh, line 46: Why not use the official one? Why use this one instead?
https://github.com/actions/setup-java

AustinShalit: This way it matches exactly what Azure had. The official one uses Azul JDK

JLLeitschuh (approved):
LGTM!

